PR #12, opened 2015-07-09
Methods for query builder to filter queries by JSON fields

elhigu:
Adds initial support for filtering queries with Json field values. Initial API is designed heavily to just implement PostgreSQL 9.4 jsonb operators http://www.postgresql.org/docs/9.4/static/functions-json.html 

Query builder API hides some ugliness that postgres requires in json syntax by using type information and adding automatic type casts when possible.

TODO (create issues of these if necessary):
* or and not variants of methods
* enable .whereJsonHas APIs when knex issue https://github.com/tgriesser/knex/issues/519 is resolved
* what to do for json keys containing ? marks
* implement escape support for json field expression parser
* figure out if we could extend all normal knex query builder methods to support json field expression
* cookbook example how to create index and use json fields
* add support for more complex syntax to compare two json field references (currently only equality and superset, subset supports it)


elhigu:
How can I add parser code to ignored from coverage calculations? There is few abandoned private functions which are never used by parser.

elhigu:
And looks like travis server has too old postgresql to run jsonb test. 9.4 is first version having the jsonb operator support.

elhigu:
Looks like the coverage problems is actually caused by methods not currently tested, because of knex issue mentioned in description. Should I comment the implementations out until knex is fixed?

src/QueryBuilder.js
```diff
@@ -1632,4 +1859,28 @@ function tryBuild(builder) {
...
+ * @param expression
+ * @returns {Array}
+ */
+function parseFieldExpression(expression, extractAsText) {
+  var parsed = jsonFieldExpressionParser.parse(expression);
+  var jsonRefs = _(parsed.access).pluck('ref').value().join(",");
+  var extractor = extractAsText ? '#>>' : '#>';
+  // TODO: Checkout if knex has some utility function to add correct kind of quotes to column name
+  //       this one is for PostgreSQL
+  var middleQuotetColumnName = parsed.columnName.split('.').join('"."');
```
elhigu: typo :)

koskimas:
You can exclude stuff from coverage report using [instanbul's annotations](https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md).

src/QueryBuilder.js
```diff
@@ -3,7 +3,11 @@
 var _ = require('lodash')
   , Promise = require('bluebird')
   , RelationExpression = require('./RelationExpression')
-  , ValidationError = require('./ValidationError');
+  , ValidationError = require('./ValidationError')
+  , utils = require('./utils')
+  , jsonFieldExpressionParser = require('./jsonFieldExpressionParser')
+  , knex = require('knex')
+  ;
```
koskimas: Put is after the last require for consistency.

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
 
 /**
+ * Json query APIs
+ */
+
+/**
+ * Json field filter.
+ *
+ * Supports having field expression in both sides of operand, but right hand value cannot
+ * be any other value, but only string reference to other column/json filed, json array
```
koskimas: Typo: "filed" --> "field". Also "be any other value, but only string reference to" could be clearer like this "be any other value than string reference to".

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+ * or json object.
+ *
+ * Converts left and right hand values to PostgreSQL acceptable format and add user chosen
+ * operator between left and right hand expressions.
+ *
+ * @param fieldExpression {String} Reference to column / jsonField.
+ * @param jsonObjectOrFieldExpression {Object|Array|String} Reference to column / jsonField or json object.
```
koskimas: Put the parameter type before the name for consistency.

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+ * @returns {QueryBuilder}
+ */
+QueryBuilder.prototype.whereJsonObject = function (fieldExpression, operator, jsonObjectOrFieldExpression) {
+  var fieldReference = parseFieldExpression(fieldExpression);
+  if (_.isString(jsonObjectOrFieldExpression)) {
+    var rightHandReference = parseFieldExpression(jsonObjectOrFieldExpression);
+    var refRefQuery = ["(", fieldReference, ")::jsonb ", operator, " (", rightHandReference, ")::jsonb"].join("");
+    return this.whereRaw(refRefQuery);
```
koskimas: Isn't here a possibility for an SQL injection using the operator? ` > 10; DROP TABLE "Person"; --` You should check the operator against supported ones.

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+ * Json equality comparison.
+ *
+ * Also supports having field expression in both sides of equality.
+ *
+ * ```js
+ * Person
+ *   .query()
+ *   .whereJsonEquals('additionalData.myDogs', 'additionalData.dogsAtHome')
+ *   .then(function (person) {
+ *     // oh joy! these persons has all their dogs at home!
```
koskimas: Typo has --> have :D

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+ *
+ * ```js
+ * Person
+ *   .query()
+ *   .whereJsonSupersetOf('additionalData.myDogs', 'additionalData.dogsAtHome')
+ *   .then(function (person) {
+ *     // oh joy! these persons has all their dogs at home!
+ *     // But there might be actually some extra dogs there since
+ *     // we requested with subset
+ *   });
```
koskimas: Aren't the arguments in the wrong order? Usually these method are read by moving the "operand part" of the method name between the operands. For example
```
whereGreaterThan(a, b) --> a greaterThan b
```
So shouldn't this be
```
whereJsonSupersetOf('a', 'b') --> a supersetOf b
```

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+ *
+ * For arrays this mean that all arrays of left side matches if it has all the elements
+ * listed in the right hand side. e.g.
+ *
+ * [1,2,3] isSuperSetOf [2] => true
+ * [1,2,3] isSuperSetOf [2,1,3] => true
+ * [1,2,3] isSuperSetOf [2,null] => false
+ * [1,2,3] isSuperSetOf [] => true
+ *
+ * @param fieldExpression {String} Reference to column / jsonField, which is tested being supoerset.
```
koskimas: supoerset --> superset

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+QueryBuilder.prototype.whereJsonSupersetOf = function (fieldExpression, jsonObjectOrFieldExpression) {
+  return this.whereJsonObject(fieldExpression, "@>", jsonObjectOrFieldExpression);
+};
+
+/**
+ * Json filters all results where left hand operator is subset of the right hand operand.
+ *
+ * Object and array are always their own subsets.
+ *
+ * see {QueryBuilder.prototype.whereJsonSupersetOf}
```
koskimas: Did you test that this becomes a link by running `grunt jsdoc`? If this doesn't work you can use `{@link QueryBuilder.prototype.whereJsonSupersetOf}

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+ *
+ * @param fieldExpression {String}
+ * @param jsonObjectOrFieldExpression
```
koskimas: Type missing

src/QueryBuilder.js
```diff
@@ -1632,4 +1859,28 @@ function tryBuild(builder) {
...
+ *
+ * Table.jsonColumn:obj[key.with.dots][0].key.0.me[0] refers to
+ * obj = { "key.with.dots" : [{"key" : { "0": { me : [ "I was referred" ] }}}]
+ *
+ * Since PostgreSql #>{field,0,field2,...} operator does not make difference if
+ * reference is string or a number, one can actually use also jsonArray.0 notation
+ * to refer index of an array. Like wise one can use object[123] notation to refer
+ * key of an object { "123" : null }.
+ *
+ * @param expression
```
koskimas: Type missing

tests/integration/json-queries.js
```diff
@@ -0,0 +1,602 @@
+var _ = require('lodash');
+var expect = require('expect.js');
+var Promise = require('bluebird');
+var MoronModel = require('../../src/Model');
```
koskimas: You could rename this to `Model`

tests/integration/json-queries.js
```diff
@@ -0,0 +1,602 @@
...
+          .to.contain('"ModelJson"."jsonArray"');
+      });
+
+      it('should quote ModelJson.jsonArray:[10] column reference properly', function () {
+        expect(BoundModel.query().whereJsonIsArray("ModelJson.jsonArray:[50]").toString())
+          .to.contain('"ModelJson"."jsonArray"');
+      });
+    });
+
+    describe('.whereJsonObject(fieldExpr, operator, <array|object|string>)', function () {
```
koskimas: Only failing tests?

tests/unit/jsonFieldExpressionParser.js
```diff
@@ -0,0 +1,131 @@
...
+  testFail('col:wat.field[fa[il]');
+  testFail('col:wat.field[fa]il]');
+
+  // these should fail because inside quotes there is same type of
+  // quote => parser tries use stringWithoutSquareBrackets token
+  // for parsing => bracket in key fails parsing
+  testFail('col:field["fa"]il"]');
+  testFail("col:field['fa']il']");
+
+  describe("yanking up test coverage...", function () {
```
koskimas: Maybe this is not the best description :D.

src/QueryBuilder.js
```diff
@@ -3,7 +3,11 @@
 var _ = require('lodash')
   , Promise = require('bluebird')
   , RelationExpression = require('./RelationExpression')
-  , ValidationError = require('./ValidationError');
+  , ValidationError = require('./ValidationError')
+  , utils = require('./utils')
+  , jsonFieldExpressionParser = require('./jsonFieldExpressionParser')
+  , knex = require('knex')
```
koskimas: knex is not a dependency anymore. Use `this.modelClass().knex()` to get the knex object. The websql build will fail if you leave this here.

koskimas:
Excellent tests!

koskimas:
It is hard to find what a `fieldExpression` means. Maybe you could make a jsdoc `@typedef {String} FieldExpression` and put a thorough explanation there. Then you could annotate each field expression with the typedef like so `@param {FieldExpression} expression`

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+/**
+ * Json field filter.
+ *
+ * Supports having field expression in both sides of operand, but right hand value cannot
+ * be any other value, but only string reference to other column/json filed, json array
+ * or json object.
+ *
+ * Converts left and right hand values to PostgreSQL acceptable format and add user chosen
+ * operator between left and right hand expressions.
+ *
```
koskimas: Maybe some examples here?

src/QueryBuilder.js
```diff
@@ -1566,6 +1570,229 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+  return this.whereJsonObject(fieldExpression, "=", jsonObjectOrFieldExpression);
+};
+
+/**
+ * Json filters all results where left hand operator is superset of the right hand operand.
+ *
+ * ```js
+ * Person
+ *   .query()
+ *   .whereJsonSupersetOf('additionalData.myDogs', 'additionalData.dogsAtHome')
```
koskimas: Isn't `additionalData.myDogs` interpreted so that `additionalData` is the table name and `myDogs` the column? Shouldn't this be `additionalData:myDogs`?

tests/integration/json-queries.js
```diff
@@ -0,0 +1,602 @@
...
+      });
+
+      it('should be able to find is not null', function () {
+        return BoundModel.query().whereJsonField("jsonObject:stringField", 'IS NOT', null)
+          .then(function (results) {
+            expectIdsEqual(results, [1]);
+          });
+      });
+
+      it('should be able to find boolean equals true', function () {
```
koskimas: boolean equals false?

src/QueryBuilder.js
elhigu: Yep, I thought about that, but then thought that nobody allows user to type operator... which actually might happen. 

However I should make this function to be private and remove it from public API, since it was produced only because I refactored .whereJsonEquals, .whereJsonSupersetOf and .whereJsonSubsetOf implementations to one helper.

If we need later on some this generic function to public API it should be named better and be yet more generic than this.

src/QueryBuilder.js
elhigu: Yep, forgot to fix examples when extended fileld parset.

src/QueryBuilder.js
elhigu: That was the intention that `firstArg is superset of secondArg`. Looks like that my example is in wrong place.

tests/integration/json-queries.js
elhigu: All successful tests of this is divided to `.whereJsonEquals`, `.whereJsonSupersetOf` and `.whereJsonSubsetOf`. This was also created when refactoring functionality and didn't want to test failures in every derived method separately. Looks like I'm also calling wrong function in test implementations `whereJsonEquals ` instead of `whereJsonObject` :)

tests/unit/jsonFieldExpressionParser.js
elhigu: well that was the truth :D since that parser is automatically generated and those features are never used in real code. should I change it to "field expression parser options" ?

elhigu:
Not ready yet, still in fixing review comments... slowly :)

elhigu:
> You can exclude stuff from coverage report using instanbul's annotations.

Last 0.13% decrease comes from generated parser code, where I wouldn't like to put any istanbul comments manually.

src/QueryBuilder.js
```diff
@@ -1566,6 +1568,191 @@ QueryBuilder.prototype.returning = queryMethod('returning');
...
+    cast = "::BOOLEAN";
+  } else if (_.isString(value)) {
+    cast = "::TEXT";
+  } else if (_.isNull(value)) {
+    cast = "::TEXT";
+    escapedValue = 'NULL';
+  } else {
+    throw new Error("Value must be string, number, boolean or null.");
+  }
+  return this.whereRaw(["(", fieldReference, ")", cast, " ", operator," ", escapedValue].join(""));
```
elhigu: This is actually public API function which takes operator in... need to check from knex, how operator validation is done there.

elhigu: And now that I think of all operators this might get, maybe I should also allow givin an array to right hand operand, e.g. for `IN` operator.

elhigu: Maybe that `IN` support needs to be added later... it will need some designing first, since casting in that case is not that straight forward. I'll add separate feature issue about it.

elhigu:
All fixed! Anything else to change?

koskimas:
Looks great!

PR #44, opened 2015-11-12
objection.transaction.start, commit and rollback for transaction

nbeydoon:
Easier way to run transactions

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -125,8 +125,8 @@ function transaction() {
   });
 };
 
-module.exports.start = function start(modelBase) {
-  if (arguments.length < 1) {
+transaction.start = function start(modelBase) {
+  if (!modelBase) {
     return Promise.reject(new Error('objection.transaction.start: provide at least one Model class to bind to the transaction'))
   }
 
@@ -136,22 +136,19 @@ module.exports.start = function start(modelBase) {
 
   var knex = modelBase.knex();
 
-
-  return new Promise(function(resolve) {
+  return new Promise(function(resolve, reject) {
     knex.transaction(function (trx) {
         modelBase = modelBase.knex(trx);
 
       resolve({
-        __trx: trx,
         commit: function() {
-          return this.__trx.commit();
+          return trx.commit();
         },
         rollback: function() {
-          return this.__trx.rollback();
+          return trx.rollback();
         }
       });
-    });
-
+    }).catch(reject);
   });
 }
 
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -82,7 +82,10 @@ var utils = require('./utils');
  * @function transaction
  * @returns {Promise}
  */
-module.exports = function transaction() {
+module.exports = transaction;
+
+
+function transaction() {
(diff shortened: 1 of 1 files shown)


nbeydoon:
Erf you check the coverage 0 :(

nbeydoon:
Need more work

lib/transaction.js
```diff
@@ -122,6 +125,33 @@ module.exports = function transaction() {
...
+
+  if (!utils.isSubclassOf(modelBase, Model)) {
+    return Promise.reject(new Error('objection.transaction.start: all but the last argument should be Model subclasses'));
+  }
+
+  var knex = modelBase.knex();
+
+  return new Promise(function(resolve, reject) {
+    knex.transaction(function (trx) {
+        modelBase = modelBase.knex(trx);
```
koskimas: This sets the the transaction globally for the model. Consider the case that 100 requests are currently running on your node server that use the `modelBase` class. You set the knex connection of _all_ of them here. Very bad things will happen. This line is actually useless.

lib/transaction.js
```diff
@@ -122,6 +125,33 @@ module.exports = function transaction() {
...
+
+      resolve({
```
koskimas: You resolve the promise with a custom object. You cannot bind models to this object using `bindKnex`. You can just resolve the promise with the `trx` object which is itself a knex instance.

lib/transaction.js
```diff
@@ -122,6 +125,33 @@ module.exports = function transaction() {
...
+        commit: function() {
+          return trx.commit();
+        },
+        rollback: function() {
+          return trx.rollback();
+        }
+      });
+    }).catch(reject);
```
koskimas: Passing the reject method for `cathch` here never does anything. To use the transaction the promise has to already be resolved (with the transaction object). The resolved promise cannot change its state. Therefore the `catch(reject)` does nothing.

nbeydoon: So when I run without the catch One 1 / 2 time i get timer error from bluebird, but I agree I don 't understand why

lib/transaction.js
nbeydoon: Yes I know, I did it like this for my test, and I have wait your review, I think is possible to do similar think like the ````.transaction```` method.

koskimas:
I just pushed a commit to master that should solve this.

nbeydoon:
Yes, it works great but so like this all knex are binded or not?

nbeydoon:
I works great! :)

koskimas:
No, only the knex that is returned by the start function. If you really want to bind all the models to the same transaction (very dangerous and extremely error-prone) you can just say `objection.Model.knex(transaction)`. Note that the transaction object is just like any knex instance. It is just bound to a single connection that has the `BEGIN` command ran.

nbeydoon:
Ok I see, Its ok now I can do all my test and you have a new cool features for objection, very thanks :)

koskimas:
Thank you! This is a cool feature. I just never needed it myself and didn't think of it. Keep the suggestions coming and open issues whenever you have a problem.

PR #65, opened 2016-01-11
Jsonb api example

elhigu:


kapouer:
Note that postgres 9.5 supports concatenation of jsonb objects, which allows one to update a jsonb field much more easily. This is a useless comment, though :)

elhigu:
Yep! Also `jsonb_set` is really nice in 9.5.

Issue #85, opened 2016-02-17
How to specify database schemas for different models?


ArsalanDotMe:
I've been at it for hours and still can't get it to work. I have a postgres database with two schemas, lets say s1 and s2 such that my table names are then s1.name or s2.name.

Now, if I rewrite all tablename references with schema included, objection relations stop working saying that I must specify the original table in either the from or to clause. 
I have also tried subclassing Model and overriding the query method with something like this:
```javascript
class ApiModel extends Model {
}
ApiModel.query = function query() {
  return Model.query.call(this).withSchema('s1');
};
```
But when I analyze the queries, objection is still making queries without the schema applied to table names. How do I handle default schemas for my models?

koskimas:
So you have two schemas and tables that only exist in one but not the other and you want to create models that always use a certain schema? And there are relations between the models so that `Schema1Model` can have, lets say, `HasOneRelation` to `Schema2Model`?

What version of objection are you using?

ArsalanDotMe:
There isn't currently a case where a model in Schema1 may have a relation with Schema2. In my case, I have a database shared by several internal applications, each storing its data inside its own schema. Models for each application will use a certain schema and will have relationships with other models of the same application on the same schema.

koskimas:
Ok. Can you check the objection version also?

ArsalanDotMe:
The objection version is 0.4.0.

koskimas:
Can you try to install `npm install objection@next` and check if the problems continue? There have just been some major changes to how `withSchema` works in the 0.5.0 alpha builds.

ArsalanDotMe:
Okay but how should I specify my schemas for my models?
As far as I understand, I have two options:

1. Override the Model class in a way that applies `withSchema` to the `QueryBuilder`.
2. Specify the schema inside tablenames. E.g. 
```javascript
class MyModel extends Model {
   static get tableName() { return 'schema.tablename'; }
}
```

Which way should I go with? I'm leaning towards the first but I'm not sure how to accomplish it.

koskimas:
Setting the schema to the tableName doesn't work. knex doesn't interpret the dot as a schema separator, but as a part of the table name.

You can try implementing your own query builder class:

```js
class DefaultSchemaQueryBuilder extends objection.QueryBuilder {
  constructor(modelClass) {
    super(modelClass);
    if (modelClass.defaultSchema) {
      this.withSchema(defaultSchema);
    }
  }
}

[...]

PR #86, opened 2016-02-17
add schema support in models

ArsalanDotMe:
This simple commit adds ability to override the default schema in queries. 
Define schemaName in your model like this:
```javascript
class MyModel extends Model {
  static get schemaName() { return 'api'; }
}
```


ArsalanDotMe:
solution for supporting schemas can be implemented without this pull request. See the discussion here: #85

PR #107, opened 2016-04-02
fix for relation definitions of tables defined in schemas

J-A-B-R:
Hi there, I was evaluating your awesome library when I stumbled upon a conflict between Objection relations and [PostgreSQL schemas](http://www.postgresql.org/docs/current/static/ddl-schemas.html).
With PostgreSQL schemas, a table named _my_table_ defined under the schema named _my_schema_ would be addressed as _my_schema.my_table_.

This PR (my first one, by the way) fixes a bug that when you are defining the _from_ and _to_ columns in a relation like
```
...
    join: {
        from: 'schema_name.model_name.column_name',
...
```
Objection interprets _schema_name_ as the table name and _model_name as_ the column name, discarding the rest.

I've added a test to the _Relation_ test suite to verify this fix and checked that all tests pass. However, I've observed that the test _"integration tests mysql Model find queries .$relatedQuery() many to many relation should return all related rows when no knex methods are chained"_ sometimes fails and sometimes not, but it seems that is not related to this PR.

Hope this helps.

### DIFF
--- a/src/relations/Relation.js
+++ b/src/relations/Relation.js
@@ -547,9 +547,10 @@ export default class Relation {
     let columns = [];
 
     for (let i = 0; i < ref.length; ++i) {
-      let parts = ref[i].split('.');
-      let tableName = parts[0] && parts[0].trim();
-      let columnName = parts[1] && parts[1].trim();
+      const refItem = ref[i];
+      const ndx = refItem.lastIndexOf('.');
+      let tableName = refItem.substr(0, ndx).trim();
+      let columnName = refItem.substr(ndx + 1, refItem.length).trim();
 
       if (!tableName || (table && table !== tableName) || !columnName) {
         return {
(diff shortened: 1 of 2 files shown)


koskimas:
Thank you for fixing this! I have zero tests for defining schema in `tableName` or `relationMappings` so you may run into other bugs also. You could consider using an approach discussed [here](https://github.com/Vincit/objection.js/issues/85#issuecomment-185183032) if things don't work.

I just fixed a bunch of randomly failing mysql tests. As the ones I fixed, the one you mentioned probably also is caused by wrong  assumptions about the order of the query results.

J-A-B-R:
Thanks for the suggestion about the query builder. I'm planning to use Objection on a project with plenty of schemas, so if there are more related bugs, certainly they will show up.
Keep up this great work.

PR #120, opened 2016-05-11
Additional Query Builder Methods Added

ozum:
Some public methods available in Knex' query builder seems missing from objection's query builder.

`increment`, `decrement`, `pluck`, `first`, `modify`, `options`, `columnInfo` methods added.

Best Regards,

### DIFF
--- a/src/queryBuilder/QueryBuilderBase.js
+++ b/src/queryBuilder/QueryBuilderBase.js
@@ -632,6 +632,48 @@ export default class QueryBuilderBase {
    * @returns {QueryBuilderBase}
    */
   @queryBuilderOperation(KnexOperation)
+  increment(...args) {}
+  
+  /**
+   * @returns {QueryBuilderBase}
+   */
+  @queryBuilderOperation(KnexOperation)
+  decrement(...args) {}
+
+  /**
+   * @returns {QueryBuilderBase}
+   */
+  @queryBuilderOperation(KnexOperation)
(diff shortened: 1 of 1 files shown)


koskimas:
Thanks for this, but most of those methods already exists. They are defined in the QueryBuilder.js instead of QueryBuilderBase.js since they have custom implementation. Also the `modify` method doesn't work the way you implemented it. I'm currently working on it though. I'll add the methods today.

ozum:
Strange. I downloaded 0.5.0-rc, added `modify` method as in my PR, then installed via `npm install`, `npm run build`.

I'm using following code without problem. Is there something I missed?

```js
'use strict';

const Knex = require('knex');
const knex = Knex({ client: 'pg', debug: true, connection: { host: 'localhost', user: 'user', password: 'password', database: 'database' } });

var objection = require('./lib/objection.js');
var Model = objection.Model;

Model.knex(knex);

class Org extends Model {
  static get tableName() { return 'Organization'; }
}

function deneme(queryBuilder) {
    queryBuilder.where('id', 2);
}

Org.query().modify(deneme).where('id', 1).then((orgs) => {
    console.log(orgs);
}).catch(err => console.log(err));
```

Nonetheless thanks for your attention.

Best Regards,

koskimas:
There are problems with passing extra arguments to the callback and other corner-case bugs.

ozum:
Thank you.

PR #141, opened 2016-06-06
Bumped up maximum knex version

rafaljanicki:
Current max version of knex is 11.5, changes are not big to not include this


rafaljanicki:
Sorry, my bad, I haven't noticed that it's < 11, without <=

I'll try to create a PR for the newest knex version

PR #169, opened 2016-07-22
Use babel preset instead of explicit plugin installation

afm-sayem:
Instead of explicitly declaring each es2015 plugins, a preset can be used that contains all the plugins necessary for compiling to es2015.


afm-sayem:
Build is failing for node v0.12. Is anyone using v0.12 to build objection.js?

koskimas:
Is there a reason for this other than simpler package.json? The plugins are listed separately for two reasons.

1. I'm only including plugins I use.
2. I'm passing extra options for some plugins.

The first reason is not important, but the second one is. I'm telling Babel to generate faster inheritance code that ignores some corner cases. If this setting still works, I will merge this.

afm-sayem:
I have checked this on node v6.0 and was able to build using `build.js` and run unit and integration tests. I just ran a diff between generated code pre and post patch and there was no change.

koskimas:
All tests need to pass. I'm not dropping three versions for this tiny change.

afm-sayem:
Far I understand, this doesn't impact anyone installing the library since the changes are in dev-dependencies. If maintaining build in pre node v5 is important then I don't think this should be merged.

koskimas:
Keeping the travis tests working pre 0.12 is important. Also you haven't yet told me the reason for this change. What does this fix?

afm-sayem:
This simplifies package.json and the build file. Not a fix.

Since travis tests will fail for the changes made. I'm closing this.

PR #207, opened 2016-10-11
knex dependency bump to 0.12.3

### LINKED ISSUE #1703: Many to Many Extra property parsing - more informative error handling.
Using Objection 2.1.3

This was a bug in my code, but the way Objection handles `extra` properties in `ManyToManyRelation`s made it hard to track down.

Per the docs, `extra` is expected to be an array, even if it's an array of length 1, but if we look at https://github.com/Vincit/objection.js/blob/master/lib/relations/manyToMany/ManyToManyRelation.js#L413 we see that if it's not an array it gets treated as an object and there's some extra handling that happens.

The problem I ran into was that if your relation definition has `extra: 'myExtra'` instead of `extra: ['myExtra']` Objection will happily consume `'myExtra'` and treat the letters of the string as individual extra elements(Object.keys() on a string will convert the string into an array of chars), so it tries to fetch "Table"."e", "Table"."x", "Table"."t"... and you wind up with an error and a very confusing looking query.

It seems like a string literal instead of an array could be treated as an error here to prevent others from shooting themselves in the foot as I did. Happy to take a stab at a PR, but I don't understand the use case represented by the object processing, so I'm not sure if a simple check is going to do the trick.

### LINKED ISSUE #1710: $beforeDelete hook not getting called
Hi There,

I'm running into issues getting the `$beforeDelete` hook to fire on an instance of my models. I've followed the guides making sure to call the instance and am also using the document [here](https://github.com/Vincit/objection.js/blob/3fcb87b59ea423bf69d114561c18a3a4ba429eef/doc/guide/hooks.md) as a reference.

in my controller, I have the following:

```
// controllers/user-cards-controller.js
let card = await Card.query().findById(req.params.id)
await card.$query().delete()
res.status(200).send({})
```

Then in my model...

```
// models/card.js
  async $beforeDelete(context) {
    await super.$beforeDelete(context);
    await this.destroyStripeCard();
    this.id ? this.scheduleWorkers('deleted') : null;
  }
```

I haven't had any difficulty with the other hooks nor objection otherwise. I'm also currently using Objection 2.1.3. Any help would be appreciated!

fl0w:
Bumps knexjs/knex to 0.12.3. Looks to be harmless and fixes a few annoyances which even @koskimas raised.

From knex CHANGELOG:

```
0.12.3 — 9 Oct, 2016
Fix #1703, #1694 - connections should be returned to pool if acquireConnectionTimeout is triggered
Fix #1710 regression in postgres array escaping

0.12.2 — 27 Sep, 2016
Restore pool min: 1 for sqlite3, #1701
Fix for connection error after it's closed / released, #1691
Fix oracle prefetchRowCount setting, #1675
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "lodash": "^4.7.0"
   },
   "peerDependencies": {
-    "knex": ">= 0.8.0 <= 0.12.1"
+    "knex": ">= 0.8.0 <= 0.12.3"
   },
   "devDependencies": {
     "babel-core": "^6.3.17",
@@ -84,7 +84,7 @@
     "grunt": "^0.4.5",
     "grunt-jsdoc": "^1.0.0",
     "istanbul": "^0.4.0",
-    "knex": "0.12.1",
+    "knex": "0.12.3",
     "mocha": "^2.3.4",
     "mysql": "^2.7.0",
     "pg": "^4.4.0",


fl0w:
Oops, shouldn't be in `dependencies`.

PR #210, opened 2016-10-11
[WIP] add custom jsonSchema string formats, fixes #199

### LINKED ISSUE #199: Json Schema Custom Formats
This is a feature request to add custom format definitions for the JSON Schema validation performed by Ajv. It makes it much easier to insert custom validations on per-field basis while still making use of the JSON Schema. I could do this by overriding `$validate`, but it would basically require me to redo all the work Objection is already doing, which seems like waste.

With tv4 I could do this by importing tv4 into my own library and adding them there, but Ajv doesn't  have a way to statically add formats, so I would need access to the instance. The easiest way would be to store the `ajv` instance on ModelBase, but that is probably weird from an API standpoint. It would also be trivial to add a `addSchemaFormat` or `addCustomValidator` function to ModelBase that looks like

```
// I'm used to TypeScript, so that's how I'm going to annotate this, I hope that's clear.
addCustomValidator(name: string, validator: (value: any) => boolean): void {
  ajv.addFormat(name, validator);
} 
```

This could reasonably well separate your api from theirs so if you change schema validators again this function can be adjusted, unless you switched to one that doesn't handle custom formats at all.

I understand if you reject this outright, since it's already possible. I only suggest it because it's a very simple change that your underlying tools already support. And my way of doing this would likely look just like what you're already doing, just so I have access to the `ajv` variable, which seems like overkill.
#### Edit

I somehow managed to submit this before I was done writing it, so I closed it and finish before re-opening, sorry about that.

fl0w:
This is a WIP as I haven't gotten to do the docs ...
Or actually, I did - but I destroyed the document somehow and I had to reset it.

Custom string formats would be added as expected by `ajv`:

``` js
// @returns void
objection.Model.addCustomFormat(string, /some-regex/);
objection.Model.addCustomFormat(string, { /* directly passed to ajv */ });
```

Currently this is somewhat at mercy of `ajv`, this maybe isn't as non opinionated as you'd like? 

I also exposed the internal `ajv` instance because I want/need to [add schema keywords](https://github.com/epoberezkin/ajv#addkeywordstring-keyword-object-definition) I opted not to add that to `ModelBase` yet as I'm not entirely sure how stable that spec is (by my understanding its a v5 proposal, implemented in `ajv` but not yet standardised)

``` js
// returns the ajv instance
objection.Model.schemaValidator().addKeyword(/* ... */)
```

My first implementation was a bit more complicated, which I reverted.
I had an idea of creating an objection.Validator class which basically acts as an interface and had the same methods implemented as used today - but I wasn't sure this was the direction you'd want to take it and I'm to noob actually draft a spec like that.

### DIFF
--- a/src/model/ModelBase.js
+++ b/src/model/ModelBase.js
@@ -413,6 +413,14 @@ export default class ModelBase {
   }
 
   /**
+   * @param {String} name
+   * @param {String|RegExp|Function|Object} format
+   */
+  static addCustomFormat(name, format) {
+    ajv.addFormat(name, format);
+  }
+
+  /**
    * @param {Object} jsonSchema
    * @param {boolean} skipRequired
    * @returns {function}
@@ -761,3 +769,4 @@ function compileJsonSchemaValidator(jsonSchema, skipRequired) {
     }
   }
 }
+
--- a/tests/unit/model/ModelBase.js
+++ b/tests/unit/model/ModelBase.js
@@ -374,6 +374,24 @@ describe('ModelBase', function () {
       expect(model.c).not.to.equal(obj);
     });
 
+    it('should validate custom string formats', function () {
+      Model.addCustomFormat('se-ssn', /^[12]{1}[90]{1}[0-9]{6}-[0-9]{4}$/)
+
+      Model.jsonSchema = {
+        properties: {
+          a: {type: 'string', format: 'se-ssn'}
+        }
+      };
+
+      var model = Model.fromJson({a: '19851007-1234'});
+
+      expect(model.a).to.equal('19851007-1234');
+
+      expect(function () {
+        Model.fromJson({a: '198510071234'})
+      }).to.throwException()
+    });
+
     // regression introduced in 0.6
     // https://github.com/Vincit/objection.js/issues/205
     it('should not throw TypeError when jsonSchema.properties == undefined', function () {


fl0w:
Updated/rebased because `const` isn't allowed in es5 strict.

koskimas:
Exposing the validator will couple objection with ajv and. If we decide to change the underlying validator at some point, code will break. `ajv. addKeyword` seems very specific to ajv and exposing a `ModelBase.addKeyword` method with the same signature would probably not be portable either.

I suppose that only option is to come up with an interface that probably will be portable to other validators (not an easy task) or simply couple objection with ajv. I don't know which is better.

src/model/ModelBase.js
```diff
@@ -413,6 +413,22 @@ export default class ModelBase {
   }
 
   /**
+   * @returns {Object}
+   */
+  @memoize
```
koskimas: memoize is not needed here since we simply return a value. Memoizing this will only make it slower

src/model/ModelBase.js
```diff
@@ -745,6 +761,14 @@ function contains(arr, value) {
   return false;
 }
 
+function getSchemaValidatorInstance() {
```
koskimas: What is the purpose of this function?

src/model/ModelBase.js
fl0w: Fixed.

src/model/ModelBase.js
fl0w: This is a relic from my first implementation, I'll remove it.

fl0w:
All things considered; I removed clutter now. I was way to tired yesterday it seems.
Rebased and updated according to comments.
Removed return of ajv instance for now - unrelated to issue.

fl0w:
@koskimas addKeyword is actually [implemented in tv4](https://www.npmjs.com/package/tv4#definekeywordkeyword-validationfunction) as well. It is really just a [JSON Schema extend](http://json-schema.org/latest/json-schema-core.html#anchor20).

If I'm reading this correctly, we could expose an `.extendSchema()` and objection would still actually be spec compliant. Implementations will differ however. E.g. in `ajv` one can define `$async`, while tv4 does not have an equivalent.

Would it be stupid to extract ajv instantiation, and if/when userspace wants to hack a simple `require('objection/ajv')` or `const { Ajv } = require('objection')` would be sufficient?

koskimas:
Your original idea with something like `objection.Validator` got me thinking and actually I think that is the way to go! It would also solve the problem of providing a way to replace json schema validation with some other validation system. (it can be done by overriding $validate but it is nasty to implement)

Instead of creating an API to objection, we would simply need to create a overridable factory method that would create a `Validator` instance which people could implement. We should provide a default `AjvValidator` implementation that people could then use like this:

```js
class SomeModel extends Model {
  static createValidator() {
    const ajv = new Ajv();
    addYourKeywordsAndWhatnot(ajv);
    return new AjvValidator(ajv);
  }
}
```

The default implementation would be:

```js
class SomeModel extends Model {
  static createValidator() {
    return new Ajv({allErrors: true, validateSchema: false, ownProperties: true});
  }
}
```

The `Validator` interface could be as simple as this:

```js
class Validator {
  validate(ModelClass, jsonObj, modelOptions) {
    // Throw whatever error if validation fails.
    return jsonObj;
  }
}
```

Most of the validator related stuff would need to be moved from the `ModelBase` to `AjvValidator` (which is a great thing). All the validator caching stuff and even the default value merging. After the refactor, `ModelBase.$validate` should look something like this:

```js
  $validate(json = this, options = {}) {
    let jsonSchema = this.constructor.getJsonSchema();

    if (options.skipValidation) {
      return json;
    }

    // No need to call $beforeValidate (and clone the jsonSchema)
    // if $beforeValidate has not been overwritten.
    if (this.$beforeValidate !== ModelBase.prototype.$beforeValidate) {
      jsonSchema = _.cloneDeep(jsonSchema);
      jsonSchema = this.$beforeValidate(jsonSchema, json, options);
    }

    json = this.constructor.getValidator().validate(this.constructor, json, options) || json;

    this.$afterValidate(json, options);
    return json;
  }
```

After this `jsonSchema` should be 100% optional.

The `getValidator` method can be a memoized so so that the `createValidator` is only called once.

What do you think?

fl0w:
It's funny, because that's remarkably close to what I initially intended. However, yours is way simpler *(as in elegant)*. This also explains why I accidentally left `getSchemaValidatorInstance` (equivalent to your getValidator()).

Internally this should be a small change, but as a consumer (user space) this is huge because except for lifecycle callbacks ($before/afterValidate) it's quite hard to practically break user space with new objection releases.

I like it. I can do a trial implementation according to your plan if you want?

koskimas:
Then this is basically your plan 😄 If you have time to implement this it would be great! Nice to have some help.

koskimas:
Maybe we should also expose the `Ajv` module through `AjvValidator.Ajv` or something so that people don't need to add `ajv` to their package.json and possibly adding a second copy of ajv if the versions don't match.

fl0w:
I'll take a look at it tonight.

Did you intend AjvValidator to be an external package, e.g. `Vincit/objection-ajv-validator`?

Would it be wrong to expose the instance through the interface? Or would that maybe get to ambiguous, and perhaps add confusion when switching packages.

```js
class Validator {
  // ...
  getInstance() {}
  // ...
}
```

koskimas:
The interface should not provide the instance. Some validators may not even have an "instance". A validator may be simply a function or a bunch of functions. Also objection should not use the `getInstance` method since then it would again need to know that object's interface to use it aaand we're married to that interface once again.

This would be ok:

```js
class AjvValidator {
  getInstance() {}
}
```

Did you have some use case in mind for the `getInstance` method?

koskimas:
Off course, with javascript this is all just a documentation issue. There are no private methods or interfaces, and effectively the result will be the same. But I'm talking about the interface that is documented and users need to implement.

fl0w:
Of course, this is a documentation thing - I'm just using pseudo code. I'll draft this up and we can take it from there - easier to correct when there's something to point at I think.

FYI I might be AFK this evening it seems so if there's no PR today, expect it tomorrow.

koskimas:
There is no rush. I probably don't have time for objection this weekend.

koskimas:
I created a 0.6.x branch since master now has breaking changes planned for 0.7 release. This pull request should be merged to 0.6.x when you are done.

fl0w:
Alright, I'll rebase it. I might actually do a separate PR for this. I'm sorry though, I had a weekend of fun planned, alas it wasn't. I'll grab this post work today (GMT+1).

rafaljanicki:
Thanks for answer @koskimas 

I've seen this PR, but thought it's closed and that's why I've created the new ticket. Basically, I was thinking about adding the ajv-keywords library which is created by the same author. Having the library allows us to specify flags for the regexp.

fl0w:
@rafaljanicki you will be able to do this within your app as soon as possible (I've been delaying on this). JSON Schema keywords are in essence an JSON Schema extend which, while the act of extending is supported - the specifics of implementation and open (i.e. not standardised). Thus, as far as we (me and @koskimas) currently agree it shouldn't (couldn't) be supported by `objection.js` while also retaining a validator marriage.

koskimas:
@fl0w Any progress on this?

koskimas:
@fl0w Do you still have time for this? I can take this off your hands if you want?

fl0w:
@koskimas I've been sick - sorry 'bout this. Just got enough strength to look at a screen again. I've got some work related stuff to catch up on, can objection on again on Sunday, unless you want to take it to get this done?

koskimas:
I have some time now so I'll implement this myself.

pwalczyszyn:
I wanted to check what's a status of this one?

koskimas:
@pwalczyszyn Implemented in 0.7 rc1. You can install it using `npm install objection@next`. You can find the 0.7 documentation [here](https://github.com/Vincit/objection.js/tree/master/doc) in a slightly crappier format 😄

koskimas:
https://github.com/Vincit/objection.js/blob/master/doc/includes/API.md#createvalidator
https://github.com/Vincit/objection.js/blob/master/doc/includes/API.md#ajvvalidator

Issue #102, opened 2016-03-26
Any plans for a .d.ts file for TypeScript projects?


dotDeeka:
First off this is a great project, it wraps knex just enough.

I am using this project in a project written in TypeScript. Is there a definition file or plans to create a definition file for the Objection.js library? Thanks.

koskimas:
We are using typescript in one of our projects at Vincit, and we have written some typings for that project, but we are not any good at it :smile:. If you have, or want to write, the typings I would be happy to accept a pull request.

dotDeeka:
Ha, I am not any good at it either. I've started one here https://github.com/dotDeeka/objection-d-ts that I can reference in my project. If I actually get it right I will certainly submit a PR for it.

albertkim:
Hoping to use this in a Typescript project one day, will keep an eye on this issue! Would go along well with the amazing documentation so far.

zam6ak:
The typings, if and when added should be referenced from project.json
see: http://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html


also this may be of interest:
https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/

juhoojala:
Thanks for this great project! 

Any progress on this issue?

koskimas:
Not on my part. @dotDeeka ?

dotDeeka:
No not really @koskimas . I have that repo in my account I've been using, I'm pretty sure its no where near complete. If anyone out there is interested in trying to make it more complete I'd certainly accept a PR, or even give what I have to someone if they're serious about owning it?

danieloprado:
+1

4nxiety:
Hello,

is there some new news about it?

elhigu:
@4nxiety There hasn't been anyone willing to own that part and keep updating typings when project evolves.

4nxiety:
oh damn :/

mceachen:
@dotDeeka, @koskimas: my PR that adds typings, #239, is ready for review.

mceachen:
Fixed by #255

dotDeeka:
Awesome @mceachen, thanks for stepping up on this.

mceachen:
#272 is merged to master. @koskimas can you close this issue?

dotDeeka:
Any chance we'll get this as part of 0.7 ?

PR #239, opened 2016-11-04
Added TypeScript 2 example project

mceachen:
I've added a partial typings for Objection and an example project that consumes the typings in the PR to address #102. I've got a solid 3 weeks of experience with TypeScript at this point, so temper expectations accordingly.

This PR compiles and runs successfully on node v6.8.0 on linux and win10.

As this typing is self-contained in the examples directory, you can merge this PR and not adversely impact anyone except others that are interested in pushing forward high-quality typings for Objection.

Remaining tasks after accepting this PR:

- [ ] Submit this to [DefinitelyTyped ](http://definitelytyped.org/guides/contributing.html). 
- [ ] Remove the typings directory and add the new types to `examples/express-ts/package.json`.

### DIFF
--- a/examples/express-ts/typings/objection/index.d.ts
+++ b/examples/express-ts/typings/objection/index.d.ts
@@ -1,0 +1,230 @@
+// Type definitions for objection v0.6.1
+// Project: Objection.js <http://vincit.github.io/objection.js/>
+// Definitions by: [~YOUR NAME~] <[~A URL FOR YOU~]>
+
+declare module "objection" {
+
+  import * as knex from 'knex';
+  import { JsonSchema } from 'jsonschema';
+
+  export interface ModelOptions {
+    patch: boolean;
+    skipValidation: boolean;
+  }
+
+  export interface ValidationError {
+    statusCode: number;
+    data: any;
+  }
+
+  export type RelationMappings = { [relationName: string]: RelationMapping }
+
+  interface Relation {
+    // TODO should this be something other than a tagging interface?
+  }
+
+  export interface RelationJoin {
+    from: string | string[];
+    to: string | string[];
+    through?: RelationThrough;
+  }
+
+  export interface RelationThrough {
+    from: string | string[];
+    to: string | string[];
+    modelClass?: string | typeof Model;
+    extra?: string[];
+  }
+
+  export interface RelationMapping {
+    relation: Relation;
+    modelClass: typeof Model | String;
+    join: RelationJoin;
+    modify?: (queryBuilder: QueryBuilder) => {};
+    filter?: (queryBuilder: QueryBuilder) => {};
+  }
+
+  export interface EagerAlgorithm {
+    // TODO should this be something other than a tagging interface?
+  }
+
+  export interface EagerOptions {
+    minimize?: boolean;
+    separator?: string;
+    aliases?: string[];
+  }
+
+  /**
+   * This is a hack to make bindKnex return subclasses
+   * See https://github.com/Microsoft/TypeScript/issues/5863#issuecomment-242782664
+   */
+  interface ModelClass<T extends Model> {
+    new (...a: any[]): T;
+    bindKnex<T extends Model>(this: ModelClass<T>, knex: knex): T;
+  }
+
+
+  export class Model {
+    static tableName: string;
+    static jsonSchema?: JsonSchema;
+    static idColumn?: string;
+    static modelPaths?: string[];
+    static relationMappings?: RelationMappings;
+    static jsonAttributes?: string[];
+    static virtualAttributes?: string[];
+    static uidProp: string;
+    static uidRefProp: string;
+    static dbRefProp: string;
+    static propRefRegex: RegExp;
+    static pickJsonSchemaProperties: boolean;
+    static defaultEagerAlgorithm?: EagerAlgorithm
+    static defaultEagerOptions?: EagerOptions;
+    static QueryBuilder: typeof QueryBuilder;
+    static RelatedQueryBuilder: typeof QueryBuilder;
+
+    static raw: knex.RawBuilder
+    static fn: knex.FunctionHelper
+
+    static BelongsToOneRelation: Relation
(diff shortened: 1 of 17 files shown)


drew-r:
Thanks @mceachen 
How is `bindKnex()` intended to be used with these type defs?

mceachen:
@drew-r that's a mistake--I misread the API guideline, and it should return a typeof Model (or in this case, `typeof T`, but that [doesn't look like it's supported](https://github.com/Microsoft/TypeScript/issues/204). 

I'll look into it a more tomorrow.

drew-r:
I fixed it and a few other things, whats the best way to get them merged?

mceachen:
@drew-r where's your diff or branch?

mceachen:
I believe subclasses are properly handed in 0bcad79 now, so `Model` types are ready. 

`QueryBuilder` is still a WIP. I'll submit another diff finishing that work soon.

examples/express-ts/typings/objection/index.d.ts
```diff
@@ -0,0 +1,692 @@
+// Type definitions for objection v0.6.1
+// Project: Objection.js <http://vincit.github.io/objection.js/>
+// Definitions by: [~YOUR NAME~] <[~A URL FOR YOU~]>
```
elhigu: Your name and url missing

examples/express-ts/typings/objection/index.d.ts
```diff
@@ -0,0 +1,692 @@
...
+  type ModelsOrObjects = Model | Object | Model[] | Object[];
+
+  type RelationOptions = { alias: boolean | string };
+
+  interface JoinRelation {
+    (relationName: string, opt?: RelationOptions): QueryBuilder;
+  }
+
+  interface WhereJson {
+    (fieldExpression: FieldExpression, jsonObjectOrFieldExpression: any): QueryBuilder;
```
elhigu: Why here `jsonObjectOrFieldExpression` is `any` and not like below?

examples/express-ts/typings/objection/index.d.ts
```diff
@@ -0,0 +1,692 @@
...
+
+    deleteById(idOrIds: IdOrIds): QueryBuilder;
+
+    relate(ids: IdOrIds | ModelsOrObjects): QueryBuilder;
+    unrelate(): QueryBuilder;
+
+    forUpdate(): QueryBuilder;
+    forShare(): QueryBuilder;
+
+    // TODO: fromJS does not exist in current knex documentation: http://knexjs.org/#Builder-fromJS
```
elhigu: I think from JS was removed from knex at some point... I had never used it either.

drew-r:
@mceachen I wonder is there a way to have QueryBuilder return Promises that are typed to the model?
I played around with a `QueryBuilder<T>` and a `query(this: T): QueryBuilder<T>` but it kind of breaks down when you have to start subclassing models for example with `bindKnex`, as `this: T` is a `typeof Model` rather than `Model`. Any ideas?

mceachen:
@drew-r I played with that a bit, but because you can make queries that
return partial models,  model unions, and even arbitrary columns, that
discouraged me from pushing forward with that idea.

If there was a query builder with a much simpler interface, I think that'd
be great. The current query builder would presumably then subclass and
extend this strongly typed query builder.

@elhigu is there anything else I can do to get this PR merged? Thanks for
your prior review!

danieloprado:
A simple solution is type the return function:

```typescript
export async function list(churchId: number): Promise<Informative[]> {
  return await Informative.query().where({ churchId });
}

export async function insert(informative: IInformative): Promise<Informative> {
  return await Informative.query().insert(informative);
}
```

Works like a charm ;)

drew-r:
thanks both.
@mceachen that makes perfect sense. 
While I have your ear so to speak, do you see an elegant way to expose the static properties defined on Model classes when using `bindKnex()`?

elhigu:
@mceachen all good, let us know about Definedlytyped stuff how it goes 👍

elhigu:
Aand thanks! I look forward to test these in our own project which we've been writing with typescript :)

drew-r:
Also there are a few edits that need to be made... `query()` needs an
optional transaction parameter, `transaction` itself needs to be exported
additionally as a namespace with the static `Transaction.start()` relocated
and exported within.
fromJson options are optional also I believe.

Sorry I would put a patch together but restricted to mobile right now

mceachen:
Ah OK, I'll make a new pr with those changes, and link to the definitely
typed pr, too.

mceachen:
@drew-r just created #246 to address your comments.

Amerzel:
I would like to see these submitted to Definitely Typed to make them easier to use. Is that still on the radar @mceachen?

Thanks for doing the work and adding these.

mceachen:
I had one more set of changes before I wanted to submit to DT, which are here: #254. I'll reference the DT PR here when I post it.

mceachen:
@Amerzel In reading the best-practices from DefinitelyTyped, they recommend the library self-host typings. See #255 which moves the types from the example and into the objection module proper.

Amerzel:
@mceachen,

That makes sense. There's pluses and minuses to doing it that way. It makes it really easy to use the types but adds a bit more work for the library maintainers. They would be expected to maintain and update the typings as new features are added to the library. If the maintainers don't use TypeScript having the typings self hosted adds some additional work that they may not care about or want to maintain.

James

elhigu:
@Amerzel yep, we've discussed about it a bit in gitter chat. The same amount of maintenance work is in both options: Someone has to update typings when new APIs are released, if there is no one who is committed to keep them updated for releases, then we can just drop the support. 

Good thing is that if we keep typings in the same repo, it is easy to keep versions synched that correct version of typings comes directly with the certain library version (I suppose that definitely typed has also mechanism for it, but to me keeping everything in the same repo feels easier).

Amerzel:
Sounds great. I'm glad it was discussed. Keeping everything in the same repo is definitely easier :)

Issue #243, opened 2016-11-10
jsonSchema.oneOf is not supported in ModelBase mergeWithDefaults


kapouer:
https://github.com/Vincit/objection.js/blob/master/src/model/ModelBase.js#L503
fails with a schema like
```
{
  type: "object",
  oneOf: [
    {properties: { what: { "enum": ["thatclass"] }, title: { type: "string"} }},
    {properties: { what: { "enum": ["thisclass"] }, body: { type: "string"} }}
  ]
}
```
because default values won't be found, while there are some.
Related a little to https://github.com/Vincit/objection.js/issues/205.

kapouer:
It seems to me that https://github.com/epoberezkin/ajv#assigning-defaults is precisely doing what mergeWithDefaults does ?

kapouer:
It's interesting to see it working, yet ajv has some limits too regarding defaults inside oneOf/anyOf/allOf, as explained at the bottom of the documention given above.
A workaround is to use json schema v5 switch mecanism.

koskimas:
We should probably remove `mergeWithDefaults` and just use ajv default system. There is only one thing that needs to be solved: ajv assings the default values to the input object where as `mergeWithDefaults` creates a clone making sure the input object is never modified. With ajv we would need to take the clone each time since we don't know if some default values will be set. That probably won't matter much, since inserts are usually far less common than reads, and they take most of their time in the database anyway.

kapouer:
Closing in favor of #244.

Issue #241, opened 2016-11-08
Support jsonschema v5 proposals


musou1500:
Hi.
What do you think about enabling `v5` option of `ajv`?
Thanks.

koskimas:
I actually thought it was already enabled 😬 I'll enabled it in the next release.

musou1500:
thanks!!
I'm looking forward to next release!

koskimas:
Fixed in 0.7.0-rc.1.

PR #244, opened 2016-11-11
Unleash ajv to set defaults and manage compiled schema cache

kapouer:
I took the opportunity to enable v5 (see also #241) as well.
Fixes: #243.

I suspect this might not be mergeable "as is" but it's a starting point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "src/*"
   ],
   "dependencies": {
-    "ajv": "^4.1.5",
+    "ajv": "^4.8.2",
     "babel-runtime": "^6.6.1",
     "bluebird": "^3.4.6",
     "lodash": "^4.7.0"
--- a/src/model/ModelBase.js
+++ b/src/model/ModelBase.js
@@ -66,6 +66,7 @@ export default class ModelBase {
     }
 
     const validator = this.constructor.getJsonSchemaValidator(jsonSchema, options.patch);
+    json = cloneObject(json);
     validator(json);
 
     if (validator.errors) {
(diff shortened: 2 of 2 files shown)


koskimas:
The own caching was there for a reason. For the 99% case where the schema never changed the compiled function was cached directly. The changing schema cache was also alot faster than AJV's because I used JSON.stringify. I don't think we should remove that. Also the tests fail.

kapouer:
The test failure
> Error: Could not connect to postgres. Make sure the server is running and the database objection_test is created.
doesn't seem to be related to what i changed.

Oh - then ajv's choice of not using JSON.stringify seems odd.
Isn't JSON.stringify not deterministic today ? I'll ask there.

koskimas:
I think `JSON.stringify` is not used so that `{a: 1, b: 2}` and `{b: 2, a: 1}` hit the same cache. In our case that doesn't matter. The most important tweak is the 99% case I mentioned that doesn't serialize the schema at all.

kapouer:
Right, then extra care must be added when updating ajv, since it caches compiled schemas by default - voiding the optimization you made to avoid extra stringification.

kapouer:
Another small thing to mention: https://github.com/nickyout/fast-stable-stringify seems a bit better maintained and faster. I also ran a quick jsperf and it seems it's 60% slower than native JSON.stringify.
https://jsperf.com/fastjson (but the js file is hosted on non-ssl url, making chrome complain).

kapouer:
Rebased, now all tests pass.

koskimas:
Thanks!

PR #246, opened 2016-11-14
Updates to typing, with examples

mceachen:
These changes address @drew-r 's comments in PR #239.

### DIFF
--- a/examples/express-ts/src/examples.ts
+++ b/examples/express-ts/src/examples.ts
@@ -97,6 +97,14 @@ objection.transaction(Person, (P: typeof Person) => {
   return Promise.resolve('yay');
 });
 
+objection.transaction.start(Person).then((trx: objection.Transaction) => {
+  Person.bindTransaction(trx).query()
+    .then(() => trx.commit())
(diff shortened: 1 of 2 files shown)


elhigu:
Seems legit 👍

mceachen:
Thanks, @elhigu !

drew-r:
@mceachen check out https://github.com/drew-r/objection.d.ts for a crude generically typed query builder. 
The promise result type will default to the type of `this` context e.g 
```
let person = await Person.query().first(); // type inference says person is a Person 
let people = await Person.query(); // type inference says people is a Person[] 
```
but can be overridden 
```
let personAggregation = await Person.query<any>().first(); // any
let personAggregations  = await Person.query<any>(); // any[]
```

sorry I don't have a patch against your implementation but do you think this approach addresses your comments in #239 ? 
Seems to me that Objection is way better placed than bookshelf or sequelize for TypeScript & ES7 adoption generally, a set of type definitions that provide a great development experience will help cement that, so I hope soon we can get either a PR over to DefinitelyTyped or get the type defs bundled with the library itself.

PR #255, opened 2016-12-01
Types!

mceachen:
The original plan was to move typings to DefinitelyTyped, but the [current advice is to self-publish](http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html): 

> "There are two main ways you can publish your declaration files to npm:
> 1. bundling with your npm package, or
> 2. publishing to the @types organization on npm.
> If you control the npm package you are publishing declarations for, then the first approach is favored. That way, your declarations and JavaScript always travel together."

This PR moves the typings out of the example project and into the Objection module proper.

To make that all work, the following changes were made:

1. TypeScript is added as a development dependency so `example.ts` can be compiled against the definition file
2. `examples/express-ts` has `../..` as a dependency to support local development and verification
3. The typings from `examples/express-ts` were updated (JsonSchema, for example, is part of the typings), and moved into `./typings/objection`.
4. `package.json` has a new "types" reference to the Objection type definition file, and the "typings" directory is included in the `files` directive.
5. The `test` and `test-travis` scripts now call `tsc` after `npm run build` and before `mocha`. On my i3 linux box, this adds less than a second to the test suite (18.5s without `tsc`, 19s with `tsc`).

While I was verifying the example code, I fixed a couple defects in the prior typings, and added tests to `examples.ts`. 

1. There was a bug in the typings for transactions, which expanded the `ModelClass` in the callback parameters. By forcing the `ModelClass` to always reference a `T extends Model`, I was able to change the transaction function signatures to a specific `ModelClass`, like `Person`. This wasn't possible before because the signature around `Model.extend` was too simplistic. 

2. `Model.extend` didn't do what it should. Now the return value is union type of both the provided subclass as well as Model.

3. Some return types for `QueryBuilder`, including `Model.$query()` and `.findById()`, exposed the need for optionality. Examples were added to `examples.ts`.

4. `update` and `patch` methods return the number of rows updated. This wasn't reflected in the typings.


elhigu:
Looks good to me... I don't see any harm of running tsc when testing, since one can still do any kind javascript changes and test will always pass unless typings file is touched.

examples/express-ts/package.json
```diff
@@ -8,26 +8,30 @@
...
-    "knex": "^0.10.0",
-    "lodash": "^4.12.0",
+    "body-parser": "^1.15.2",
+    "express": "^4.14.0",
+    "knex": ">= 0.8.0 <= 0.12.5",
+    "lodash": "^4.17.2",
     "morgan": "^1.7.0",
-    "objection": "^0.6.1",
-    "sqlite3": "^3.1.3"
+    "objection": "../..",
```
koskimas: The local version (head) may not be compatible with the example project. Unless you are willing to keep it up to date with the most recent updates all the time, this should be reverted

koskimas:
Tests are failing with older node versions.

examples/express-ts/src/api.ts
```diff
@@ -64,15 +64,15 @@ export default function (app: express.Application) {
       .catch(next);
```
koskimas: That was because you wrote `next(throwNotFound)` instead of `next(throwNotFound())`. Actually even the latter wouldn't work and it should have been something like `next(createNotFoundError())`. `res.sendStatus(404)` is fine though.

examples/express-ts/src/api.ts
```diff
@@ -150,15 +151,15 @@ export default function (app: express.Application) {
...
-    await movie
-      .$relatedQuery('actors')
-      .relate(req.body.id)
-      .catch(next);
+      res.sendStatus(404);
+    } else {
+      await movie
+        .$relatedQuery('actors')
+        .relate(req.body.id)
+        .catch(next);
```
mceachen: OK, I'll address that now.

koskimas:
Fantastic job @mceachen by the way! Thank you so much for doing this. I'll be more than happy to merge this as soon as the travis tests are fixed.

mceachen:
I've replaced this PR with #267.

build.js
```diff
@@ -55,4 +55,17 @@ for (var i = 0; i < src.length; ++i) {
...
+  var process = require('process');
+
+  // TypeScript requires node > 4:
+  if (process.version > 'v4') {
+    // sanity-check the TypeScript definitions:
+    child_process.execSync(path.join(__dirname, "node_modules", ".bin", "tsc"), { cwd: __dirname });
+  }
+} catch (e) {
+  // node is too old, nevermind.
 }
```
koskimas: Doesn't the execSync throw an exception of tsc fails and it gets caught here? It seems to me this never fails.

mceachen:
Good catch, I'll make the catch conditional.

On Dec 23, 2016 12:26 PM, "Sami Koskimäki" <notifications@github.com> wrote:

*@koskimas* commented on this pull request.
------------------------------

In build.js
<https://github.com/Vincit/objection.js/pull/255#pullrequestreview-14336895>
:

> @@ -55,4 +55,17 @@ for (var i = 0; i < src.length; ++i) {

   fsExtra.ensureDirSync(path.dirname(dst));
   fs.writeFileSync(dst, code);
+}
+
+try {
+  var child_process = require('child_process');
+  var process = require('process');
+
+  // TypeScript requires node > 4:
+  if (process.version > 'v4') {
+    // sanity-check the TypeScript definitions:
+    child_process.execSync(path.join(__dirname, "node_modules",
".bin", "tsc"), { cwd: __dirname });
+  }
+} catch (e) {
+  // node is too old, nevermind.
 }

Doesn't the execSync throw an exception of tsc fails and it gets caught
here? It seems to me this never fails.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Vincit/objection.js/pull/255#pullrequestreview-14336895>,
or mute the thread

build.js
mceachen: Good catch, I'll make the catch conditional.

mceachen:
@koskimas if there's anything else I can update to make this PR more acceptable, holler.

koskimas:
@mceachen There appears to be something truly weird going on with this PR. The diff shows changes (my commits) that already exist in the master branch. Am I not getting something obvious here? Why is the diff like that?

koskimas:
Damn.. I just realized that I must have force pushed some stuff to the master and you have different copies of the commits 😬 I should really stop force pushing stuff to master...

mceachen:
Do you want me to make another PR from my diff?

koskimas:
Yes please! Either that or drop my commits from the PR branch.

mceachen:
OK, just made #272.

PR #269, opened 2017-01-02
Add script for knex migrations in examples

p4bloch:
Knex is already declared as a dependency, so we could be using the knex CLI tool directly and avoid asking users to install a global dependency. `db:migrate` might not be the best name, so if you think it should be called something else please let me know.

The other change I propose is that since `node_modules/.bin` is in the path for npm scripts and `babel-cli` makes `babel-node` available through that folder, we use that instead of hardcoding the full path to the file (which may change in a future version of babel).

If you think this should be proposed in different PRs, please let know and I will happily split them out :)

### DIFF
--- a/examples/express-es6/README.md
+++ b/examples/express-es6/README.md
@@ -4,9 +4,8 @@
 git clone git@github.com:Vincit/objection.js.git objection
 cd objection/examples/express-es6
 npm install
-# We use knex for migrations in this example.
-npm install knex -g
-knex migrate:latest
+# Run knex migrations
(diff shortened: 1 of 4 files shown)


koskimas:
While this is a great way to do this I don't think we should add it to the example projects. The examples are there to explain how things work, and this adds one more indirection to the process. A beginner needs to understand what `npm run` does and then see package.json to understand that knex cli is used.

I know that many other parts of the examples are too complex and contradict what I just said, but simplifying them is on my TODO list :D

PR #270, opened 2017-01-03
Initial ref syntax

elhigu:
Adds support to easy places to be able to use ref syntax instead of knex.raw... check test cases which parts are now supported.

Stuff that should be done nicer and implemented later on: 

* Documentation!!
* JoinBuilder is currently inherited from BaseQueryBuilder so there are way too many methods available
* Typings at least for ReferenceBuilder
* Add also shorthand wrappers for lit (literals with casting info and e.g. which sql-array/list syntax certain passed array value is) and raw (convenience wrapper for knex.raw, which will be converted to knex.raw by objection APIs like ref)
* Implementations for mysql and sqlite json syntax

### DIFF
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import ModelBase from './ModelBase';
 import QueryBuilder from '../queryBuilder/QueryBuilder';
+import ReferenceBuilder from '../queryBuilder/ReferenceBuilder';
 import inheritModel from './inheritModel';
 import RelationExpression from '../queryBuilder/RelationExpression';
 import {inheritHiddenData} from '../utils/hiddenData';
@@ -23,6 +24,7 @@ import JoinEagerOperation from '../queryBuilder/operations/JoinEagerOperation';
 import WhereInEagerOperation from '../queryBuilder/operations/WhereInEagerOperation';
 
 const KnexRaw = require('knex/lib/raw');
+const KnexQueryBuilder = require('knex/lib/query/builder');
 
 const JoinEagerAlgorithm = () => {
   return new JoinEagerOperation('eager');
@@ -267,7 +269,8 @@ export default class Model extends ModelBase {
         const attr = jsonAttr[i];
         const value = json[attr];
 
-        if (_.isObject(value) && !(value instanceof KnexRaw)) {
+        // list of omitted objects is copy pasted from splitQueryProps
+        if (_.isObject(value) && !(value instanceof KnexQueryBuilder || value instanceof KnexRaw || value instanceof ReferenceBuilder)) {
           json[attr] = JSON.stringify(value);
         }
       }
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -22,6 +22,8 @@ import InstanceDeleteOperation from '../queryBuilder/operations/InstanceDeleteOp
 import JoinEagerOperation from '../queryBuilder/operations/JoinEagerOperation';
 import WhereInEagerOperation from '../queryBuilder/operations/WhereInEagerOperation';
 
+const KnexRaw = require('knex/lib/raw');
+
 const JoinEagerAlgorithm = () => {
   return new JoinEagerOperation('eager');
 };
@@ -245,7 +247,11 @@ export default class Model extends ModelBase {
         const value = json[attr];
 
         if (_.isString(value)) {
-          json[attr] = JSON.parse(value);
+          try {
+            json[attr] = JSON.parse(value);
+          } catch (err) {
+            // json column might contain plain single string which is not wrapped to array / object
+          }
         }
       }
     }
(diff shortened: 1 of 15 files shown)


elhigu:
This stuff requires psql >= 9.5 gotta set travis to use it also... tests might have some problems with node < 4 . I'll check those a bit later...

kapouer:
That looks fantastic but is missing documentation ?

elhigu:
@kapouer yes it is, I wanted sami to check this out before writing it to have kind of final code ready first.

tests/integration/jsonQueries.js
```diff
@@ -43,7 +46,267 @@ module.exports = function (session) {
...
+            expect(_.first(result)).to.eql({ id: 2, foo: 2 });
+          });
+      });
+
+      it('should be able to use ref with groupBy and nested having', function () {
+        return BoundModel.query()
+          .select(['id', ref('jsonObject:attr').as('foo')])
+          .groupBy([ref('jsonObject:attr'), 'id'])
+          .having('id', '>=', ref('jsonObject:attr').castInt())
+          // knex doesnt seem to support nested having
```
elhigu: outdated comment, it just didn't support raw as a first parameter for having.

tests/integration/jsonQueries.js
```diff
@@ -43,7 +46,267 @@ module.exports = function (session) {
...
+      it('should be able to use knex.raw to jsonb column in update', function () {
+        return BoundModel.query()
+          .update({
+            jsonArray: BoundModel.knex().raw('to_jsonb(??)', ['name'])
+          }).then(function (result) {
+            expect(result).to.be(4);
+          });
+      });
+
+      it('should update nicely', function () {
```
elhigu: need to update more descriptive test case name

tests/integration/jsonQueries.js
```diff
@@ -43,7 +46,267 @@ module.exports = function (session) {
...
+            expect(result).to.eql([{
+              id: 1,
+              name: '1',
+              jsonObject: { attr: 'test1' },
+              jsonArray: 'test1' }
+            ]);
+          });
+      });
+
+      it('should patch nicely', function () {
```
elhigu: bad name here as well

src/model/Model.js
```diff
@@ -245,7 +247,11 @@ export default class Model extends ModelBase {
         const value = json[attr];
 
         if (_.isString(value)) {
-          json[attr] = JSON.parse(value);
+          try {
+            json[attr] = JSON.parse(value);
+          } catch (err) {
+            // json column might contain plain single string which is not wrapped to array / object
+          }
```
koskimas: Do we really need this try-catch?

src/model/Model.js
```diff
@@ -261,7 +267,7 @@ export default class Model extends ModelBase {
         const attr = jsonAttr[i];
         const value = json[attr];
 
-        if (_.isObject(value)) {
+        if (_.isObject(value) && !(value instanceof KnexRaw)) {
           json[attr] = JSON.stringify(value);
```
koskimas: Is this needed after the `UpdateOperation` stuff is moved to `onBuild`?

src/queryBuilder/operations/UpdateOperation.js
```diff
@@ -13,6 +17,36 @@ export default class UpdateOperation extends QueryBuilderOperation {
...
+        );
+        // (looks like I just can't set new stuff to args[0] that
+        //  would be the converted object so I just modify the
+        //  original args|0] object)
+        delete args[0][key];
+      } else {
+        args[0][key] = loweredValue;
+      }
+    });
+
```
koskimas: This should be moved to `onBuild` and the references should be stashed in `splitQueryProps` function as discussed.

src/queryBuilder/operations/WrappingQueryBuilderOperation.js
```diff
@@ -38,6 +42,10 @@ function wrapArgs(op, builder, args) {
       } else if (includesUndefined(arg)) {
         throw new Error(`undefined passed as an item in argument #${l} for '${op.name}' operation. Call skipUndefined() method to ignore the undefined values.`);
       }
+      // convert reference builders to knex.raw
+      args[i] = args[i].map(arg => {
+        return arg instanceof ReferenceBuilder ? knex.raw(...arg.toRawArgs()) : arg;
+      })
```
koskimas: Is this needed since you normalize the `select` args into a flat array?

src/model/Model.js
elhigu: Looks like it was necessary, because pg driver converts json-string to non quoted string.

src/model/Model.js
elhigu: Yes. I added test case, where json column value is updated with knex.raw and it fails even without reference builder stuff. 

This failed because circular data structure:
```js
        return BoundModel.query()
          .update({
            jsonArray: BoundModel.knex().raw('to_jsonb(??)', ['name'])
          }).then(function (result) {
            expect(result).to.be(4);
          });
```

src/queryBuilder/operations/WrappingQueryBuilderOperation.js
elhigu: `.groupBy()` with multiple arguments will fail without this.

elhigu:
Fixed everything!

src/model/Model.js
koskimas: shitballs

PR #277, opened 2017-01-08
Documentation for ref() and updating / patching attribute inside json column

elhigu:
Any improvements are welcome.

### DIFF
--- a/doc/includes/API.md
+++ b/doc/includes/API.md
@@ -84,6 +84,87 @@ Type|Description
 
 
 
+### Query building global helpers
+
+#### import { ref } from 'objection';
+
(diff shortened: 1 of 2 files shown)


koskimas:
The code blocks won't be placed where you write them but right after the previous header. So for example saying this

`ref()` can be passed to all `QueryBuilder` APIs which supports `Model.raw()` arguments like:

and writing a code block after it will look confusing since the code won't come after that but actually before it. The text needs to be reformatted so that it works with the out-of-place code blocks.

elhigu:
Maybe now? I didn't see any other places in addition to the `ref() can be passed to all QueryBuilder APIs which supports Model.raw() arguments like:`

PR #287, opened 2017-01-24
(typings) allow Raw in columnName

kbkk:
The docs mention you can pass just a Raw column name.

```
.where(Person.raw('lower("firstName")'), 'like', '%ennif%')
```


koskimas:
Thank you!

PR #288, opened 2017-01-24
(typings) Models can have composite keys

kbkk:
idColumn can be an array of strings  

Example from docs:
```ts
Person.idColumn = ['firstName', 'lastName', 'dateOfBirth'];
```


koskimas:
Thanks!

PR #291, opened 2017-02-02
Update types with generics for insert

rodrigopivi:
and define the static methods `JoinEagerAlgorithm` and `WhereInEagerAlgorithm`.

It would also be nice that the type defs are released at the new version

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -210,6 +210,9 @@ declare module "objection" {
     static HasManyRelation: Relation;
     static ManyToManyRelation: Relation;
 
+    static JoinEagerAlgorithm: () => any;
+    static WhereInEagerAlgorithm: () => any;
+
     // "{ new(): T }" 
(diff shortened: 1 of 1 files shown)


koskimas:
Thank you @rodrigopivi!

typings/objection/index.d.ts
```diff
@@ -298,9 +301,9 @@ declare module "objection" {
    */
   export interface QueryBuilder<T> extends QueryBuilderBase<T>, Promise<T[]> { }
 
-  interface Insert {
-    <M extends ModelOrObject>(modelOrObject?: M): QueryBuilderSingle<M>;
-    <M extends ModelsOrObjects>(modelsOrObjects?: M): QueryBuilder<M>;
+  interface Insert<T> {
+    <M extends ModelOrObject>(modelOrObject?: M): QueryBuilderSingle<T>;
+    <M extends ModelsOrObjects>(modelsOrObjects?: M): QueryBuilder<T>;
     (): this;
```
mceachen: Why was `<T>` introduced here? Did you want to remove `M`?

PR #310, opened 2017-02-25
Upgrade to TS2.2 and fix Insert typing

mceachen:
`Insert<T>` used `ModelsOrObjects`, which includes `ModelOrObject`. 

The new, (better) typescript parser found this issue that caused `insert([])` to now return singular values. I changed the plural result matcher to be `Object[] | Model[]` before matching on Object (which matches arrays).

I updated all other packages as well in the example project (and wrote a simple repl.js to ensure this was all behaving well in the node REPL).

### DIFF
--- a/examples/express-ts/package.json
+++ b/examples/express-ts/package.json
@@ -14,23 +14,23 @@
   ],
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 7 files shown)


koskimas:
Thank you!

PR #321, opened 2017-03-09
optimization: added limit to QueryBuilder, fixes #307

### LINKED ISSUE #307: Possible optimization with first() method
Looking through the source code I noticed that the `first` query method pulls down the entire dataset and then returns the first element in the array. A possible optimization to this would be to automatically append `limit(1)` if not present in order to limit the returned dataset to just one record. Thoughts?

fl0w:


fl0w:
I think I've borked my environment because I don't get these locally, brb.

koskimas:
This seems to break eager queries with `JoinEagerAlgorithm` because of the way rows are returned for joins. Even if there is just one root level item, the result row set of a join query may contain multiple rows (the related models). Currently if `first` is chained the result is what you expect, but after this change `first` can no longer be used with `JoinEagerAlgorithm`. That may be ok, but it makes this a breaking change.

fl0w:
Oh I get it, nice explanation.

I'm still a bit confused though because I don't get these errors locally. I run pg9.6 and don't have mysql installed. Gonna replicate travis env locally.

koskimas:
Weird. Maybe the transpilatipn is borked somehow? Have you tried the good old rm -rf node_modules?

fl0w:
I tried reinstalling node_modules again, without profit. I did however upgrade to 0.7 from rc3 which got TS to stop complaining. By not having mysql installed, the error still doesn't crop up.

```
 1218 passing (47s)
  4 pending
  1 failing

  1) integration tests mysql "before all" hook:
```

When I installed mysql 5.7.x (through brew) but didn't start it, tests started to behaving as expected:

```
 1215 passing (38s)
  4 pending
  4 failing

# same as travis now
```

Not that it matters to this PR but could there be some header/binary dependency of MySQL which borks the tests if not available? Though, clearly, I should have had tested for MySQL as well - my bad (don't use it so never pay attention to it).

Anyway, I can reproduce the error now at least. I'll take a read how `JoinEagerAlgorithm` actually works.

fl0w:
Closing this based on your comment for now; At least got my local objection.js clone to work!

PR #328, opened 2017-03-13
Update ValidationError typedef to be class

pvatterott:
ValidationError is currently defined in index.d.ts as an interface.
However, typescript interfaces are not runtime constructs, so it is
impossible to make calls like: `if (err instanceof ValidationError)`.

In ValidationError.js, ValidationError is defined as a class that
extends Error, so update the type definition accordingly.


koskimas:
Loogs good to me. Thanks!

PR #329, opened 2017-03-13
Change ModelOrObject to Partial<T>

pvatterott:
As far as I can tell, the functions that accept ModelOrObject parameters
(Insert<T>, InsertGraphAndFetch<T>, insertAndFetch<T>. update*, patch,
relate) all allow Object as well as Model so that the caller can provide
an object that does not define all the properties of the Model.

This seems like a weaker type check than is possible with the Partial<T>
type that is provided by typescript:
https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types

This commit removes the ModelOrObject and ModelsOrObjects types and replaces 
them with Partial<T> and Partial<T>[] where appropriate. This will provide stronger 
type checking in the affected calls: callers will no longer be able to set arbitrary
properties in their insert/update/patch calls. They will be restricted to the subset of
properties that exist on the model definition.


koskimas:
Sounds good. @mceachen Do you have any thoughts about this?

mceachen:
This sounds great, I'll review the PR later today.

typings/objection/index.d.ts
```diff
@@ -302,15 +299,14 @@ declare module "objection" {
   export interface QueryBuilder<T> extends QueryBuilderBase<T>, Promise<T[]> { }
 
   interface Insert<T> {
-    <M extends Model[] | Object[]>(modelsOrObjects?: M): QueryBuilder<T>;
-    <M extends ModelOrObject>(modelOrObject?: M): QueryBuilderSingle<T>;
+    (modelsOrObjects?: Array<Partial<T>>): QueryBuilder<T>;
```
mceachen: I think it's generally preferred to use `Partial<T>[]` instead of `Array<Partial<T>>`.

pvatterott: Sounds good! Thanks for taking a look

mceachen: Also, could you add an example to `tests/ts/examples.ts`? Something as simple as
```
// Verify that insert accepts Partial<Person>:

Person.query()
  .insert({ firstName: "test" })
  .then((p: Person) => console.log("inserted", p))
```

would be great. Thanks.

pvatterott:
Let me know what to do about the build failure here. The build fails on two older node versions when compiling a dependency (fs-extra). But, the dependency doesn't support that older node version:
`
npm WARN engine fs-extra@2.1.0: wanted: {"node":">=4.5.0"} (current: {"node":"0.12.18","npm":"2.15.11"})
`
Looks like fs-extra 2.1 was released 2 hours ago: https://github.com/jprichardson/node-fs-extra/releases

Easy solution would be to pin to 2.0.0 but I'll let your team decide on the best solution.

mceachen:
> Looks like fs-extra 2.1 was released 2 hours ago

And https://github.com/jprichardson/node-fs-extra/issues/390 is tracking the fix. I think it's fine to pin to the prior version so they can merge this.

pvatterott:
Should be good to merge. Added the type assertion and pinned to 2.0.0

koskimas:
Thanks!

PR #344, opened 2017-03-31
Support passing in attributes on model constructor

ackerdev:
This patch changes the Model constructor to support the same options `fromJson` does. I'd like this functionality as there are some handy utilities out there that you can use with ORMs, but they sometimes expect the constructor to accept an object representing desired attributes as it's first parameter.

I wasn't sure how you'd want to handle updating the unit tests for this; didn't think it would have been appropriate to duplicate the entire block of tests for fromJson.

### DIFF
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -126,6 +126,10 @@ export default class Model {
    */
   static defaultEagerOptions = null;
 
+  constructor(json, opts) {
+    this.$setJson(json || {}, opts);
+  }
+
   /**
    * @param {*=} id
    * @returns {*}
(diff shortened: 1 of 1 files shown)


koskimas:
This seems like a good addition if you can make it work without breaking anything. Now the tests fail.

You also need to write tests and documentation for this. Copy-pasting a whole block of tests is probably not the best idea. Test either of them (the constructor or fromJson) and then test that one calls the other.

ackerdev:
@koskimas Yeah I'm not sure what's up with that failure yet but I'll try to take a look. I noticed there weren't any spying libraries installed already; do you have any preferences?

koskimas:
I'm guessing this can be closed now?

PR #347, opened 2017-04-01
Added information about PostgreSQL JSONB indexing to recipes

elhigu:
Pretty long entry, probably still has bunch of stupid errors, need more eyes to look it through.


elhigu:
Ah right. I didnt knew that recipes were also in slate pages. I'll fix markdown to be slate compatible.

elhigu:
Maybe now its also slate proof.

PR #376, opened 2017-05-03
Adding more Postgres "returning" tricks for "delete".

newhouse:
- Added more Postgres "returning" tricks examples for `delete()`.
- Find/replaced incorrect relative link/anchor from my prior doc updates.
- Fixed an unmatched paren from some earlier doc entry.


koskimas:
Thank you!

PR #377, opened 2017-05-03
Adding tests for PostgreSQL delete returning

newhouse:
I wrote a few tests in what I think is the right spot, but I couldn't verify that they were passing on my local. Received this error:
```
2) integration tests postgres "before all" hook:
     Error: Could not connect to postgres. Make sure the server is running and the database objection_test is created. You can see the test database configurations from file /Users/newhouse/projects/objection.js/testUtils/index.js
```

If that's something I can ignore, then they're probably good?

### DIFF
--- a/doc/includes/RECIPES.md
+++ b/doc/includes/RECIPES.md
@@ -262,7 +262,7 @@ the [`joinRelation`](#joinrelation) method family.
 
 ## PostgreSQL "returning" tricks
 
-> Insert and return the data in 1 query:
+> Insert and return a Model instance in 1 query:
 
 ```js
 Person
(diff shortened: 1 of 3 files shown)


newhouse:
@koskimas Let me know if you want me to change anything, or feel free to take over this branch and fix it up!

koskimas:
@newhouse Some tests are failing. [this](http://vincit.github.io/objection.js/#development-setup) may help you to setup the test environment.

newhouse:
Thanks @koskimas looks like they're all passing now. Happy to contribute. When do the doc updates get into the github.io site?

koskimas:
Awesome! Thank you!

PR #381, opened 2017-05-04
Fix typings

drew-r:
It doesn't follow that deleteById, patch or update always produce a number result (postgres `.returning('*')`) - so use the QueryBuilder result type param. For queries that do result in a number, use `.query<number>()`.

Similarly, it doesn't follow that patch or update always return a single result so full QueryBuilder here.

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -336,23 +336,14 @@ declare module "objection" {
     insertWithRelated: Insert<T>;
     insertWithRelatedAndFetch: InsertGraphAndFetch<T>
 
-    /**
-     * @return a Promise of the number of updated rows
-     */
-    update(modelOrObject: Partial<T>): QueryBuilderSingle<T>;
+    update(modelOrObject: Partial<T>): QueryBuilder<T>;
     updateAndFetch(modelOrObject: Partial<T>): QueryBuilderSingle<T>;
     updateAndFetchById(id: Id, modelOrObject: Partial<T>): QueryBuilderSingle<T>;
 
-    /**
-     * @return a Promise of the number of patched rows
-     */
-    patch(modelOrObject: Partial<T>): QueryBuilderSingle<T>;
+    patch(modelOrObject: Partial<T>): QueryBuilder<T>;
     patchAndFetchById(id: Id, modelOrObject: Partial<T>): QueryBuilderSingle<T>;
     patchAndFetch(modelOrObject: Partial<T>): QueryBuilderSingle<T>;
 
-    /**
-     * @return a Promise of the number of deleted rows
-     */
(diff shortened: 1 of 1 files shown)


koskimas:
The default output is number and not an object. Doesn't this make the default an object? Also saying `.query<number>` makes the input `Partial<number>` that makes no sense. Or am I missing something here?

koskimas:
@mceachen Could you check this out. I'm still 100% noob when it comes to typescript.

typings/objection/index.d.ts
```diff
@@ -336,24 +336,15 @@ declare module "objection" {
...
-    update(modelOrObject: Partial<T>): QueryBuilderSingle<number>;
+    update(modelOrObject: Partial<T>): QueryBuilder<T>;
     updateAndFetch(modelOrObject: Partial<T>): QueryBuilderSingle<T>;
     updateAndFetchById(id: Id, modelOrObject: Partial<T>): QueryBuilderSingle<T>;
 
-    /**
-     * @return a Promise of the number of patched rows
-     */
-    patch(modelOrObject: Partial<T>): QueryBuilderSingle<number>;
+    patch(modelOrObject: Partial<T>): QueryBuilder<T>;
```
mceachen: Thanks for the catch, I agree it shouldn't be `QueryBuilderSingle<number>`. It seems like this should be `QueryBuilderSingle<T>`. If I'm remembering correctly, I made it a number because the docs say

> NOTE: The return value of the query will be the number of affected rows. If you want to patch a single row and retrieve the patched row as a result, 

and I didn't think about subsequent useful calls to the queryBuilder like `.where`.

typings/objection/index.d.ts
```diff
@@ -336,24 +336,15 @@ declare module "objection" {
...
     patchAndFetchById(id: Id, modelOrObject: Partial<T>): QueryBuilderSingle<T>;
     patchAndFetch(modelOrObject: Partial<T>): QueryBuilderSingle<T>;
 
-    /**
-     * @return a Promise of the number of deleted rows
-     */
-    deleteById(idOrIds: IdOrIds): QueryBuilderSingle<number>;
+    deleteById(idOrIds: IdOrIds): QueryBuilderSingle<T>;
```
mceachen: 👍

chamini2:
Hello, I have a question:

```ts
const deleted: number = await Model.query().delete().where({some, columns});
```

This is not working, since it thinks that it returns a `Model[]`, shouldn' this be fixed in this PR too?

chamini2:
Maybe

```
delete(): number
delete(returning: string | string[]): this
```

mceachen:
These have been incorporated into the typings (see especially 2a077f10f8a). @drew-r can you open another issue if things aren't still quite right?

Thanks!

PR #408, opened 2017-06-12
correct typo on example `$beforeInsert` hook

brianphillips:
hook is named `$beforeInsert` not `beforeInsert`


koskimas:
Wow, that was fast 😄. Thank you!

PR #449, opened 2017-07-21
Linted typings and added model-less transaction signature

mceachen:
* s/String/string
* s/Object/object
* replaced callable genericized types with interfaces. See https://github.com/Microsoft/TypeScript/issues/17097#issuecomment-316857733
* coallesced some method calls where union types made sense
* replace Function with `() => void`
* tslint fixed a bunch of missing semicolons


mceachen:
Thanks!

examples/express-ts/package.json
```diff
@@ -14,24 +14,24 @@
...
-    "knex": "^0.12.7",
+    "body-parser": "^1.17.2",
+    "express": "^4.15.3",
+    "express-promise-router": "^2.0.0",
+    "knex": "^0.13.0",
     "lodash": "^4.17.4",
-    "morgan": "^1.8.1",
-    "objection": "^0.7.0",
+    "morgan": "^1.8.2",
+    "objection": "../..",
```
fl0w: That's a neat trick.

koskimas: Actually I didn't notice that. I think that's not the best idea since the example may break if the master branch has breaking changes. I usually only update the examples with actual releases so that I don't have to constantly be changing them. Also if a breaking change is updated to the example project and then someone starts to use objection based on that example, then `npm install objection` will work different from the example.

mceachen: IMHO, I'd actually add the examples and tests to ci, which would predicate this change. 

I think the people that just clone this folder to start their project should use a release tag, but it might be better to tell people to use a [generator](https://expressjs.com/en/starter/generator.html).
Do you want me to back out this change, though?

koskimas: > IMHO, I'd actually add the examples and tests to ci, which would predicate this change.
Sure, but we would need to write tests for the example projects.

> Do you want me to back out this change, though?
No it's fine.

mceachen: > Sure, but we would need to write tests for the example projects

I think that might be a good effort--I'm always nervous/suspicious of untested code (and it would provide some examples to people on how to test code that uses Objection).

Do you want some PRs that add tests? I'd like to see how you guys handle testing, in any event...

fl0w: I've got to stop ninja-reading code when it's not my place. Always raising some kind of ruckus.

mceachen: @flow Actually I totally disagree--especially if there's a functional discussion that happens from the comment! In almost any case, scrutiny should be encouraged. 

I thought it was a neat trick, too (and first saw it in one of @sindresorhus' repos, I think?)

PR #457, opened 2017-07-27
typings: Remove generic type from Transaction

### LINKED ISSUE #455: Generic Transaction
With the new update, the Transaction became a generic type, but its broken all my code, see the example below:

image

I can't use the same transaction for two different queries and it create a idea that is not possible.

danieloprado:
Fix #455

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,10 @@
   "license": "MIT",
   "scripts": {
     "test": "mocha --slow 10 --timeout 15000 --reporter spec --recursive tests",
-    "test-travis": "istanbul --config=.istanbul.yml cover _mocha -- --slow 100 --timeout 60000 --reporter spec --recursive tests",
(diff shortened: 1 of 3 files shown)


elhigu:
Should there be a regression test case added for #455 to https://github.com/Vincit/objection.js/blob/master/tests/ts/examples.ts

koskimas:
The examples.ts file is not even ran with the tests at the moment. PR to fix that too would be welcome

danieloprado:
Ok I will add this file to the test

danieloprado:
Any updates?

koskimas:
Thank you! One more thing: `commit` and `rollback` return promises, not QueryBuilders. I know this wasn't your mistake, but could you fix that while you're at it? Then I'm ready to merge this.

danieloprado:
No problem!

danieloprado:
I merged #458 to this pull request too

PR #458, opened 2017-07-27
typings: Fix whereIn accepts any queryBuilder

### LINKED ISSUE #456: WhereIn Typings
Before update, the query above is correctly, but now **whereIn** expect a query based on Event this is not right because I can do a subquery in another table.

image

danieloprado:
Fix #456


danieloprado:
Waiting for #457

danieloprado:
Merged with #457

PR #462, opened 2017-07-28
typings: add HasOneThroughRelation

loganhasson:
This type exists in documentation and is required for this type of association to work.


koskimas:
Thanks!

PR #468, opened 2017-08-04
docs: ternary relationships

rafaelkallis:
Added Ternary relationships to docs.

### DIFF
--- a/doc/includes/RECIPES.md
+++ b/doc/includes/RECIPES.md
@@ -800,7 +800,7 @@ class User extends Model {
     return {
       groups: {
         relation: Model.ManyToManyRelation,
-        modelClass: require("Group"),
+        modelClass: require("./Group"),
         join: {
           from: "user.id",
           through: {
@@ -813,7 +813,7 @@ class User extends Model {
       },
       permissions: {
         relation: Model.ManyToManyRelation,
-        modelClass: require("Permission"),
+        modelClass: require("./Permission"),
         join: {
           from: "user.id",
           through: {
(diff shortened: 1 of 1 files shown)


doc/includes/RECIPES.md
```diff
@@ -778,3 +778,232 @@ explain select * from "Hero" where details#>>'{type}' = 'Hero';
...
+// User.js
+const { Model } = require("objection");
+
+class User extends Model {
+  static get tableName() { return "user"; }
+  static get relationMappings() {
+    return {
+      groups: {
+        relation: Model.ManyToManyRelation,
+        modelClass: require("Group"),
```
koskimas: This should be require('./Group') or some other relative path. Not super important for this example, but it's better to be precise

koskimas: Same goes for other relation mappings..

doc/includes/RECIPES.md
```diff
@@ -778,3 +778,232 @@ explain select * from "Hero" where details#>>'{type}' = 'Hero';
...
+UserGroupPermission
+  .query()
+  .select(
+    "first_name",
+    "last_name",
+    "label",
+    "extra_attribute"
+  )
+  .joinRelation("user")
+  .joinRelation("permission")
```
koskimas: joinRelation now takes relation expressions. These two calls can be replaced with `joinRelation('[user, permission]')`

rafaelkallis:
Thanks for the feedback, updated the PR.

PR #471, opened 2017-08-04
Fix `throwIfNotFound` when query result is zero

### LINKED ISSUE #469: query.throwIfNotFound().delete() not working
Hi,
`throwIfNotFound` method is not working with `delete` (and maybe other methods) since the result of deleting an inexistent row is `0`.

Maybe we can change the `throwIfNotFound` method to be like:

```js
  throwIfNotFound() { 
    return this.runAfter((result, builder) => {
      if (Array.isArray(result) && result.length === 0) {
        throw this._modelClass.createNotFoundError(builder.context());
      } else if (result === null || result === undefined || result === 0) {
        throw this._modelClass.createNotFoundError(builder.context());
      } else {
        return result;
      }
    });
  }
```

I'm not sure if there is another case that the result is `0` and we don't want to throw the error.

What do you think?

Thank you

joaonice:
This fixes the `throwIfNotFound` method when the query result is `0` (number of affected rows). 

Closes #469

### DIFF
--- a/lib/queryBuilder/QueryBuilder.js
+++ b/lib/queryBuilder/QueryBuilder.js
@@ -82,7 +82,7 @@ class QueryBuilder extends QueryBuilderBase {
   tableNameFor(modelClass) {
     const ctx = this.internalContext();
     const tableMap = ctx.tableMap;
-    
+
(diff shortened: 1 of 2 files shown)


koskimas:
Brilliant! Thank you!

PR #483, opened 2017-08-20
Add knex withRaw and withWrapped operations

jordansexton:
These methods are useful for building complex CTE queries, especially necessary for row-level locking (`FOR UPDATE ... SKIP LOCKED`).

I didn't see tests for the existing `with` method or other straightforward knex operations, so I didn't include them here.

### DIFF
--- a/lib/queryBuilder/QueryBuilderBase.js
+++ b/lib/queryBuilder/QueryBuilderBase.js
@@ -622,6 +622,20 @@ class QueryBuilderBase extends QueryBuilderOperationSupport {
     );
   }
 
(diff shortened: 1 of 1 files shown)


jordansexton:
Actually, argument argument against this: as `withRaw` and `withWrapped` and not documented by Knex, they may be regarded as private APIs, and you may as well get to them by just passing a raw query or a function.

jordansexton:
It's worth noting that the TypeScript definitions of Objection include `withRaw` and `withWrapped`, so maybe this should be supported.

PR #487, opened 2017-08-23
Missing typings and some fixes

elhigu:


typings/objection/index.d.ts
```diff
@@ -471,6 +471,7 @@ declare namespace Objection {
     // Non-query methods:
 
     context(queryContext: object): this;
+    context(): object;
```
koskimas: Isn't `object` as a return value pretty useless? You cannot access any property through it without first casting it to `any` or some other type. This should return `QueryContext` type.

elhigu: Yes they are pretty useless, I just took the same type out that was passed in in existing `context(queryContext: object): this;` Is there any problems if I change both of them to `QueryContext` type?

koskimas: `object` ans an argument is ok since any object can be downcasted to `object`. The argument shouldn't be `QueryContext` because `QueryContext` type contains some properties that are added by objection and should never be passed to `context` or `mergeContext` methods.

elhigu: Ah right, now I get it. I'll just fix the new API declaration.

PR #500, opened 2017-08-31
Add objection-dynamic-finder plugin

snlamm:
This adds objection-dynamic-finder to the list of plugins.


koskimas:
Awesome idea and plugin! Thank you!

PR #508, opened 2017-09-05
fixed Markdown tables

bennyn:
A few Markdown tables were broken because they only have 3 table headings but 4 columns have been defined.

### Before my fix

1-before

### With my fix

2-after


koskimas:
@bennyn Thanks! The documentation site seems to show these correctly despite the error.

Issue #499, opened 2017-08-31
Typings for raw(), ref() and lit()


elhigu:
Currently these has not been exposed through typescript bindings.

koskimas:
We should add the `raw`, `ref` and `lit` methods to the main object and those should return `RawBuilder`, `ReferenceBuilder` and `LiteralBuilder` instances respectively. There are corresponding javascript classes for those.

I started writing the typings, but I don't know how I should go about it. The typings are a bit confusing to me:

There are types like

```ts
type ColumnName<T>
type Value
```

that list all possible column name and value types, but those are not used in all places. Most methods like `where` have a separate typing for each signature and types are listed there in-place. `ColumnName` and `Value` types are not used there for some reason. Is there a reason for this or should we start using the `ColumnName` and `Value` types everywhere (where they can)?

@mceachen Could you explain this to me?

mceachen:
If those types aren't used in places you think they're applicable, it's most likely a mistake. Do you want me to take a swing at adding those methods? Do you want to share what you've got so far?

Thanks again for objection!

koskimas:
> Do you want me to take a swing at adding those methods?
That would be great! Thank you! I have nothing so far 😄

suubee:
How does one import ref() currently with TypeScript without getting the dreaded``` "no exported member 'ref'"```

mceachen:
@suubee you can use bracket notation, like ```.query()["lit"]```, but that's the stuff of monsters. I'll do this now.

mceachen:
@elhigu , @koskimas if you guys could sketch out examples (here or in a gist or in a pr) that'd be helpful.

mceachen:
Also, in the example

```js
Person.query()
  .select(raw('coalesce(sum(??), 0) as ??', ['age', 'childAgeSum']))
  .where(raw(`?? || ' ' || ??`, 'firstName', 'lastName'), 'Arnold Schwarzenegger')
  .orderBy(objection.raw('random()'));
```

shouldn't the bound params in the where clause be in an array?

```
  .where(raw(`?? || ' ' || ??`, ['firstName', 'lastName']), 'Arnold Schwarzenegger')
```

mceachen:
@suubee unfortunately, `.ref` and `.lit` are a bit more involved. You can temporarily add the type to your project with namespace merging, though: https://www.typescriptlang.org/docs/handbook/declaration-merging.html

mceachen:
[...]

PR #511, opened 2017-09-10
Added .raw to Objection namespace

mceachen:
This PR addresses the first third of #499 

* Added `raw` to the top-level namespace
* Upgraded typescript to 2.5.2 and tslint
* Ran `prettier` on example.ts (which caused most of this diff), which fixed several missing semis and inconsistent parens/whitespace
* Fixed linting errors
* Added `.raw` calls to the bottom of `example.ts` from the docs
* Updated `Where` and `OrderBy` to accept the new examples

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,8 +57,8 @@
     "knex": "0.x"
   },
   "devDependencies": {
-    "@types/knex": "^0.0.55",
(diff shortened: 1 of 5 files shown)


mceachen:
I included this diffs in #512. Closing.

PR #514, opened 2017-09-13
query, $query, and $relatedQuery can take a Knex or Transaction instance

mceachen:
The prior typings only allowed for a transaction object. I just found out that they also accept knex instances.


koskimas:
Awesome! Thanks again 👍

Issue #523, opened 2017-09-26
upsertGraph, should update root object when relational change


johannessjoberg:
When passing an object to `upsertGraph` with only id and relational property, the relational property is not updated. But when adding an extra property which is not relational, this property and the relational property is updated.

I have boiled it down to that this is because the relational property is not considered as a "change" on it's own.

Below are four tests which can verify this behaviour. Is this expected behaviour?

Happy to supply a PR if suitable! Thanks for a great repo.

Objection.js 0.8.9
For verification purposes, add these to:`tests/integration/upsertGraph.js`
```
it('should update belongsToOneRelation with only relational property => fails', () => {
      const upsert = {
        id: 1,
        // belongsToOne: From null => 3
        model1Relation1: { id: 3 },
      }

      return transaction(session.knex, trx => {
        return Model1
          .query(trx)
          .upsertGraph(upsert, { relate: true, unrelate: true })
          .then(result => {
            // Fetch the graph from the database.
            return Model1
              .query(trx)
              .findById(result.id)
              .eager('model1Relation1')
              .select('id')
              .modifyEager('model1Relation1', qb => qb.select('id'));
          })
          .then(omitIrrelevantProps)
          .then(result => {
            expect(result).to.eql({
              id: 1,
              model1Relation1: {
                id: 3,
              }
            });
          });
      }); 
    });

    it('should update belongsToOneRelation with model1Prop1 and relational property => works', () => {
      const upsert = {
        id: 1,
        model1Prop1: 'updated',
        // belongsToOne: From null => 3
        model1Relation1: { id: 3 },
      }

      return transaction(session.knex, trx => {
        return Model1
          .query(trx)
          .upsertGraph(upsert, { relate: true, unrelate: true })
          .then(result => {
            // Fetch the graph from the database.
            return Model1
              .query(trx)
              .findById(result.id)
              .eager('model1Relation1')
              .select('id')
              .modifyEager('model1Relation1', qb => qb.select('id'));
          })
          .then(omitIrrelevantProps)
          .then(result => {
            expect(result).to.eql({
              id: 1,
              model1Relation1: {
                id: 3,
              }
            });
          });
      }); 
    });

    it('should update belongsToOneRelation with only new relation property => fails', () => {
      const upsert = {
        id: 1,
        // belongsToOne: From null => new
        model1Relation1: { model1Prop1: 'new' },
      }

      return transaction(session.knex, trx => {
        return Model1
          .query(trx)
[...]

PR #524, opened 2017-09-26
Issue #523, upsertGraph, should update if relationship property

johannessjoberg:
As discussed in #523.

This impose a breaking change since `$beforeUpdate` and `$afterUpdate` are now called?

### DIFF
--- a/lib/queryBuilder/graphUpserter/UpsertNode.js
+++ b/lib/queryBuilder/graphUpserter/UpsertNode.js
@@ -97,12 +97,17 @@ function hasChanges(currentModel, upsertModel, queryProps) {
 
     if (!relations[key] && currentModel[key] !== upsertModel[key]) {
       return true;
-    } else if (relations[key] && relations[key].constructor.name === BelongsToOneRelation.name) {
-      const idColumn = currentModel.constructor.idColumn;
-      const currentModelBelongsToOneId = currentModel[key] ? currentModel[key][idColumn] : null;
-      const upsertModelBelongsToOneId = upsertModel[key] ? upsertModel[key][idColumn] : null;
+    } else if (relations[key] instanceof BelongsToOneRelation) {
+      const relatedProp = relations[key].relatedProp;
+      const hasChanges = [];
 
-      if (currentModelBelongsToOneId !== upsertModelBelongsToOneId) {
+      for (let i = 0, l = relatedProp.size; i < l; ++i) {
+        const currentProp = currentModel[key] && relatedProp.getProp(currentModel[key], i);
+        const upsertProp = upsertModel[key] && relatedProp.getProp(upsertModel[key], i);
+        hasChanges.push(currentProp !== upsertProp);
+      }
+
+      if (hasChanges.indexOf(true) !== -1) {
         return true;
       }
     }
--- a/lib/queryBuilder/graphUpserter/UpsertNode.js
+++ b/lib/queryBuilder/graphUpserter/UpsertNode.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const isTempColumn = require('../../utils/tmpColumnUtils').isTempColumn;
+const BelongsToOneRelation = require('../../relations/belongsToOne/BelongsToOneRelation');
 
 const UpsertNodeType = {
   Update: 'Update',
@@ -82,6 +83,7 @@ function copyCurrentToUpsert(currentModel, upsertModel) {
 
 function hasChanges(currentModel, upsertModel, queryProps) {
   const keys = Object.keys(upsertModel);
+  const relations = upsertModel.constructor.getRelations();
   const upsertQueryProps = queryProps.get(upsertModel);
 
   // If the upsert model has query properties, we cannot know if they will change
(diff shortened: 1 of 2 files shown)


johannessjoberg:
True!

How about: 
```

File: lib/queryBuilder/graphUpserter/UpsertNode.js [hunk 1]
```diff
@@ -94,7 +93,7 @@ function hasChanges(currentModel, upsertModel, queryProps) {
   for (let i = 0, l = keys.length; i < l; ++i) {
     const key = keys[i];
 
-    if (!relations[key] && currentModel[key] !== upsertModel[key]) {
```
koskimas: I already accidentally wrote this comment in the commit in your fork :D

Doesn't this cause an update query to be made each time a model has any relations because the equality of two objects (or arrays) is always false? This causes a bunch of useless update queries.

tests/integration/upsertGraph.js
```diff
@@ -154,8 +154,8 @@ module.exports = (session) => {
               ]);
             }
 
-            expect(result.$beforeUpdateCalled).to.equal(undefined);
-            expect(result.$afterUpdateCalled).to.equal(undefined);
```
koskimas: Here's an example of an useless update query. The test was here on purpose to make sure useless updates are not made.

lib/queryBuilder/graphUpserter/UpsertNode.js
johannessjoberg: Hehe, and I replied in the conversation :)

True, how about:
`if (relations[key] || currentModel[key] !== upsertModel[key])`

tests/integration/upsertGraph.js
johannessjoberg: Yes, but in this case I believe they should be called since the `model1Relation1` was passed.

I can create a second test that verifies that the aren't called if `model1Relation1` is not passed?

lib/queryBuilder/graphUpserter/UpsertNode.js
johannessjoberg: Last comment was a bit naive of me. I'll look into to it more closely tomorrow.

tests/integration/upsertGraph.js
koskimas: The model should only be updated if there is something to actually update. If there is `model1Relation1` it doesn't mean that it's parent should necessarily get updated. The parent should get updated if there is a `BelongsToOneRelation` that gets related, inserted or deleted which causes the foreign key in the parent to get update, which requires the parent to be updated 😅 If a model has some other relation or a `BelongsToOneRelation` that only gets updated, the parent shouldn't get updated.

lib/queryBuilder/graphUpserter/UpsertNode.js
```diff
@@ -96,6 +97,14 @@ function hasChanges(currentModel, upsertModel, queryProps) {
 
     if (!relations[key] && currentModel[key] !== upsertModel[key]) {
       return true;
+    } else if (relations[key] && relations[key].constructor.name === BelongsToOneRelation.name) {
+      const idColumn = currentModel.constructor.idColumn;
```
koskimas: You can use `relations[key] instanceof BelongsToOneRelation` instead of comparing constructor names.

johannessjoberg: 👍

lib/queryBuilder/graphUpserter/UpsertNode.js
```diff
@@ -96,6 +97,14 @@ function hasChanges(currentModel, upsertModel, queryProps) {
...
+      const currentModelBelongsToOneId = currentModel[key] ? currentModel[key][idColumn] : null;
+      const upsertModelBelongsToOneId = upsertModel[key] ? upsertModel[key][idColumn] : null;
+
```
koskimas: This doesn't handle the case when the relation key is not a primary key (it doesn't have to be). Also this doesn't handle composite keys or json keys. You should use the [Relation](https://github.com/Vincit/objection.js/blob/master/doc/includes/API.md#relation) class's methods (relations[key] is an instance of `Relation`). Those methods are not even documented yet since they have been added after 0.8.

You need to use the `getProp` method of `RelationProperty` class. [Here's an usage example in the code](https://github.com/Vincit/objection.js/blob/1e61de996947d5494aa5ad2cf755fb5676cf9297/lib/queryBuilder/graphInserter/BelongsToOneDependency.js#L12)

I'm sorry, I'm not nitpicking here. Objection is pretty complex beast and there are alot of corner cases that need to be supported.

koskimas: The methods and classes are documented in the `docs` folder, but not in the github.io page.

johannessjoberg: No worries! No need to be sorry :) It's clear you're nowhere near nitpicking.

lib/queryBuilder/graphUpserter/UpsertNode.js
```diff
@@ -96,6 +97,19 @@ function hasChanges(currentModel, upsertModel, queryProps) {
 
     if (!relations[key] && currentModel[key] !== upsertModel[key]) {
       return true;
+    } else if (relations[key] instanceof BelongsToOneRelation) {
+      const relatedProp = relations[key].relatedProp;
+      const hasChanges = [];
+
+      for (let i = 0, l = relatedProp.size; i < l; ++i) {
```
johannessjoberg: Does this also support the json keys case you commented about earlier?

koskimas: Yes. `relation` abstracts away those differences. If you use `relation.getProp` it will select the correct implementation based on the relation type. If the foreign key is a property in a json (or some other non-scalar) column, `getProp` fetches the nested value correctly.

koskimas:
Now everything looks good to me. Thank you so much for this fix! 👍

lib/queryBuilder/graphUpserter/UpsertNode.js
johannessjoberg: Cool! :) Is there any plans/tickets concerning adding these things to the docs? Or are docs only released on certain version bumps?

johannessjoberg:
No worries! Glad to help out :) And thank you for guiding me in the right direction!

Issue #536, opened 2017-10-05
Typings are missing .mixin and .compose


mtpultz:
Typings are for v0.8.4 not v0.8.9 and are missing at least `mixin` and `compose` exports.

koskimas:
The typings usually lag a little behind since they were contributed by the community and are not actively maintained by more active members of the project. I don't use typescript myself, but I'll try to remember to at least add simple things like this in the future.

drew-r:
@mceachen has kindly been maintaining the typings for the most part.

There's also `QueryBuilder.resultSize(): Promise<Number>` but objection returns a string for this as far as I can tell.

mceachen:
I can take this.

@koskimas maybe you can open issues (and at-reference me?) to track adding typings for new features, so typings expose new features as they're released. I'll try to get to them.

@drew-r I'll ping you when my PR is ready so you can make sure I understood you correctly.

mceachen:
> There's also QueryBuilder.resultSize(): Promise<Number> but objection returns a string for this as far as I can tell.

I'm floating off master:

```
$ grep objection package.json
    "objection": "Vincit/objection.js",
```

and I see a number from the promise:

```
 Asset.query().where("foo", 123).resultSize().then(console.dir)
Promise {
  _bitField: 0,
  _fulfillmentHandler0: undefined,
  _rejectionHandler0: undefined,
  _promise0: undefined,
  _receiver0: undefined }
> 0
```

koskimas:
@mceachen I'd love to delegate the typing related tasks to you if you have time to take care of them! I've avoided doing that because I haven't wanted to bother you. I'm really grateful for your contributions!!

mceachen:
I'm happy to help when I can. Feel free to ping me, just know there might be latency in my response.

I'm grateful for objection!

mtpultz:
I don't know if it's helpful for me to point out other missing typings or not, but the chained query() method on models appears to be missing typings as well.

```
User
  .query() // <--- query does not exist on typeof User
  .insert({ username: 'info@example.com', password: 'secret' })
  .then((user: User) => {
    console.log(user.id);
  });
```

mceachen:
@mtpultz there are examples that prove that use case is handled correctly, so you've got an issue with your setup. Does your User model extend Model?

mtpultz:
Thanks @mceachen.  

[...]

PR #537, opened 2017-10-05
Add types for .mixin and .compose

mceachen:
Addresses https://github.com/Vincit/objection.js/issues/536

* added `.compose` and `.mixin` support. YIKES THAT IS SOME UGLY COPYPASTA but that's how you do it, afaik.
* removed unnecessary `this` capture from `.knexQuery` typing
* updated typings and typescript. @koskimas you might want to upgrade the other root package.json deps, many are not on the latest version (`yarn upgrade --latest`).
* added missing semis that tslint grumped about
* added assertions based on @drew-r's concerns (Drew, could you check this diff, because it seems like things are working as expected? I think I'm misunderstanding you, perhaps?)

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   },
   "devDependencies": {
     "@types/knex": "^0.0.61",
-    "@types/node": "^8.0.28",
+    "@types/node": "^8.0.32",
     "coveralls": "^2.13.1",
     "expect.js": "^0.3.1",
     "fs-extra": "3.0.1",
@@ -70,6 +70,6 @@
     "pg": "^6.2.2",
     "sqlite3": "^3.1.8",
     "tslint": "^5.7.0",
-    "typescript": "^2.5.2"
+    "typescript": "^2.5.3"
   }
 }
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -217,7 +217,7 @@ const rowsEager: Promise<Person[]> = Person.query()
   .eagerAlgorithm(Person.NaiveEagerAlgorithm)
   .eager('foo.bar');
 
-const rowsPage: Promise<{ total: number, results: Person[] }> = Person.query().page(1, 10);
+const rowsPage: Promise<{ total: number; results: Person[] }> = Person.query().page(1, 10);
 
 const rowsRange: Promise<objection.Page<Person>> = Person.query().range(1, 10);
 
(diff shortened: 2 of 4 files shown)


koskimas:
@mceachen `compose` and `mixin` don't take model classes, but only mixin functions. A mixin function is  a function that takes a model class as an argument and returns an anonymous subclass of the input. 

```js
const plugin1 = (Model) => {
  return class extends Model {
    foo() {
      console.log('foo');
    }
  }
}

const plugin2 = (Model) => {
  return class extends Model {
    bar() {
      console.log('bar');
    }
  }
}

const BaseModel1 = mixin(Model, [plugin1, plugin2]);
const BaseModel2 = mixin(Model, plugin1, plugin2);

const plugin1AndPlugin2 = compose(plugin1, plugin2);
const BaseModel3 = plugin1Plugin2(Model);
const BaseModel4 = mixin(Model, plugin1Plugin2);

class Person extends BaseModel1 {

}

const p = new Person();
p.foo();
p.bar();
```

mceachen:
Thanks for those examples.

If I just look at those, it seems that you expect `.compose` to only takes plugins, and `.mixin` calls to always starts with `Model`. Is that correct?

I don't understand what you mean by "mixin don't take model classes," because every mixin example you have has a Model in the first position. Do you mean you expect them to take Model subclasses?

Plugins only accept Model classes, correct?

Thanks for being patient with my questions!

mceachen:
@koskimas when you have time, could you answer these questions, please? ⬆️

koskimas:
> If I just look at those, it seems that you expect .compose to only takes plugins, and .mixin calls to always starts with Model. Is that correct?

Yes. `compose` composes multiple plugins into one. `mixin` applies a bunch of plugins to a model (the first argument) and returns the result. `D = mixin(Model, [A, B, C])` is equivalent to `D = C(B(A(Model)))`

> I don't understand what you mean by "mixin don't take model classes," because every mixin example you have has a Model in the first position. Do you mean you expect them to take Model subclasses?

In your examples, you mixed models with other models. I just meant that that is not possible. All other arguments but the first one, need to be plugins.

> Plugins only accept Model classes, correct?

Yes. Plugin is a function that takes a model class and returns a subclass of the input model class.

> I just looked at the three plugins that you reference in the objection docs, and they don't add to the signatures of the classes they are perturbing. I think the typings could get away with not capturing the plugin signatures, in that case, which would make these types much simpler. Are you ok with that?

I'm absolutely ok with that. Correctly typing plugins could be nasty and would require the plugin developers to create correct kind of typings also.

Sorry for the delay.

mceachen:
I'm going to rebuild this PR based on this new understanding of these methods.

mceachen:
Rebuilt here: https://github.com/Vincit/objection.js/pull/546

PR #542, opened 2017-10-08
Fix typings for .fromJson, .fromDatabaseJson, and .findOne

mceachen:
Addresses the following:

* `.fromJson` and `.fromDatabaseJson` should return an **instance** of the Model, not a typeof Model (or ModelClass). Examples were added.

* `.findOne` did not return a QueryBuilderOption, nor did it accept all the parameters of .where() (as the documentation says). Examples were added.

* `.knexQuery` captured a type and did nothing with it. Removed.

* I rewrote some of the example typings tests to be a bit more rigorous, as well.

I'm going to rebase the .mixin and .compose PR off this PR. If you have time to answer the questions in that PR, that'd be great.

Thanks again for objection!

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -15,36 +15,6 @@ class Person extends objection.Model {
   lastName: string;
   examplePersonMethod = (arg: string) => 1;
 
-  static async truncate(): Promise<void> {
-    await this.query().truncate();
-  }
-
-  static async withLastName(lastName: string): Promise<Person[]> {
-    return this.query().where('lastName', lastName);
-  }
-
-  static async firstWithLastName(lastName: string): Promise<Person | undefined> {
-    return this.query()
-      .where({ lastName })
-      .first();
-  }
-
-  static async findById(id: number): Promise<Person | undefined> {
-    return this.query().findById(id);
-  }
-
-  static async findWithFirstName(firstname: string): Promise<Person | undefined> {
-    return this.query().findOne({ firstName: firstname });
-  }
-
-  async loadMovies(): Promise<this> {
-    return this.$loadRelated('movies');
-  }
-
-  async reload(): Promise<this> {
-    return this.$query();
-  }
-
   async petsWithId(petId: number): Promise<Animal[]> {
     // Types can't look at strings and give strong types, so this must be a Model[] promise:
     const pets: objection.Model[] = await this.$relatedQuery('pets').where('id', petId);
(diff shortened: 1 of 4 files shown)


mceachen:
@elhigu @koskimas sorry about the churn here--tell me if there's any changes you'd like to make!

tests/ts/examples.ts
```diff
@@ -57,6 +27,54 @@ class Person extends objection.Model {
   }
 }
 
+const takesVoid = (v: void) => 1;
+
+const takesPerson = (person: Person) => {
+  person.examplePersonMethod("");
+};
+const takesMaybePerson = (_: Person | undefined) => 1;
```
elhigu: This just caught my eye, not sure if its wrong or not. 

Is it really necessary to explicitly allow also passing `undefined`? I thought that undefined would be always allowed (or would that be equivalent with `_?: Person`).

typings/objection/index.d.ts
```diff
@@ -857,6 +861,17 @@ declare namespace Objection {
     (column: ColumnRef, callback: (this: QueryBuilder<T>, queryBuilder: QueryBuilder<T>) => void): QueryBuilder<T>;
   }
 
+  interface FindOne<T> {
```
elhigu: Since findOne should accept anything that where accepts, shouldn't the call signatures be the same with Where<T> with different return value?

Now there seem to be extra methods:

```
(condition: boolean): QueryBuilderOption<T>;
...
(sql: string, ...bindings: any[]): QueryBuilderOption<T>;
(sql: string, bindings: any): QueryBuilderOption<T>;
```

elhigu:
I wrote couple of comments there, I'm not sure if they are implemented intentionally that way or not. Also looks like yarn lockfile was accidentally committed, AFAIK that shouldn't be in repo.

elhigu:
And thank you for the PR 👍  :)

koskimas:
@elhigu I was convinced to add the lockfile [here](https://github.com/Vincit/objection.js/pull/484). And updating it is fine.

typings/objection/index.d.ts
mceachen: Yup, that's the point! [This diff](https://github.com/Vincit/objection.js/commit/559d38df9992a76cbd9213e6e64b8acc0c8cc776) failed to make `.findOne` and `.where` have equivalent argument parameters. Now they match.

tests/ts/examples.ts
mceachen: I wanted to do type verification for methods that optionally return a Person. Both signatures would suffice, either ` (_: Person | undefined) => ...` or ` (_?: Person) =>...`.

mceachen:
I think I addressed @elhigu 's comments. Thanks for looking at it!

Issue #533, opened 2017-10-04
Typings: throwIfNotFound() doesn't have one


kblomster:
And I'm not clever enough to figure out how to add it. I'm pretty new to TypeScript, but I think using it would at the very least change the return type of all the query methods that are currently typed to return `QueryBuilderOption<T>`, so maybe it needs a different `QueryBuilder` type all of its own?

koskimas:
@mceachen Could you take a look at this at some point?

mceachen:
Typescript (afaict) doesn't support interface signatures that declare what errors can be raised. It seems that `.throwIfNotFound()` would be an interface passthrough. Am I correct?

kblomster:
Well, kinda. It's not about typing the exception. Right now e.g. `findById` is typed to return a `QueryBuilderOption<T>`, which means it can return one or zero results (`T | undefined`, https://github.com/Vincit/objection.js/blob/master/typings/objection/index.d.ts#L397). What I would want is that if you use `throwIfNotFound` earlier in the query builder, `findById` would now be expected to return exactly one result (because otherwise it'd raise an error). If I can't get that there's no real point in using `throwIfNotFound` because I'll have to either typecast the `| undefined` away or write my own check for an empty result to get TypeScript to agree with me that the value can't be undefined.

mceachen:
When I did research for this a while back, it seemed that the TypeScript team didn't want to add support for a method to be able to declare `throws <T extends Error>`. With the idea that "Exceptions" should be "exceptional," having missing db records not be exceptional, and coming from scala that had `Some/None/Maybe` support, I've been living with [this Option class](https://gist.github.com/mceachen/ba885d4ef56c35f963fb9c6bb68d5cd4) to handle these maybe-undefined values.

kblomster:
I don't mind not having `throws <T extends Error>`, that's no problem. I do like that options class too (thanks!), but in this case it doesn't do what I want. If I expected there to maybe be a result, it'd do what I wanted, but in my mind the use case for `throwIfNotFound` is to very explicitly declare that not finding this particular db record *is* an exceptional condition that I *don't* want to handle gracefully.

mceachen:
@kblomster the following blew my mind, coming from 30 years of experience with 10 other languages, but make sure you really internalize this: [uncaught exceptions should kill the current process](https://nodejs.org/api/process.html#process_warning_using_uncaughtexception_correctly). 

[...]

PR #547, opened 2017-10-11
Add typings for .throwIfNotFound

mceachen:
Closes https://github.com/Vincit/objection.js/issues/533

* added example tests and typings. This required **every** QueryBuilder flavor to be touched, as `QueryBuilderOption` needed to return a `QueryBuilderSingle`, but every other subclass needed to return `this`. Should we add a new `QueryBuilderArray` base class? Will there be more of these sorts of methods?
* removed `--noEmit` from package.json test-typings script. It wasn't needed, as the `tsconfig.json` already has `noEmit: true`
* replaced `Array<Something>` with `Something[]`, as is preferred by the TypeScript style guide (generics support isn't as solid as arrays)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "test-travis": "istanbul --config=.istanbul.yml cover _mocha -- --slow 100 --timeout 60000 --reporter spec --recursive tests && npm run test-typings",
     "test-bail": "mocha --slow 10 --timeout 15000 --reporter spec --recursive tests --bail",
     "test-opt": "mocha --slow 10 --timeout 15000 --reporter spec --recursive tests --bail --trace_opt --trace_deopt --trace_inlining",
-    "test-typings": "tsc --noEmit",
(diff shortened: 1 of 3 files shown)


mceachen:
@kblomster how's this PR look? Specifically, do the added examples behave as you expect?

mceachen:
I will merge tomorrow if there are no comments.

kblomster:
@mceachen Thanks for doing this!

Overall it does what I want, but there is one issue - `throwIfNotFound` can be used anywhere in the query builder - you can say `Model.query().throwIfNotFound().findOne()`, and that does not change the return type with this PR. Arguably, writing it that way is confusing and I myself wouldn't do it, but it is a thing you can do. This one issue is basically the reason I didn't want to touch this myself, because I can't figure out a non-obnoxious way to do it. :/ You're probably more clever than I am though.

I would be okay with merging this as it is though since it does handle my actual use of the method.

PR #548, opened 2017-10-11
Working example of timestamps

mceachen:
* added the timestamps to the initial schema, rather than a migration, because I didn't want to deal with existing database rows violating a not-null constraint

* added to/from example code to Person

* ran salita --update on package.json and updated yarn.lock

### DIFF
--- a/examples/express-ts/migrations/20150613161239_initial_schema.js
+++ b/examples/express-ts/migrations/20150613161239_initial_schema.js
@@ -7,6 +7,8 @@ exports.up = function (knex) {
       table.string('lastName');
       table.integer('age');
       table.json('address');
+      table.bigInteger("createdAt").notNullable();
+      table.bigInteger("updatedAt").notNullable();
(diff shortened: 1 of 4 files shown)


mceachen:
@koskimas is there a cleaner way you'd recommend to do this? I'd like the example to not lead people astray.

mceachen:
I will merge tomorrow if there are no comments.

jadengore:
@mceachen Why not make this a timestamp column?

mceachen:
@jadengore you absolutely *should* use your rdbms' native timestamp type, if you can. Sqlite's datetime support is a bit exotic, though, acting more like a string than a numeric value, so I'm actually applying this technique for my own sqlite-based project. 

If you want, you can think of this as a contrived example for custom datatype serialization/deserialization with objection.

jadengore:
@mceachen I mostly use Postgres, but cool!

PR #550, opened 2017-10-12
Simplify parseValidationError()

lehni:
This is a simple PR: There is no need to parse `error.message`, we can use `error.params.missingProperty` instead.

`tests\unit\model\Model.js` has tests for this, and they are all still passing with this change.


koskimas:
Thanks!

Issue #549, opened 2017-10-11
Shared Validator


lehni:
Is there a reason why a separate validator object is created for each model through `createValidator()`?

Having a shared AJV validator across all models would offer a few benefits:

* Additional keywords and formats could be registered and used across all JSON schemas, see http://epoberezkin.github.io/ajv/custom.html
* Definitions could be easily shared and referenced through `$ref`

I can already achieve this by returning the same instance in each model's static `createValidator()` method, but I think it's worth considering making this easier.

koskimas:
It is not shared so that you can have different validator for different model. For example if your node app is connected to two different databases, and both need to have different Model hierarchies that use different validation, you can do that.

Can't you simple define a shared base class, say `BaseModel`, and define your custom definitions and keywords there?

lehni:
Yes, that makes sense, thanks!

lehni:
Actually, there are two things still that make sharing a validator and definitions of keywords and formats difficult:

- Caching of non-changing schemas all happen under the key `'default'`, thus they override each other with multiple models sharing the same validator. I would recommend using the `constructor.name` or `tableName` here instead. Happy to create a PR for it.

- There are two instances of `ajv` created, one called `ajvNoDefaults` and one `ajv`. If I add keywords or formats to be used in the json schema, I have to add them to both.

The 1st is easy to address, the 2nd has me wondering a bit. Isn't this perhaps something that really should be addressed in AJV, so one validator can handle both scenarios?

lehni:
And as an example why I would like to make shared validators work, beyond the mentioned additional keywords and formats, is that the use of `$ref` allows for a lot of interesting scenarios. Here a simple example:

#### Schema for Author model
```json
{
  "id": "Author",
  "type": "object",
  "properties": {
    "id": {
      "type": "number",
    },
    "name": {
      "type": "string"
    },
    "comments": {
      "type": [ "array", "null" ],
      "items": { "$ref": "Comment" }
    }
  }
}

#### Schema for Comment model
````js
{
  "id": "Comment",
  "type": "object",
  "properties": {
    "id": {
      "type": "number",
    },
    "comment": {
      "type": "string"
    },
    "author": {
      "$ref": {
        "anyOf": [
          { "$ref": "Autho" },
          { "type": "null" }
        ]
      }
    }
  }
}
````

[...]

PR #551, opened 2017-10-12
Use model name or table name for validator caching

lehni:
This simple change allows one AjvValidator instance to be shared across multiple models.
See #549 for further discussions.


koskimas:
Looks good to me. Thanks!

PR #567, opened 2017-10-22
Support insertMissing option in upsertGraph() for root models

lehni:
As discussed on [Gitter](https://gitter.im/Vincit/objection.js?at=59ec719532e080696e0eb50a), the `insertMissing` option in `upsertGraph()` currently does not work for root models.

This patch addresses it by leveraging the same mechanism already in place for child models.

I've started adding unit tests for `insertMissing: true` on root models, but am not sure if I'm doing it right, and also wanted to first check if this fix is the right way to go about the issue.

### DIFF
--- a/tests/integration/upsertGraph.js
+++ b/tests/integration/upsertGraph.js
@@ -969,7 +969,47 @@ module.exports = (session) => {
             model2Relation1: []
           }]
         });
-      });
+      })
+    });
+
+    it('should insert root model with an id instead of throwing an error if `insertMissing` option is true', () => {
+      let upsert = {
+        // This doesn't exist.
+        id: NONEXISTENT_ID,
+        model1Prop1: `updated root ${NONEXISTENT_ID}`,
+
+        model1Relation1: {
+          model1Prop1: 'inserted belongsToOne'
+        }
+      };
+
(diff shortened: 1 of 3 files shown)


tests/integration/upsertGraph.js
```diff
@@ -969,7 +969,40 @@ module.exports = (session) => {
...
         });
-      });
+      })
+    });
+
+    it('should insert root model with an id instead of throwing an error if `insertMissing` option is true', () => {
+      const upsert = {
+        // This doesn't exist.
+        id: 3,
+        model1Prop1: 'updated root 3',
```
koskimas: Actually model with id 3 does exist. See the inserted object in `beforeEach`. Use some ridiculous number like 1234567 or the `NONEXISTENT_ID` constant instead.

koskimas:
This seems like a good approach if it works. The test fails currently. I added one comment that may be the cause.

lehni:
@koskimas I've addressed this now and also changed the test to run the upsert twice, first to insert the new root, then to upsert it. Let's see how this performs on Travis CI...

koskimas:
Awesome! Thank you!

PR #578, opened 2017-10-24
Support named filters on relations

lehni:
It would be nice if named filters could also be used on relations. This PR addresses that, by looking them up on the related model class.

### DIFF
--- a/lib/relations/Relation.js
+++ b/lib/relations/Relation.js
@@ -340,14 +340,25 @@ function createRelationProperty(ctx, refString, propName) {
 
 function parseModify(ctx) {
   const mapping = ctx.mapping;
+  const value = mapping.modify || mapping.filter;
+  let modify;
 
-  let modify = mapping.modify || mapping.filter;
-  let type = typeof modify;
-
-  if (modify && type === 'object') {
-    const obj = modify;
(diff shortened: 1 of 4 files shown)


koskimas:
Nice one! Could you add a simple test case for this also? And a mention to the docs would also be nice.

lehni:
Yes I will work on both, along with the other pending ones... Just wanted to see what you thought first.

lehni:
@koskimas I'm struggling to find a good place to put a test for this. There's also no real test for `modify` yet, so I can add that one too.

I was wondering if they should go here: https://github.com/Vincit/objection.js/blob/master/tests/integration/schema.js

koskimas:
This should be tested in these files:

https://github.com/Vincit/objection.js/blob/master/tests/unit/relations/BelongsToOneRelation.js
https://github.com/Vincit/objection.js/blob/master/tests/unit/relations/HasManyRelation.js
https://github.com/Vincit/objection.js/blob/master/tests/unit/relations/ManyToManyRelation.js

Check out the existing `filter:` and `modify:` tests

lehni:
@koskimas done: b0165408d94687a23f82a7e893ce337519f34e41

koskimas:
Looks good, thanks!

PR #582, opened 2017-10-25
Move filter handling from modify to applyFilter

lehni:
As discussed on [Gitter](https://gitter.im/Vincit/objection.js?at=59f01c8ad6c36fca3192f499), here now the PR with documentation

### DIFF
--- a/lib/queryBuilder/QueryBuilder.js
+++ b/lib/queryBuilder/QueryBuilder.js
@@ -131,19 +131,19 @@ class QueryBuilder extends QueryBuilderBase {
     return this;
   }
 
-  modify() {
-    if (typeof arguments[0] === 'string') {
-      const namedFilters = this._modelClass.namedFilters;
-
-      for (let i = 0, l = arguments.length; i < l; ++i) {
-        const filter = namedFilters[arguments[i]];
-        filter(this);
-      }
-
-      return this;
-    } else {
-      return super.modify.apply(this, arguments);
+  applyFilter() {
+    const namedFilters = this._modelClass.namedFilters;
+    for (let i = 0, l = arguments.length; i < l; ++i) {
+      const filter = namedFilters[arguments[i]];
+      filter(this);
     }
+    return this;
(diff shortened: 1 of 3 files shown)


lib/queryBuilder/QueryBuilder.js
```diff
@@ -130,19 +130,19 @@ class QueryBuilder extends QueryBuilderBase {
...
+      const filter = namedFilters[arguments[i]];
+      filter(this);
     }
+    return this;
+  }
+
+  modify() {
+    return typeof arguments[0] === 'string'
+      ? this.applyFilter.apply(this, arguments)
+      : super.modify.apply(this, arguments);
```
lehni: I don't know if you dig ternary operators or not... I do, as you can see here : )
Happy to rewrite it with if / else if you prefer.

koskimas:
Looks great. This needs to have at least one test somewhere. Then we're good to go.

lib/queryBuilder/QueryBuilder.js
koskimas: I also like ternary operators when they make stuff simpler. I wish ecmascript handled `if`, `try` etc. blocks as expressions that return a value like kotlin does:
```js
const foo = if (x < 10) "bar" else "baz"
const bar = try {
  doStuff();
} catch (err) {
  null
}
```

lehni: Yeah that would be nice. CoffeeScript does this also, and I believe Ruby and a bunch of others too.

lehni:
@koskimas done: 95a5d0cd45c8cafcf053441e8abfd57a70dfa860

koskimas:
Looks good, thanks!

lehni:
@koskimas I've been thinking, wouldn't `mergeFilter()` have been the better name for this? 

There is already `mergeEager()`, and the effect of applying a filter is similar: Previous filters don't get erased, they get "merged"...

PR #583, opened 2017-10-25
Simplify UpsertNode by the addition of decideType

lehni:
This is a small PR. While looking at `upsertGraph()` I realised that this could be simplified a bit. Not sure it's to your licking though, @koskimas

### DIFF
--- a/lib/queryBuilder/graphUpserter/UpsertNode.js
+++ b/lib/queryBuilder/graphUpserter/UpsertNode.js
@@ -80,17 +80,9 @@ function getType(node) {
 
   if (node.upsertModel !== null && node.currentModel === null && node.upsertModel.$hasId()) {
     if (hasOption(node, 'relate') && isInRelation) {
-      if (hasOption(node, 'noRelate')) {
-        return UpsertNodeType.None;
-      } else {
-        return UpsertNodeType.Relate;
-      }
+      return decideType(node, UpsertNodeType.Relate, 'noRelate');
     } else if (hasOption(node, 'insertMissing')) {
-      if (hasOption(node, 'noInsert')) {
-        return UpsertNodeType.None;
-      } else {
-        return UpsertNodeType.Insert;
-      }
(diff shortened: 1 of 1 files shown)


lib/queryBuilder/graphUpserter/UpsertNode.js
```diff
@@ -154,6 +128,10 @@ function hasOption(node, optName) {
   }
 }
 
+function decideType(node, defaultType, option, optionType = UpsertNodeType.None) {
```
koskimas: Node 4 and 5 don't support default values.

koskimas:
Good simplification. I'll merge this if you fix the default value issue I commented on.

lib/queryBuilder/graphUpserter/UpsertNode.js
lehni: Oh I keep forgetting that you still target these versions. Will adjust now.

lib/queryBuilder/graphUpserter/UpsertNode.js
```diff
@@ -154,6 +128,12 @@ function hasOption(node, optName) {
   }
 }
 
+function decideType(node, defaultType, option, optionType) {
+  return hasOption(node, option)
+    ? optionType || UpsertNodeType.None
+    : defaultType;
+}
+
```
lehni: @koskimas optionType is always null currently, but I was going to use this to implement the `update: true` option also, where it will have another value.

lehni: @koskimas currently, `optionType` is never provided so this could be simplified. But I was going to use this also to support the `update: true` option...

lehni: @koskimas optionType is always null currently, but I was going to use this to implement the `update: true` option also, where it will have another value.

lehni:
@koskimas the issue with the default value is fixed.

Issue #557, opened 2017-10-16
Add patchGraph() and maybe updateGraph() methods?


lehni:
I would like to suggest the addition of `patchGraph()`that is similar to `upsertGraph()`, but would receive a sub-set of the full graph and just patch, without creating new instances in the graph, and throwing an error if data for new instances was included.

Perhaps it is then also worth considering `updateGraph()`, as a version of `upsertGraph()`that would just update and not create new instances? 

And in symmetry to #556, it would then be great to then also add `*AndFetch()` versions of these: `patchGraphAndFetch()` and `upsertGraphAndFetch()`.

koskimas:
I'm going to add options for `upsertGraph` that allow you to select any subset of `[insert, update, delete, relate, unrelate]` to be selected for an `upsertGraph` call. This case will be covered by only enabling `update`. Let's continue this discussion in #526

lehni:
@koskimas there is a difference between update and patch, yes? would this cover both use-cases?

koskimas:
`upsertGraph` always uses `patch`. It's possible to add an option for that too.

lehni:
Thanks for clarifying! Those options sounds perfect, BTW. Exactly what I need. That and the `upsertGraphAndFetch()` variant : )

Issue #527, opened 2017-09-29
upsertGraph, should update property when relating new association


johannessjoberg:
When creating a new relationship (null => some id) for a resource that already exists in the database `upsertGraph` is not updating passed properties. But when the relation already exists it's possible update the properties on the child as expected.

Verification tests (add to `/tests/integration/upsertGraph.js`):
```
    it.only('should relate and update property of belongsToOne relation (fails)', () => {
      const upsert = {
        id: 1,
        // update BelongsToOneRelation. null => 3 and model1Prop1 change
        model1Relation1: { id: 3,  model1Prop1: 'updated belongsToOne' },
      };

      return transaction(session.knex, trx => {
        return Model1
          .query(trx)
          .upsertGraph(upsert, { relate: true })
          .then(result => {
            expect(result.$beforeUpdateCalled).to.equal(1);
            expect(result.$afterUpdateCalled).to.equal(1);

            expect(result.model1Relation1.$beforeUpdateCalled).to.equal(1);
            expect(result.model1Relation1.$afterUpdateCalled).to.equal(1);

            return Model1
              .query(trx)
              .findById(result.id)
              .eager('model1Relation1')
              .then((result2) => {
                expect(result2.model1Relation1.model1Prop1).to.equal('updated belongsToOne');
              });
          });
      });
    });

    it.only('should update property of belongsToOne relation (works)', () => {
      const upsert = {
        id: 2,
        // update BelongsToOneRelation. 3 => 3(no change), but model1Prop1 change
        model1Relation1: {
          id: 3,
          model1Prop1: 'updated belongsToOne',
        },
      };

      return transaction(session.knex, trx => {
        return Model1
          .query(trx)
          .upsertGraph(upsert, { relate: true })
          .then(result => {
            expect(result.$beforeUpdateCalled).to.equal(undefined);
            expect(result.$afterUpdateCalled).to.equal(undefined);

            expect(result.model1Relation1.$beforeUpdateCalled).to.equal(1);
            expect(result.model1Relation1.$afterUpdateCalled).to.equal(1);

            return Model1
              .query(trx)
              .findById(result.id)
              .eager('model1Relation1')
              .then((result2) => {
                expect(result2.model1Relation1.model1Prop1).to.equal('updated belongsToOne');
              });
          });
      });
    });
```

The behaviour is the same for both `BelongsToOne` and `ManyToMany` relations.

So is this expected behaviour? The docs are a little vague concerning this case.

[...]

PR #584, opened 2017-10-25
Add option to use update instead of patch in graph upserts

lehni:
As discussed [before](https://github.com/Vincit/objection.js/issues/557#issuecomment-336901820), here now the PR to add this.

This is a follow up PR for #583, I will rebase once that one is merged.

Please note that there are things to be figured out still, as outlined in this comment:

```js
// TODO: Should we distinguish between update and patch for noUpdate also?
// If we want different options, then:
// return hasOption(node, 'update')
//   ? decideType(node, UpsertNodeType.Update, 'noUpdate')
//   : decideType(node, UpsertNodeType.Patch, 'noPatch');
```

Note that `'noUpdate'` in the current code base would become `'noPatch'` if we did this.


lib/queryBuilder/graphUpserter/UpsertNode.js
```diff
@@ -131,15 +110,18 @@ function getType(node) {
...
-        return UpsertNodeType.Update;
-      }
+      return decideType(node,
+        hasOption(node, 'update') ? UpsertNodeType.Update : UpsertNodeType.Patch,
+        'noUpdate')
+      // TODO: Should we distinguish between update and patch for noUpdate also?
+      // If we want different options, then:
+      // return hasOption(node, 'update')
+      //   ? decideType(node, UpsertNodeType.Update, 'noUpdate')
+      //   : decideType(node, UpsertNodeType.Patch, 'noPatch');
```
koskimas: I don't think there is any need for separate `noUpdate` and `noPatch`.

koskimas:
This still needs tests and documentation.

lehni:
@koskimas yes I was waiting with this until the open question was decided. I just rebased and merged with #527, will work on tests this week.

lehni:
@koskimas I can't figure out how to test if patch is used instead of update when passing the option... It looks like the only difference is really in validation (through `modelOptions` that get passed on), so I guess I could cause a validation error that only happens when doing an incomplete update instead of a patch. But how to achieve this in `tests/integration/upsertGraph.js`?

lehni:
I've rebased again now and merged all conflicts. I hope you're not planning another general large-scale change of code conventions any time soon, @koskimas :)

koskimas:
Sorry about that @lehni. I merged all your PR's that I was able before running `prettier`.

To answer your earlier question, you can update a model's jsonSchema like [this](https://github.com/Vincit/objection.js/blob/master/tests/integration/insertGraph.js#L122). It's a bit hacky, but gets the job done.

lehni:
@koskimas I've tried to achieve this test for the past hour, but failed... Could you please take a look? I've pushed it here now. From my own tests separately, a `upserGraph(..., { update: true })` does fail validation if the provided data is not complete. I don't understand why in the test this isn't happening, but maybe it has something to do with internal caching and the fact that another test with another `jsonSchema` runs right before it?

tests/integration/upsertGraph.js
```diff
@@ -1777,6 +1778,110 @@ module.exports = session => {
...
+        delete Model1.$$jsonSchema;
+        delete Model1.$$validator;
+      });
+
+      it('should fail to do an incomplete upsert', () => {
+        const fails = [
+          {
+            id: 2,
+            // This fails because of missing property.
+            // model1Prop1: 'updated root 2',
```
koskimas: There's an optimization that no updates are done if there is only an identifier present. Your code never reaches the validation line. A better way to test this would be to set the `model1Prop1` and give it a value of wrong type.

tests/integration/upsertGraph.js
```diff
@@ -1777,6 +1778,110 @@ module.exports = session => {
...
+              model1Prop1: 'updated belongsToOne'
+            }
+          }
+        ];
+
+        return Promise.map(fails, fail => {
+          return transaction(session.knex, trx =>
+            Model1.query(trx).upsertGraph(fail, {update: true})
+          ).reflect();
+        })
```
koskimas: Sigh... I hate the way prettier aligns code with these. Well, there nothing we can do..

lehni: I guess you can configure it?

tests/integration/upsertGraph.js
lehni: Oh I see, that makes sense. But I need to catch the case where an update does not provide all required properties and hence causes that kind of validation error, which a patch wouldn't cause due to the use of `skipRequired` in compileValidator() (which as far as I can tell is the only difference between patch and update?)

tests/integration/upsertGraph.js
koskimas: Nope, the there is not much to configure in prettier.

tests/integration/upsertGraph.js
koskimas: The optimization is in `UpsertNode.hasChanges` it tests if any of the values has actually changed. Since we only fetch ids and foreign keys from the db, the test only compares those. I think the way it currently works is correct. If there is nothing to update, even if the model is invalid, nothing should be done.

koskimas: We can't move the validation to an earlier position, since some properties get set by the insert part. The validation must be done after insert, but before update.

lehni: Not saying it's not correct. I guess I'd need a model with more than one property to create an incomplete update?

lehni:
@koskimas I believe I've figured it out now: a8af40cd9405d0089761c13b9b505186b4e3fed5

Using two props isntead of one like in the test right before it allows me to produce and test the described scenarios

tests/integration/upsertGraph.js
lehni: That's why I use a heavily configured ESLint with precommit hooks instead of prettier. Prettier is great, but for me there are too many edge cases where it behaves a tad too weirdly.

lehni:
@koskimas ooops the tests were still failing because deleting `$$jsonSchema` doesn't get rid of the cached `$$validator`. I have to delete both after the first test that overrides `$$jsonSchema` and validates. I've added it in both places now, but you may want to look into caching, since changin `$$jsonSchema` should cause the cache to invalidate, perhaps?

tests/integration/upsertGraph.js
```diff
@@ -1551,6 +1551,7 @@ module.exports = session => {
 
       after(() => {
         delete Model1.$$jsonSchema;
+        delete Model1.$$validator;
```
lehni: @koskimas this is the addition that was necessary for the following test to get a new validator.

lehni:
Oh wow, now it's failing on postgres. That test doesn't work for me locally, not sure how to get it set up?

```
integration tests postgres upsertGraph upserts with update: true option should fail to do an incomplete upsert:
     error: update "Model1" set "id" = $1, "model1Id" = $2, "model1Prop1" = $3, "model1Prop2" = $4 where "Model1"."id" = $5 - invalid input syntax for integer: "updated root 2, prop 2"
```

lehni:
Oh I see, that property is an integer, not  a string there. That's easy, I'll change the test accordingly.

lehni:
@koskimas ok, I'm pretty confident that this one will turn green... 10th time lucky, or something? Talk of a tiny feature, with a unit test that blows out of proportion!

tests/integration/upsertGraph.js
koskimas: Oh crap! I think I've made the same mistake in other tests too. I'll check them out. Good catch!

koskimas:
Looking good! Could you add a short description of the `update` option in `UpsertGraphOptions` object's documentation in API.md? Then I'm ready hit merge.

lehni:
Will take care of it tomorrow!

lehni:
@koskimas I've pushed the documentation. PTAL

koskimas:
Awesome++ Thank you so much for fighting with this!

lehni:
The feeling's mutual! : )

PR #585, opened 2017-10-25
Add reduce method to queryBuilder.

### LINKED ISSUE #571: reduce is missing
Knex has the option for a reduce operation on the query, directly reducing all records:

```js
knex.from('table').reduce((result, record) => {
  return `some complex operation on the entire record`
})
```

This doesn't work with Objection:

```js
Model.query().reduce(...)
```

Results in:
```
reduce is not a function
```

It seems to be missing from the Objection QueryBuilder. Now of course I could first map all records, which returns a promise and then apply (bluebird) reduce on that promise but I have some complex cases where there is no purpose for a mapping operation and I directly want to reduce the record into something. It makes no sense to add a bodus: `.map(x => x)` inbetween.

juanjoLenero:
Resolves #571.

I mirrored `map`'s implementation, let me know if anything is missing.

### DIFF
--- a/doc/includes/API.md
+++ b/doc/includes/API.md
@@ -3876,6 +3876,32 @@ Type|Description
 
 
 
+#### reduce
+
+```js
+const promise = queryBuilder.reduce(reducer, initialValue);
+```
+
+Executes the query and calls `reduce(reducer, initialValue)` for the returned promise.
+
+##### Arguments
+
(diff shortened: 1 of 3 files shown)


koskimas:
Awesome! Thank you!

koskimas:
Nothing missing 😄

appedemic:
Thanks!

PR #573, opened 2017-10-23
Add upsertGraphAndFetch method

lehni:
As discussed #556, here the PR to add `upsertGraphAndFetch()`

I still need to add unit tests to this, but from my own testing this all works as expected!

Issue #572, opened 2017-10-23
Omitting babel-plugin-transform-es2015-classes


shtse8:
https://github.com/Vincit/objection.js/tree/887bc424145dd7ca1639063b8224dedf935f5960/examples/express-es7
As the document stating that 
> For this reason we need to omit the babel-plugin-transform-es2015-classes plugin. There is no way to omit plugins from presets.

We can exclude the plugin in babel as follow in babel-preset-env:
```
["env", {
      "targets": {
        "browsers": ["> 1%", "last 2 versions", "not ie <= 8"]
      },
      exclude: [
        "babel-plugin-transform-es2015-classes"
      ]
    }],
```

I think we can have a note in readme for others.

koskimas:
You are right, I should update the example project to use the "env" plugin. By the way, this should be enough:

```
{
  "presets": [
    ["env", {
      "targets": {
        "node": "current"
      }
      }],
    "stage-1"
  ],
  "plugins": []
}
```

AnianZ:
I tried to update the examples `package.json` to the suggested presets as a first contribution to the project. I hope it is working as expected.

koskimas:
@AnianZ Fixed this in 3867ab5. Thank you!

PR #588, opened 2017-10-29
Cleared up babel plugins and presets in express-es7-examples

AnianZ:
I tried to apply the enhancement described in issue #573 as a first contribution to the project. I added the `env` preset to the `package.json` and removed the explicit plugin imports and it still seems to work.

I also removed the note about the imports in the README.md

### DIFF
--- a/examples/express-es7/README.md
+++ b/examples/express-es7/README.md
@@ -8,13 +8,6 @@ npm install
 npm start
 ```
 
-Note that starting from 0.8.0 objection uses native ES2015 classes and no longer supports
-legacy ES5 classes. This also means Babel generated ES5 code. For this reason we need to
-omit the `babel-plugin-transform-es2015-classes` plugin. There is no way to omit plugins
-from presets. That's why the package.json explicitly lists all plugins in Babel `es2015`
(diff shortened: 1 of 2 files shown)


examples/express-es7/package.json
```diff
@@ -3,45 +3,26 @@
...
-      "transform-async-to-generator"
+      [
+        "env",
+        {
+          "targets": {
+            "node": "current"
+          }
+        }
+      ],
+      "stage-1"
```
AnianZ: The original `package.json` used `stage-0` but @koskimas used `stage-1` in his example in #572 so I changed it here.

examples/express-es7/README.md
```diff
@@ -8,13 +8,6 @@ npm install
 npm start
 ```
 
-Note that starting from 0.8.0 objection uses native ES2015 classes and no longer supports
```
AnianZ: Is it ok to remove this note after changing to the `env` preset? Or should I change it to inform users about that.

koskimas: Yes, I think this can be removed now.

PR #596, opened 2017-11-05
Add createValidationError() method to AjvValidator

lehni:
This simple change allows the overriding of the default error parsing behaviour. I could really use this in my app to have more fine-grained control over how validation errors are handled and displayed.

### DIFF
--- a/lib/model/AjvValidator.js
+++ b/lib/model/AjvValidator.js
@@ -68,10 +68,15 @@ class AjvValidator extends Validator {
     }
 
     if (validator.errors) {
-      throw parseValidationError(validator.errors, model.constructor);
+      throw this.createValidationError(model.constructor, validator.errors);
     }
 
     return json;
+  }
+
+  createValidationError(modelClass, errors) {
+    const errorHash = parseValidationError(errors, modelClass);
+    return modelClass.createValidationError(errorHash);
   }
 
   getValidator(ModelClass, jsonSchema, skipRequired) {
@@ -150,7 +155,7 @@ function parseValidationError(errors, modelClass) {
     ];
   }
 
-  return modelClass.createValidationError(errorHash);
+  return errorHash;
 }
 
 function hasDefaults(obj) {


koskimas:
I have said this before but here it is again 😄 : you shouldn't use private methods! They can and will change at a random point in a random 0.0.1 version change and then your app is broken. I can merge this, but It will not prevent me from changing that method without any notice at any given time.

lehni:
@koskimas maybe I should explain the reason for this a bit:

I am only proposing to expose it so I can override it and prevent the default parsing of validation errors and override it with my own. the other option is to override the already exposed `validate(args)`which at the end calls `createValidationError()`, but I'd rather not.

You already have such a hook in `Model`, I thought it would be consistent to add one here too. I'll restructure the commit a bit to make this clearer.

koskimas:
@lehni I think you missed my point. Everything that's not documented in the doc site is considered private implementation and I reserve the right to change whenever. Even if I merge this it is a very bad idea to override the new method because you are relying on an unstable private method.

The only way you should use that method would be to make it public which means proper documentation and tests. Then it would be part of the public API and it wouldn't get changed. Accidentally changing it would be caught by the tests.

You would need to create docs for the `AjvValidator` class since there are none at the moment (there is only a mention that such class exists and that it implements the `Validator` interface). Your method should not be added to `Validator`.

Having said that, I'm not sure this should be added to the public API.

lehni:
@koskimas ok I get your point. It is true that I misunderstood your use of private vs public, as in JS land this can mean a lot of things. Through my previous PRs I have gotten used to reading the source code directly instead of publicly available API docs, and since there is a `createValidator()` method on the `Model` class I assumed it is fine to override the default validator.

I am not doing much in my override that I can't do just through the configuration options passed to `AjvValidator`, mainly through the `onCreateAjv()` callback. And I just realised that I wrongly assumed that `parseValidationError()` removes some information validation errors that I'd need for post-processing, but realized now that it's still there in `params`.

So I will rewrite my code to not override `AjvValidator`, and we can probably close this PR.

But before we do, I'd like to still discuss one use-case not currently covered:

I am running Ajv validations elsewhere in my code with data other than model instances, and it would be really nice to be sure that it is post-processed the same way as the internal data. Isn't there any way that this could be exposed?

koskimas:
It is ok to override the default validator. Thst is documented and tested. It is not ok to rely on the method you added in this pull request, which is not documented or tested.

lehni:
Ok. I just wanted to avoid having to imitate the way Objection parses validation errors if possible. But I see your point, will close this now.

lehni:
Actually, the main problem is that my other validation schemas can also reference model data schemas through `$ref` as discussed earlier in another issue. If I create a new instance of Ajv, then it doesn't see these definitions. There are ways around that too, I guess I just got carried away by how close and easy this seemed to achieve...

lehni:
@koskimas some good news here: I managed to remove all dependencies on private APIs and maintain functionality 🎉

PR #602, opened 2017-11-07
Add GitHub templates for issues and pull requests

lehni:
Based on the suggestion on  [Gitter](https://gitter.im/Vincit/objection.js?at=5a01636032e080696e73e9a0) earlier today, here a proposal of templates for new issues and pull requests, to reduce the amount of time @koskima has to spend answering things that can be avoided or for which the community can help out also.

Please feel free to completely change the wording or reject the PR all together : )

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,39 @@
+<!--
+This issues database is used to keep track of bugs and new features.
+For questions and support, please visit the Gitter channel instead:
+  https://gitter.im/Vincit/objection.js
+-->
+
+# Description/Steps to reproduce
+
+<!--
+If a feature: A description of the feature.
+If a bug: The steps to reproduce the issue.
+-->
+
+# Link to reproduction test-case
+
+<!--
+Please link to a test-case for reproduction, ideally as a repository with a
+`packages.json` that installs all required dependencies to reduce confusion.
+
+Note: Failure to provide a test-case for reproduction purposes will result in
+the issue being closed.
+-->
+
+# Expected result
+
+<!--
+Also include actual results when reporting a bug.
+-->
+
+# Additional information
+
+<!--
+Please include the type and versions of Operating System, Node, as well as
+the underlying database that the issue is encountered on.
+
+Examples:
+  macOS 10.12.6, Node 8.9.0, PostgreSQL 10.0
+  Windows 10 Pro 10586.962, Node 8.8.1, SQLite 3
+-->
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,37 +1,37 @@
+# Type of issue
 <!--
-This issues database is used to keep track of bugs and new features.
-For questions and support, please visit the Gitter channel instead:
-  https://gitter.im/Vincit/objection.js
+Is this a *bug* report, a *feature* request, or a *question* or request for *support*?
 -->
 
-# Description/Steps to reproduce
+# Description / Steps to reproduce
 
 <!--
-If a feature: A description of the feature.
-If a bug: The steps to reproduce the issue.
+- For features: A description of the feature.
+- For bugs: The steps to reproduce the issue.
 -->
 
-# Link to reproduction test-case
+# Steps to reproduce / Link to reproduction test-case
 
 <!--
-Please link to a test-case for reproduction, ideally as a repository with a
-`packages.json` that installs all required dependencies to reduce confusion.
+Please link to a test-case for easier reproduction of the issue, for example a RunKit playground or a repository with a `packages.json` that installs all required dependencies and is able to execute the script.
 
-Note: Failure to provide a test-case for reproduction purposes will result in
-the issue being closed.
+Note: Failure to provide a test-case for reproduction purposes will result in the issue being closed.
 -->
 
 # Expected result
-
 <!--
-Also include actual results when reporting a bug.
+Please describe the expected results.
+-->
+
+# Actual result
+<!--
+When reporting a bug, also include the actually encountered results.
 -->
 
 # Additional information
 
 <!--
-Please include the type and versions of Operating System, Node, as well as
-the underlying database that the issue is encountered on.
+Please include the type and versions of Operating System, Node, as well as the underlying database that the issue is encountered on.
 
 Examples:
   macOS 10.12.6, Node 8.9.0, PostgreSQL 10.0
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,11 +17,10 @@ Please list related issues and discussion by using the following syntax:
 ### Checklist
 
 <!--
-- Please mark your choice with an "x" (i.e. [x], see
-https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments)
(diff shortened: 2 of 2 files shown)


lehni:
Travis fails on this one, but it can't be the templates I've added. It looks like Prettier jokes on a bunch of files.

koskimas:
Master builds without a problem. Maybe you need to rebase?

koskimas:
In general I think templates like these are needed. Now that objection is growing faster, most of my time is spent on trying to understand issues and prying out relevant information from the reporters. What we need even more is contributors! People to help me answer the issues. Can you whip up a pull request for those 😄 @VladShcherbin May I ask why the confused emoji?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
+<!--
+This issues database is used to keep track of bugs and new features.
+For questions and support, please visit the Gitter channel instead:
+  https://gitter.im/Vincit/objection.js
+-->
```
koskimas: I think questions and support can still be handled in the issues. They get lost pretty fast in gitter.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
...
+<!--
+If a feature: A description of the feature.
+If a bug: The steps to reproduce the issue.
+-->
+
+# Link to reproduction test-case
+
+<!--
+Please link to a test-case for reproduction, ideally as a repository with a
+`packages.json` that installs all required dependencies to reduce confusion.
```
koskimas: Also a test case for objection's test suite would be execellent and a standalone file is fine. runkit is also a great option.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+### Checklist
+
+<!--
+- Please mark your choice with an "x" (i.e. [x], see
+https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments)
+- PR's without test coverage will be closed.
+-->
+
+- [ ] New tests added or existing tests modified to cover all changes
+- [ ] Code conforms with the ESLint and Prettier rules (`yarn eslint` passes
```
koskimas: I think we should use `npm` examples instead of `yarn` here.

.github/ISSUE_TEMPLATE.md
koskimas: They should be tagged as questions and in the future they will receive less attention, since the amount of issues is growing.

koskimas:
It seems that `prettier` was updated and old version creates different results than the new one. Lint passes for master with the old prettier, but not with the new one 😄

koskimas:
Rebasing with master should now fix the prettier errors.

.github/ISSUE_TEMPLATE.md
koskimas: I was thinking that a reproduction as a repository is probably too much to ask and people would just forget the bug and work around it most of the time. We could add a standalone file template to the project that can be copy-pasted and with small modifications, the bug could be reproduced?

.github/ISSUE_TEMPLATE.md
lehni: Yes I agree, but I don't think people who post issues can tag them by themselves? The first question could be "what kind of issue is this" similar to how yarn handles it:

https://github.com/yarnpkg/yarn/issues/new

.github/ISSUE_TEMPLATE.md
lehni: That sounds like a good idea. I've never used runkit. Would you like to prepare this standalone file?

.github/PULL_REQUEST_TEMPLATE.md
lehni: Oops yes, that's just the habit taking over : )

lehni:
Rebasing did the trick indeed. I can work more on these templates but at one point it's probably easier if you just do the fine tuning to what you think it should be, @koskimas ?

.github/ISSUE_TEMPLATE.md
koskimas: Sure, I'll write the file.

koskimas:
@lehni Yeah, thanks for the templates. I'll finetune them myself.

lehni:
@koskimas I've done some more work on these templates now, pushed as an incremental commit so you can better see what changed. I can squash them if you like.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
...
+- For features: A description of the feature.
+- For bugs: The steps to reproduce the issue.
+-->
+
+# Steps to reproduce / Link to reproduction test-case
+
+<!--
+Please link to a test-case for easier reproduction of the issue, for example a RunKit playground or a repository with a `packages.json` that installs all required dependencies and is able to execute the script.
+
+Note: Failure to provide a test-case for reproduction purposes will result in the issue being closed.
```
VladShcherbin: This one is too strict.

`If possible, please create a reproduction...` or smth like this is probably a better and a more realistic one.

lehni: I'm maintaining an open source project as well. Getting bug reports without a means to reproduce the bug can be really quite frustrating, and often leads to dangling issues that never get resolved. Most of them end up being closed unresolved... Making it sound optional will maybe not be enough of an encouragement? People can still erase these templates and write whatever they want anyway, and many of them do. I'll leave it to @koskimas to decide how determined the wording should be.

koskimas: @VladShcherbin If some kind of reproduction is not required, I end up writing them myself. Something like 95% of the bug reports end up being user errors. That is all out of my time to fix actual bugs and write new features. Do you realize that I do this on my free time for free? We need more contributors or we need to enforce some kind of reproduction. If an issue is clearly a bug, It won't get closed, but issues like "eager doesn't work!!!11! Fix iT!!11!" will be closed.

elhigu:
Did I miss it or was there place to tell which objection and knex versions are used?

.github/ISSUE_TEMPLATE.md
VladShcherbin: @koskimas I meant the last part, after `Note`. As a developer, I'd be very sad to read that if I don't add a reproduction repo than issue will be closed. This sounds very unfriendly to me.

elhigu: @VladShcherbin As a maintainer it is sad to leave open hundreds of issues that cannot be reproduced and might be real errors or not. If the issue seems critical, I'm sure that contributors investigate it anyways. Also if they are real issues they will raise from noise of closed issues, since multiple people will report the same thing over and over again.

koskimas: @VladShcherbin Yeah I think the wording could be less strict there. That seems harsh, even though it is true 😄

koskimas: Something like "Issues with a reproduction get solved much faster", or "Issues without a reproduction can get ignored for a long time until someone else finds time to reproduce them"

koskimas:
@elhigu Yep, those should be added. @lehni I can take it from here. You don't need to be our typist 😄 Thank you for the templates! I only need to change a couple of small things and they are good to go.

.github/ISSUE_TEMPLATE.md
lehni: 👍

lehni:
@koskimas you're welcome, happy to help where I can!

igor-savin-ht:
@koskimas What is left before this can be merged?

PR #605, opened 2017-11-07
Allow AjValidator to return multiple errors per property

lehni:
Ajv is capable of returning multiple validation errors per property, but `AjvValidator` currently swallows all but the last error, due to the way `parseValidationError()` is handling errors conversion.

This PR fixes that and adds a unit test for this scenario as well.

### DIFF
--- a/lib/model/AjvValidator.js
+++ b/lib/model/AjvValidator.js
@@ -141,13 +141,15 @@ function parseValidationError(errors, modelClass) {
       }
     }
 
-    errorHash[key] = [
-      {
-        message: error.message,
-        keyword: error.keyword,
-        params: error.params
-      }
-    ];
+    // More than one error can occur for the same key in Ajv, merge them in the array:
+    const array = errorHash[key] || (errorHash[key] = []);
+    // Use unshift instead of push so that the last error ends up at [0],
+    // preserving previous behavior where only the last error was stored.
+    array.unshift({
+      message: error.message,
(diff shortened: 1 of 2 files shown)


lib/model/AjvValidator.js
```diff
@@ -141,13 +141,13 @@ function parseValidationError(errors, modelClass) {
...
-        params: error.params
-      }
-    ];
+    // More than one error can occur for the same key in Ajv, merge them in the array:
+    const array = errorHash[key] || (errorHash[key] = []);
+    array.push({
+      message: error.message,
+      keyword: error.keyword,
+      params: error.params
+    });
```
koskimas: Isn't this a breaking change because:

the old version always replaced the error with the next one, causing the error to be an array with the last error inside. Now you push errors to the array an therefore the last error is the last item in the array.

I think this could be made a non-breaking change by adding errors in reverse order causing the last error to be the first just like before.

I'd assume people out there are using code like `err.data.myProp[0].message`.

koskimas:
Rebasing with master should now fix the prettier errors.

lib/model/AjvValidator.js
lehni: OK I agree, although I think it's a bit strange to get the errors in reverse order as they occur in the schema...

lehni:
@koskimas I've adjusted the sequence now as you suggested. This should be good to go.

PR #610, opened 2017-11-11
Typings: fix the return type of QueryBuilder.execute()

### LINKED ISSUE #608: Typescript: invalid type for query().execute()
First of all thanks for this awesome library, I really like it :), you guys did an amazing job!

I run into a problem with the typings in typescript.
When you run execute in this case it should return an array of objects but currently it return only a single object.

```typescript
const repositories = await RepositoryModel.query().execute();
 // error Property 'length' does not exist on type 'Repository'.
console.log(repositories.length);
```

kblomster:
`execute()` turns a query builder into a promise, but the return type of that promise in turn depends on the subtype of query builder. I removed it from `QueryBuilderBase` and added it to each of the subtypes, which should get the return type right. There might be a more clever way to do this, but I couldn't figure anything out and this does work.

Fixes #608.

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -331,6 +331,29 @@ const rowsInsertReturning: Promise<Person[]> = Person.query()
   .insert([{}])
   .returning('*');
 
+// Executing a query builder should be equivalent to treating it
+// as a promise directly, regardless of query builder return type:
+
+const maybePersonQb = Person.query().findById(1);
+let maybePersonPromise: Promise<Person | undefined> = maybePersonQb;
+maybePersonPromise = maybePersonQb.execute();
+
+const peopleQb = Person.query();
+let peoplePromise: Promise<Person[]> = peopleQb;
+peoplePromise = peopleQb.execute();
+
+const insertQb = Person.query().insert({});
+let insertPromise: Promise<Person> = insertQb;
+insertPromise = insertQb.execute();
+
+const deleteQb = Person.query().delete();
+let deletePromise: Promise<number> = deleteQb;
+deletePromise = deleteQb.execute();
+
+const pageQb = Person.query().page(1, 10);
+let pagePromise: Promise<objection.Page<Person>> = pageQb;
+pagePromise = pageQb.execute();
+
 // non-wrapped methods:
 
 const modelFromQuery: typeof objection.Model = qb.modelClass();
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -437,10 +437,14 @@ declare namespace Objection {
     throwIfNotFound(): this;
   }
 
+  export interface Executable<T> extends Promise<T> {
+    execute(): Promise<T>;
+  }
+
   /**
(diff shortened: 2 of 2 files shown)


kblomster:
Maybe I should've add a test for this, though?

mceachen:
Thanks, @kblomster !

I think we could define an interface to reduce this diff. Do you think this is better/as readable?

```ts
export interface Executable<T> extends Promise<T> {
  execute(): Promise<T>
}

export interface QueryBuilderSingle<T> extends 
  QueryBuilderBase<T>, ThrowIfNotFound, Executable<T> { }

export interface QueryBuilderUpdate<T> extends 
  QueryBuilderBase<T>, ThrowIfNotFound, Executable<number> {

}
export interface QueryBuilderInsert<T> extends 
  QueryBuilderBase<T>, ThrowIfNotFound, Executable<T[]> { }
```

In any event, any typings diff should require either a change to an existing test, or a new test, please.

kblomster:
Ah, I didn't think of that. I considered the interface solution at first but for some reason I didn't realize that the new interface could be merged with the `Promise`, so it looked kinda messy. Your idea is more DRY-ish and looks nicer. I'll change it to that.

Re: tests: I was trying to come up with some way to test for the *absence* of methods of the query builder and couldn't figure out a way to do it, but I've now realized that was a brainfart and there's no point to that. I'll get on it.

As an aside, while I was poking around I noticed TS 2.6 breaks a lot of stuff, but I'll file a separate issue for that.

mceachen:
> As an aside, while I was poking around I noticed TS 2.6 breaks a lot of stuff, but I'll file a separate issue for that.

Done--can you review #623 please?

kblomster:
@mceachen You're leaving me in the dust, here! I'm not so quick at this yet :)
I'll take a look and rebase this on top of it after you've merged it.

mceachen:
@kblomster Sorry, I thought you might be asleep/past working hours! If the rebase isn't clean because the `prettier` diffs, please let me clean up the mess. Sorry about that, I didn't think about this pending PR.

kblomster:
@mceachen Oh, no worries! It is getting late here indeed, so I'll get to this tomorrow, but it's no problem, it's like five lines of code I need to fix.

kblomster:
@mceachen I've pushed a new attempt. Could use your 👀  when you have the time.

After some waffling back and forth on what to test, I ended up simply adding `execute()` to all the query builder signature change tests. At first I was thinking I'd add a function that took a query builder and tested that it was both chainable and executable, but maybe that's overkill? What are your thoughts?

mceachen:
I think we should have both non-execute and execute calls in the examples. Can you only add .execute to, say, the 5 unique types of results, and revert the others?

Thanks!

kblomster:
I went with the first option you proposed, mostly. Let me know if you have any further comments regarding how I did it, I don't mind making cosmetic changes.

mceachen:
Thanks!

kblomster:
Thank you too for your patient and helpful reviews!

PR #623, opened 2017-11-16
Support for TypeScript 2.6

mceachen:
* Upgraded ts-related deps with `salita --update`
* Explicitly added the new strict parameters to tsconfig
* The only thing that seemed to be broken was Mixin. Updated the index.d.ts and example.ts
* Ran `prettier` over the index.d.ts (it was missing a bunch of semis and indent was incorrect). This is isolated in a commit.

@kblomster if you have time, can you take a look?

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -15,7 +15,7 @@ declare namespace Objection {
   const mixin: Mixin;
 
   interface LiteralObject {
-    [key: string]: Value
+    [key: string]: Value;
   }
 
   export interface LiteralBuilder {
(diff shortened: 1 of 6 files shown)


typings/objection/index.d.ts
```diff
@@ -106,7 +108,7 @@ declare namespace Objection {
 
   export interface RelationMapping {
     relation: Relation;
-    modelClass: typeof Model | string;
+    modelClass: ModelClass<any> | string;
```
mceachen: this and `Mixin` were the only two consequential updates here. Everything else is from `prettier`.

koskimas:
LGTM. Merge away 👍

typings/objection/index.d.ts
```diff
@@ -64,8 +64,8 @@ declare namespace Objection {
     // than an identity function type. <M extends typeof Model> retains the
     // model subclass type in the return value, without requiring the user
     // to type the Mixin call.
```
kblomster: This comment is obsolete and no longer matches the code below. Consider removing?

kblomster:
You already merged but I didn't really have anything to add other than the comment anyway. :)

typings/objection/index.d.ts
mceachen: will do

Issue #558, opened 2017-10-16
PLEASE VOTE: Should we change object references from JSON methods in the TypeScript typings?


mceachen:
Currently all references to JSON objects in the typings are `objects`, like

```ts
    fromJson(json: object, opt?: ModelOptions): M;
    fromDatabaseJson(row: object): M;
    loadRelated(models: (Model | object)[],
    ...
```

TypeScript's `object` has no methods, so if you override, say, `$formatDatabaseJson`, and your overriding method uses `object`, you have to use bracket-notation to get member values (like `obj["id"]` rather than `obj.id`.

We could use the `any` type, but then strings, numbers, functions, and symbols would be legal arguments to these methods, which clearly isn't correct. We *could*, however, introduce a new `JsonObject`:

```ts
  type JsonValue = string | number | boolean | Date | null | undefined | JsonObject 

  interface JsonObject {
    [key: string]: JsonValue | JsonValue[]
  }
```

Please thumb-up if you'd like to see this change, or thumb-down if you think the typings are fine as they are. Feel free to comment, as well.

newhouse:
I *think* I understand the problem you're highlighting, and the pros/cons of your solution...but I leave it up to you TS experts, of which I am definitely not one.

jordansexton:
Given that `undefined` and `Date` are not valid JSON values, is the idea of a JSON value here an ad-hoc convention to follow, but not necessarily JSON standard?

I'm also wondering what would happen if parsing values into something novel with https://github.com/brianc/node-pg-types. Would a looser object type achieve the desired effect of being able to use property dot notation? e.g.

```typescript
interface PlainObject {
  [key: string]: any
}
```

mceachen:
@jordansexton I'm using "JSON" in reference to the methods with that substring--you're correct, a naked Date wouldn't make sense in a JSON string, but the objects you pass into `.fromJSON()` will absolutely support Date or other value types.

I'm finding it a bit ungainly to have to cast the result from, say, `$toJSON()`, before I can do normal dotted field access (like `foo.bar` instead of `foo["bar"]`. Another alternative from the above `JsonObject` would be to simply use `any`.

kblomster:
[...]

PR #624, opened 2017-11-17
Typings updates

koskimas:
I added new methods and fixed some errors along the way.

Biggest change is using a new `AnyObject` interface in model hooks. Previously they used `object` which made overriding them really difficult since you cannot access any property through `object`. If there is a better way to achieve the same, feel free to point it out.

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -234,6 +234,8 @@ qb = qb.where(Person.raw('random()', 1, '2', raw('3')));
 const rowInserted: Promise<Person> = qb.insert({firstName: 'bob'});
 const rowsInserted: Promise<Person[]> = qb.insert([{firstName: 'alice'}, {firstName: 'bob'}]);
 const rowsInsertedWithRelated: Promise<Person> = qb.insertWithRelated({});
+const rowsInsertGraph1: Promise<Person> = qb.insertGraph({});
+const rowsInsertGraph2: Promise<Person> = qb.insertGraph({}, {relate: true});
 const rowsUpdated: Promise<number> = qb.update({});
 const rowsPatched: Promise<number> = qb.patch({});
 const rowsDeleted: Promise<number> = qb.delete();
@@ -241,10 +243,16 @@ const rowsDeletedById: Promise<number> = qb.deleteById(123);
 const rowsDeletedByIds: Promise<number> = qb.deleteById([123, 456]);
 
 const insertedModel: Promise<Person> = Person.query().insertAndFetch({});
-const insertedModels: Promise<Person[]> = Person.query().insertGraphAndFetch([
+const insertedModels1: Promise<Person[]> = Person.query().insertGraphAndFetch([
   new Person(),
   new Person()
 ]);
+const insertedModels2: Promise<Person[]> = Person.query().insertGraphAndFetch([
+  new Person(),
+  new Person()
+], {
+  relate: true
+});
 
 const upsertModel1: Promise<Person> = Person.query().upsertGraph({});
 const upsertModel2: Promise<Person> = Person.query().upsertGraph({}, {relate: true});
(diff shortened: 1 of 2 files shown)


typings/objection/index.d.ts
```diff
@@ -14,10 +14,17 @@ declare namespace Objection {
...
 
+  const snakeCaseMappers: () => ColumnNameMappers;
+  const knexSnakeCaseMappers: () => KnexMappers;
+
   interface LiteralObject {
     [key: string]: Value;
   }
 
+  export interface AnyObject {
+    [key: string]: any;
```
mceachen: Should the value here really be any? Like, is it OK to have an array of model objects (not just object representations of model objects, but actual instances of Model subclasses)? It should nest other instances, correct?

See https://github.com/Vincit/objection.js/issues/558#issue-265887726 for a more constrained "AnyObject" definition, if we need that.

I think we can do better with the name, too, given that "object" already means two things in [TypeScript](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#object-type) (with `object` and `Object`) and `any` is a superset of those, so it's compounding two names with three definitions. 

Is `SimpleObject` better?

```ts
  type SimpleValue = string | number | boolean | Date | null | undefined | SimpleObject 

  interface SimpleObject {
    [key: string]: SimpleValue | SimpleValue[]
  }
```

typings/objection/index.d.ts
```diff
@@ -622,9 +651,10 @@ declare namespace Objection {
     isExecutable(): boolean;
     isFindQuery(): boolean;
 
-    runBefore(fn: (builder: this) => void): this;
+    runBefore(fn: (result: any, builder: this) => any): this;
```
mceachen: Is `any` the best we can do here for these?

koskimas: For runBefore the value can literally be anything. runAfter gets passed the query result, so we could improve that, but we would need to move the method to each of the different QueryBuilder classes. Return value can be anything.

typings/objection/index.d.ts
koskimas: Usually the objects are JSON objects, but they can be basically anything. Database drivers can have their own conversions and users can pass anything to `fromJson` etc. Also I don't really see the use of limiting that type. Can you give me an example where that would be useful?

tests/ts/examples.ts
```diff
@@ -24,6 +27,16 @@ class Person extends objection.Model {
   async $beforeInsert(queryContext: objection.QueryContext) {
     console.log(queryContext.someCustomValue);
   }
+
+  $formatDatabaseJson(json: objection.AnyObject) {
+    json.bar = json.foo;
```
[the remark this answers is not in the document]
koskimas: I'll do that

tests/ts/examples.ts
```diff
@@ -13,6 +13,9 @@ const {lit, raw, ref} = objection;
 class Person extends objection.Model {
   firstName: string;
   lastName: string;
+
+  static columnNameMappers = objection.snakeCaseMappers();
```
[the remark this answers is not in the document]
koskimas: I'll add a "test" that uses the return value.

typings/objection/index.d.ts
koskimas: You are right about the name, but I don't think SimpleObject is better. The object may be very complex.

koskimas: What about `Pojo`?

koskimas: or `PlainObject`?

mceachen: "Pojo" is perfect.

typings/objection/index.d.ts
mceachen: ok!

koskimas: I fixed the `runAfter` `result` type.

koskimas:
@mceachen Is there still something you'd like to change or can I merge this?

mceachen:
Looks good to me!

PR #633, opened 2017-11-23
errors: More informative error message if wrong table is specified

igor-savin-ht:
for relation.

### DIFF
--- a/lib/relations/RelationProperty.js
+++ b/lib/relations/RelationProperty.js
@@ -10,7 +10,12 @@ const createRaw = require('../queryBuilder/RawBuilder').raw;
 const propToStr = require('../model/modelValues').propToStr;
 const PROP_KEY_PREFIX = require('../model/modelValues').PROP_KEY_PREFIX;
 
-class ModelNotFoundError extends Error {}
+class ModelNotFoundError extends Error {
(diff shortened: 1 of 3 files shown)


lib/relations/RelationProperty.js
```diff
@@ -182,7 +182,9 @@ function createPaths(refs, modelClassResolver) {
     const modelClass = modelClassResolver(ref.tableName);
 
     if (!modelClass) {
-      throw new ModelNotFoundError();
+      const error = new ModelNotFoundError();
+      error.tableName = ref.tableName;
```
koskimas: This property should be added to the class and passed to the constructor:

```js
class ModelNotFoundError extends Error {
  constructor(tableName) {
    super();
    this.tableName = tableName;
  }
}
```

koskimas:
Thanks for the fix! I had one small suggestion. If you fix that, I'll merge this.

lib/relations/RelationProperty.js
fl0w: May I quickly ask why/the reason for this?

igor-savin-ht:
@koskimas Done, thank you!

koskimas:
Awesome! Thank you!

Issue #19, opened 2015-07-12
Support for polymorphic associations?


juanjoLenero:
Are you planning to support polymorphic associations?

[Here's](https://gist.github.com/p-baleine/7703145) an example of it in another ORM. 
Another [one](http://docs.sequelizejs.com/en/latest/docs/associations/#1m).

I haven't delved into moron's source code but perhaps an API like this would work:

``` javascript
// Schema for a comment model which relates polymorphically 
// to several tables through the commentable and commentableId attributes.
Comment.jsonSchema = {
    type: 'object',

    properties: {
        id: {type: 'integer'},
        text: {type: 'string'},
        commentable: {type: 'string'},
        commentableId: {type: 'integer'},
        authorId: {type: 'integer'},
        createdAt: {type: 'string', format: 'date-time'},
        updatedAt: {type: 'string', format: 'date-time'}
    }
};

Comment.relationMappings = {
    // A comment can be related to a topic.
    topic: {
        relation: Model.OneToOneRelation,
        modelClass: __dirname + '/topic',
        join: {
            from: 'Comment.commentableId',
            to: ['Comment.commentable', 'id']
        }

    // A comment can also be related to a user's profile.
    profile: {
        relation: Model.OneToOneRelation,
        modelClass: __dirname + '/profile',
        join: {
            from: 'Comment.commentableId',
            to: ['Comment.commentable', 'id']
        }
    }
};

```

Or maybe something like this:

``` javascript
    topic: {
        ...
        join: {
           from: 'Comment.commentableId',
           morph: {table: 'Comment.commentable', attribute: 'id'},
        }
```

koskimas:
I will definitely add some way to support polymorphic associations at some point. I'm just about to add a feature that allows you add a filter to relations like this:

```js
Topic.relationMappings = {
  comments: {
    relation: Model.OneToManyRelation,
    modelClass: Comment,
    // Extra filter query for the relation. This can be a hash of `{column: 'value'}` pairs
    // or a function that takes a query builder as a parameter.
    filter: {commentable: 'Topic'},
    join: {
      from: 'Topic.id',
      to: 'Comment.commentableId'
    }
  }
};

Profile.relationMappings = {
  comments: {
    relation: Model.OneToManyRelation,
    modelClass: Comment,
    filter: {commentable: 'Profile'},
    join: {
      from: 'Profile.id',
      to: 'Comment.commentableId'
    }
  }
};
```

[...]

PR #657, opened 2017-12-12
Model class thunks

mceachen:
I'm webpacking my server-side code (to both reduce the number of files I'm installing and to support uglification), and path references to model classes don't work in that environment.

I can't give a direct reference to my Model subclasses in my `relationMappings` because that causes circular dependencies.

I'm adding thunk-dereferenced Model subclasses in this PR, so Objection doesn't have to do `require`s in a environment where webpack has concatenated all the files, or requiring all the model class modules to be explicitly `exclude`d from webpack (and then hope to figure out path resolution in different environments.

### DIFF
--- a/doc/includes/API.md
+++ b/doc/includes/API.md
@@ -4907,7 +4907,7 @@ class Person extends Model {
   static relationMappings = {
     pets: {
       relation: Model.HasManyRelation,
-      modelClass: Animal,
+      modelClass: () => Animal,
       join: {
         from: 'Person.id',
         // Any of the `to` and `from` fields can also be
@@ -4921,7 +4921,7 @@ class Person extends Model {
 
     father: {
       relation: Model.BelongsToOneRelation,
-      modelClass: Person,
+      modelClass: () => Person,
       join: {
         from: 'Person.fatherId',
         to: 'Person.id'
@@ -4930,7 +4930,7 @@ class Person extends Model {
 
     movies: {
       relation: Model.ManyToManyRelation,
-      modelClass: Movie,
+      modelClass: () => Movie,
       join: {
         from: 'Person.id',
         through: {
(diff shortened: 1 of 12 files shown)


examples/express-ts/package.json
```diff
@@ -21,7 +21,7 @@
     "knex": "^0.14.2",
     "lodash": "^4.17.4",
     "morgan": "^1.9.0",
-    "objection": "^0.9.2",
+    "objection": "../..",
```
mceachen: I needed to restore the `../..` to pull in the typings changes in this same PR. I believe clean rebuilds of node_modules would have fixed the prior issue, as my scratch build is fine.

mceachen:
This is ready for review now, @koskimas.

jordansexton:
Could this be expanded on to allow something like https://github.com/Vincit/objection.js/issues/19#issuecomment-350504103?

mceachen:
> Could this be expanded on to allow something like #19 (comment)?

I don't believe so, given that you need a discriminator column specific to a given instance, and the Relation will be set up at most once per query.

(Right?)

doc/includes/API.md
```diff
@@ -4968,10 +4968,16 @@ case of ManyToManyRelation also the join table needs to be defined. This is done
...
-2. An absolute path to a module that exports a model class
-3. A path relative to one of the paths in [`modelPaths`](#modelpaths) array.
+1. A thunk (a function with no arguments) that returns a model class constructor
+1. An absolute path to a module that exports a model class
+1. A path relative to one of the paths in [`modelPaths`](#modelpaths) array.
+
+The thunk and file path versions avoid require loops.
 
-The file path versions are handy for avoiding require loops.
+Thunks are recommended if you are packing up your server-side code, as the
```
mceachen: After successfully using thunks in my own project, I'd suggest we make this the default recommendation. `() => Movie` is straightforward, doesn't cause `webpack` or similar tools heartache due to synthetic `require`s, and avoids circular require loops (whose errors are always confusing). I'll let you decide how much you want to push this, of course.

koskimas:
You can already give a thunk for `relationMappings`. What additional benefit does this bring to the table?

```js
static relationMappings = () => ({
  movies: {
    ...
  },

  pets: {
    ..
  }
});
```

doc/includes/API.md
fl0w: Just to add to this:
I have a habit of memoizing at the top of file, e.g.

```js
const { memoize } = require('lodash')
const MyModel = memoize(() => require('./relative/path/to'))

// ...
await Model().query()
```
Sort of helps normalise model dependency throughout the application, and also help with re/up-start time as projects grows. Memoizing was more relevant pre node 8 (or so) as `require` wasn't really optimised at all. Habits are tough to shake.

mceachen:
@koskimas wrote:
> You can already give a thunk for relationMappings. 

Ah, that's great! I didn't realize that (and the typings don't reflect this). I'll drop this PR and fix this typings now.

PR #661, opened 2017-12-13
Add TypeScript support to thunked relationMappings

mceachen:
* Thanks to https://github.com/Vincit/objection.js/pull/657#issuecomment-351302085, it turns out relationMappings can be a thunk to a RelationMappings. Added the thunk as a valid value for the relationMappings field.

* added thunkified relationMappings examples to express-ts and examples.ts to verify the typings and implementation work as expected.

* prettier now reformats *.js and *.ts files. Ran against tree, so some of these diffs are just whitespace due to prettier.

* removed tslint, as it was predominantly making whitespace assertions which don't agree with the prettier config.


koskimas:
Awesome, thank you!

Issue #664, opened 2017-12-14
Should the typings include AjvValidator?


jtlapp:
I'm using Typescript with the AjvValidator in a custom query builder. I noticed that the type definitions don't include `Objection.AjvValidator`. Is that intentional? If so, it's easy enough to add it locally.

I realize custom query builders are currently broken with Typescript (#319), but I'm hoping to offer a fix for that. (I'm porting working JS code.)

jtlapp:
After resolving this, I believe that the type does belong in `objection.js`. The signature is specific to this module, requiring people to reverse engineer the code to get it right.

Here's a fix. I'll shortly be working on a fix for #319 and can send a PR for everything at once.

```ts
import * as ajv from 'ajv';

declare namespace Objection {
  /*...*/
  export type AjvConfig = {
    onCreateAjv(ajv: ajv.Ajv): void;
    options?: ajv.Options;
  };
  const AjvValidator: new(config: AjvConfig) => ajv.Ajv;
  /*...*/
}
```

jtlapp:
I've also got a fix for adding the `createValidationError()` tyepdef. Looks like I'll be collecting mods until I've got everything ported.

koskimas:
AjvValidator is a class, not a function that returns an instance of `Ajv`.

koskimas:
To answer the original question, yes the typings should include `AjvValidator`. My previous comment was about your suggested typings.

jtlapp:
Got it. I see what the problem is. The typedef for `Model.createValidator()` is missing too, so Typescript inferred its return value. My code never calls the created object, so everything works.

Okay, looks like I've got more to add thanks.

BTW, the `new` keyword in a typedef defines a constructor rather than a function per se, so the above code effectively includes the Ajv class without redefining it. (Might just be your word choice I'm responding to, in which case I apologize.)

koskimas:
@jtlapp A constructor that returns the wrong type though. `new AjvValidator()` returns an instance of `AjvValidator` which is a wrapper around `ajv.Avj`. `AjvValidator` doesn't inherit `ajv.Ajv`.

jtlapp:
Yeah, got all that. Sorry, I quit when the first thing I threw in satisfied Typescript, not realizing that satisfying Typescript didn't mean the types were right yet.

Okay, I'm going to PR new typings for just the validation stuff. Most of it was missing, but a bit of what was there was also wrong. Plus I may have found a bug in AjvValidator. Will explain below...

[...]

PR #666, opened 2017-12-18
validator typings, for discussion

jtlapp:
Validator typings for #664. Will discuss issues discovered there.

Note: This PR includes a TBD for the type of `Validator.args` passed to the Validator methods.

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -316,7 +316,7 @@ declare namespace Objection {
 
     bindKnex(knex: knex): this;
     bindTransaction(transaction: Transaction): this;
-    createValidator(): AjvValidator;
+    createValidator(): Validator;
     createValidationError(errorHash: ErrorHash): ValidationError;
 
     fromJson(json: object, opt?: ModelOptions): M;
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -405,7 +405,7 @@ declare namespace Objection {
     $id(): any;
     $id(id: any): void;
 
-    $beforeValidate(jsonSchema: JsonSchema, json: Pojo, opt: ModelOptions): void;
+    $beforeValidate(jsonSchema: JsonSchema, json: Pojo, opt: ModelOptions): JsonSchema;
     $validate(json: Pojo, opt: ModelOptions): Pojo; // may throw ValidationError if validation fails
     $afterValidate(json: Pojo, opt: ModelOptions): void; // may throw ValidationError if validation fails
 
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -4,6 +4,7 @@
 
 /// <reference types="node" />
 import * as knex from 'knex';
+import * as ajv from 'ajv';
 
 export = Objection;
 
@@ -95,11 +96,37 @@ declare namespace Objection {
     skipValidation?: boolean;
   }
 
+  export type ValidatorArgs = {
+    ctx: object, // TBD: always {}; perhaps should be QueryContext (class QueryBuilderUserContext)?
+    model: Model,
+    json: Pojo,
+    options: ModelOptions
+  }
+
+  export class Validator {
+    beforeValidate(args: ValidatorArgs): void;
+    validate(args: ValidatorArgs): Pojo;
+    afterValidate(args: ValidatorArgs): void;
+  }
+
+  export type AjvConfig = {
+    onCreateAjv(ajv: ajv.Ajv): void;
+    options?: ajv.Options;
+  };
+
+  export class AjvValidator extends Validator {
+    constructor(config: AjvConfig);
+  }
+
   export class ValidationError extends Error {
     constructor(errors: any);
     statusCode: number;
     data: any;
     message: string;
+  }
+  
+  export interface ErrorHash {
+    [columnName: string]: ValidationError[];
   }
 
   export interface RelationMappings {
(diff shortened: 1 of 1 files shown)


jtlapp:
The docs gave me my answer, so the validation typings are now complete, but I'll note other changes in the relevant issue.

"A context object shared between the validation methods. A new object is created for each validation operation. You can store any data here."

typings/objection/index.d.ts
```diff
@@ -349,6 +378,8 @@ declare namespace Objection {
     static knexQuery(): knex.QueryBuilder;
     static bindKnex<T>(this: T, knex: knex): T;
     static bindTransaction<T>(this: T, transaction: Transaction): T;
+    static createValidator(): AjvValidator;
```
jtlapp: Corrected to Validator in subsequent push.

koskimas:
@jtlapp Thank you so much! Let's wait for @mceachen to comment before merging.

jtlapp:
Sure thing! @mceachen, be sure to see my [notes posted for the issue](https://github.com/Vincit/objection.js/issues/664#issuecomment-352311641).

Thanks for the thanks!

typings/objection/index.d.ts
```diff
@@ -374,8 +405,8 @@ declare namespace Objection {
     $id(): any;
     $id(id: any): void;
 
-    $beforeValidate(jsonSchema: JsonSchema, json: Pojo, opt: ModelOptions): JsonSchema;
-    $validate(): void; // may throw ValidationError if validation fails
+    $beforeValidate(jsonSchema: JsonSchema, json: Pojo, opt: ModelOptions): void;
```
koskimas: This methods mustreturn `JsonSchema`. See the line where it is used.

typings/objection/index.d.ts
```diff
@@ -95,12 +96,38 @@ declare namespace Objection {
     skipValidation?: boolean;
   }
 
+  export type ValidatorArgs = {
+    ctx: object,
```
koskimas: Maybe a `Pojo` is a better type since it is ugly to access anything through an `object` type. The point of `ctx` is that implementors can store any data there.

jtlapp: I just spent several hour investigating this, including having to isolate a [bug that crashed TSC](https://github.com/Microsoft/TypeScript/issues/20886), and I've made an interesting discovery.

I think you're right that the `object` type is useless. However, if you instead do this:

```ts
  export interface ValidatorContext { }

  export type ValidatorArgs = {
    ctx: ValidatorContext,
    model: Model,
    json: Pojo,
    options: ModelOptions
  }

  export class Validator {
    beforeValidate(args: ValidatorArgs): void;
    validate(args: ValidatorArgs): Pojo;
    afterValidate(args: ValidatorArgs): void;
  }
```

**AND** if you remove the `declare namespace Objection` and `export = Objection` from the typings, then you can do this in your app to get static type checking on the context:

```ts
declare module 'objection' {
    export interface ValidatorContext {
        foo: number;
    }
}
```

Why is Objection declared as a namespace anyway? I wonder if that's what's making it impossible to open Objection types for declaration merging. @mceachen?

Removing the outer namespace would allow apps to open QueryContext for extension to get static type checking.

jtlapp: Woo! Another interesting discovery. We can have our cake and eat it too. If I declare:

```ts
export interface ValidatorContext {
    [key: string]: any;
}
```

Then I can assign `ctx.foo` to whatever I want, no problem.

But if the app opens that same Pojo ValidatorContext class as follows:

```ts
declare module 'objection' {
    export interface ValidatorContext {
        foo: number;
    }
}
```

Now suddenly `foo` has to be a number, but all other values remain unconstrained.

This would allow apps to decide whether to enforce static type checking, though it does require that the dev think ahead enough to add the decls necessary to get the checking.

So it appears that the only real problem is opening up the Objection namespace.

jtlapp: The above `declare module 'objection'` decl does indeed open up the namespace! I thought I had tried that. Okay, so I've updated this PR with the following:

```ts
  export interface ValidatorContext extends Pojo { }

  export type ValidatorArgs = {
    ctx: ValidatorContext,
    model: Model,
    json: Pojo,
    options: ModelOptions
  }
```

I'd personally prefer to **have to** extend ValidatorContext in an app to be able to add app-specific properties, but in keeping with the rest of the typings, I went ahead and extended Pojo. That way at least it's possible for an app to restrict some properties.

So there's no need to remove the Objection namespace wrapper after all.

mceachen: Sorry for the delay. 

In reference to `type` vs `interface`: In general, `interface`s should be what you use for rather than a `type` alias. The [type alias docs](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases) are fairly clear. (Note that the TypeScript 2.3 series would actually crash when compiling older versions of objection, because I hadn't internalized the fact that types were uncached by the compiler, and caused memory issues. The solution was to convert most type aliases to interfaces).

I'm happy to change any type alias in the typings to an interface.

jtlapp: I should have thought to do that. I find myself frequently turning types into interfaces.

How do you feel about ValidatorContext extending Pojo? I think that's implementation inheritance -- ValidatorContext isn't always and forever a thing of type Pojo. I should nix that and put `[key: string]: any` back in, as you've done elsewhere.

jtlapp:
I just discovered the [typings tests](https://github.com/Vincit/objection.js/blob/master/tests/ts/examples.ts) and should add some for the new Validator typings.

Issue #659, opened 2017-12-13
Return single item after deleting single item


VladShcherbin:
Same as https://github.com/Vincit/objection.js/issues/423, but for `delete`:

```js
// expected
const user = await User.query().where('id', id).first()
const deletedUser = await user.$query().delete().returning('*')

// current
const deletedUser = await user.$query().delete().first().returning('*')
```

Currently, this returns an array, even if I use `first` in the first query. It would be great to return a single item when `first` was used.

koskimas:
I'd merge a PR for this.

VladShcherbin:
@koskimas sure, I can give it a shot next week.

PR #679, opened 2017-12-29
Return object on single item delete

VladShcherbin:
Closes https://github.com/Vincit/objection.js/issues/659


koskimas:
Thanks! This is a breaking change, so I'll merge this later.

PR #685, opened 2017-12-31
Add simple string matching to forEachOperation()

lehni:
I thought it would be nice to be able to use `builder.findLastOperation('range')`.

This simple change adds that, and makes the current code a bit more compact also.

### DIFF
--- a/lib/queryBuilder/QueryBuilderOperationSupport.js
+++ b/lib/queryBuilder/QueryBuilderOperationSupport.js
@@ -152,10 +152,23 @@ class QueryBuilderOperationSupport {
   forEachOperation(operationSelector, callback, match) {
     match = match == null ? true : match;
 
+    let check;
     if (operationSelector instanceof RegExp) {
-      forEachOperationRegex(this._operations, operationSelector, callback, match);
+      check = op => operationSelector.test(op.name);
+    } else if (typeof operationSelector === 'string') {
+      check = op => op.name === operationSelector;
     } else {
-      forEachOperationInstanceOf(this._operations, operationSelector, callback, match);
+      check = op => op.is(operationSelector);
+    }
+
+    for (let i = 0, l = this._operations.length; i < l; ++i) {
+      const op = this._operations[i];
+
+      if (check(op) === match) {
+        if (callback(op, i) === false) {
+          break;
+        }
+      }
     }
 
     return this;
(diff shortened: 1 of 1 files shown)


koskimas:
I'll review this later. The old code was not compact for performance reasons, but we can probably start dropping these optimizations since they were written mostly for node < 5 and V8 has gotten alot better since then. For the same reason I use for loops instead of `forEach` etc. too much in the code. It used to be that anonymous function were pretty costly. I have prematurely optimized a bunch of places and this is most probably one of those.

lehni:
@koskimas yeah I was supposing something like that.
An additional, related thought for improving performance: Wouldn't it be better to check against the name instead of using `instanceof`checks where the same name is used? With my proposed change here, I believe `builder.has(FindOperation)` could become `builder.has('find')`, and this should match faster.

koskimas:
Checking for name is not the same thing when inheritance is involved.

lehni:
I am aware. But I think there is no inheritance with `FindOperation`? It could be exposed as `FindSelector`, like it's already being done with other selectors?

lehni:
Scratch that, there is `InstanceFindOperation`, but it also uses `'find'` as the name. It was just a suggestion for a potential speed up.

koskimas:
Why can't you use a regex that is already supported: `/find/`?

lehni:
I can of course, but I thought that this would be slower and I was thinking that simple string matching could be a nice addition, especially if the rewrite is accepted, as it would only two lines of code. It's just a suggestion.

koskimas:
I don't really see a point in this. There is already a way to do this. No need to add another. Regex literals like `/find/` are insanely fast while closure creations are not. This PR would only make things slower if anything.

lehni:
I'm sorry to disagree, but I think that claim is not quite right. I've done a quick jsperf test to test my hunch, and the difference is rather big, and regexps appear to be much slower, while there is barely a difference between the closure and a direct string match:

https://jsperf.com/regexps-vs-string-matching-vs-closures/1

screen shot 2018-01-02 at 22 19 13

lehni:
I've tried `/^three$/` also to make sure it behaves the same way as the other tests, but the result is pretty much the same. Sometimes, closures are even faster than the direct check. I think modern JS engines do an amazing job at inlining them rather than calling functions, and I think the code in this PR would have been such a case.

screen shot 2018-01-02 at 22 24 49

koskimas:
Your closure isn't a closure at all. It doesn't depend on any dynamic arguments and can therefore easily be inlined. The result could be different if you take `three` as an argument so that the compiler cannot know that the value is.

Another thing is that you don't use the `count` at all and the optimizer can probably eliminate all the code from the two last cases. If you do something with the `count` value so that the compiler cannot eliminate it, you can see different results.

koskimas:
According to this https://jsperf.com/closure-regex-loop-crap/1 regexes are still slower, but the difference is not as significant.

koskimas:
What the hell... I suddenly started getting completely different results from the same test. Now the tests says regexes are as slow as in your test, but earlier I got results that regexes were only 50% slower. Go figure.

Once again this is a proof that microbenchmarks are useless. Even with the regex version we are getting 10 million calls per second, so that's never going to be a problem.

lehni:
Yeah I understand, your explanations make a lot of sense. When trying to make sense of what I observed, I was reading on slow / fast paths in regexps in recent V8 optimizations, perhaps that explains what you were observing:

https://medium.com/dailyjs/js-regexp-fast-and-slow-d29d6b77b06

lehni:
Your test results on my machine. Regexps do look quite a bit slower here:
perf

PR #686, opened 2017-12-31
Add property data paths to nested insert and upsert validations

lehni:
When validation errors happen in graph inserts and upserts of nested relations data, Objection 
currently doesn't specific the location of the error.

This PR calculated dataPaths for graph inserts and upserts, and then provides these as relative JSON pointers instead of plain property names in the validation errors. If the data is only on the root level of the graph, then these two behaviors are actually the same.

As discussed on [Gitter](https://gitter.im/Vincit/objection.js?at=5a48d9f368d092bb620aace4), the convention of how to handle JSON pointers, and wether they should be absolute or relative paths to the data still needs some more thought and a proper decision.

For now, here the current state of things, so the discussion can be continued on in the issue here.

### DIFF
--- a/lib/queryBuilder/graphInserter/DependencyGraph.js
+++ b/lib/queryBuilder/graphInserter/DependencyGraph.js
@@ -28,10 +28,10 @@ class DependencyGraph {
 
     if (Array.isArray(models)) {
       for (let i = 0, l = models.length; i < l; ++i) {
-        this.buildForModel(modelClass, models[i], null, null, this.allowedRelations);
+        this.buildForModel(modelClass, models[i], null, null, this.allowedRelations, []);
       }
     } else {
-      this.buildForModel(modelClass, models, null, null, this.allowedRelations);
+      this.buildForModel(modelClass, models, null, null, this.allowedRelations, []);
     }
 
     this.solveReferences();
@@ -47,7 +47,7 @@ class DependencyGraph {
     return this.nodes;
   }
 
-  buildForModel(modelClass, model, parentNode, rel, allowedRelations) {
+  buildForModel(modelClass, model, parentNode, rel, allowedRelations, dataPath) {
     if (!model || !model.$isObjectionModel) {
       throw modelClass.createValidationError({
         type: ValidationErrorType.InvalidGraph,
@@ -59,7 +59,7 @@ class DependencyGraph {
       model[modelClass.uidProp] = this.createUid();
     }
 
-    const node = new DependencyNode(parentNode, model, modelClass, rel);
+    const node = new DependencyNode(parentNode, model, modelClass, rel, dataPath);
     const isRelate = this.isRelate(modelClass, model, parentNode, rel);
     const dbRef = model[modelClass.dbRefProp];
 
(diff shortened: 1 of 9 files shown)


lib/queryBuilder/graphInserter/DependencyGraph.js
```diff
@@ -126,12 +126,14 @@ class DependencyGraph {
...
-        for (let i = 0, l = relModels.length; i < l; ++i) {
-          this.buildForModel(rel.relatedModelClass, relModels[i], node, rel, nextAllowed);
+      dataPath = `${dataPath}/${rel.name}`;
+      const relModelClass = rel.relatedModelClass;
+      if (Array.isArray(relData)) {
+        for (let i = 0, l = relData.length; i < l; ++i) {
+          this.buildForModel(relModelClass, relData[i], node, rel, nextAllowed, `${dataPath}/${i}`);
         }
-      } else if (relModels) {
-        this.buildForModel(rel.relatedModelClass, relModels, node, rel, nextAllowed);
```
lehni: @koskimas I've renamed `relModels` to the shorter `relData` to avoid Prettier moving every argument to a new line here... It's also the reason why I'm putting `rel.relatedModelClass` into its own variable.

lehni:
@koskimas I've now updated this PR against the `1.0.0` branch as base.

lib/queryBuilder/graphUpserter/UpsertNode.js
```diff
@@ -31,6 +31,16 @@ class UpsertNode {
...
+    if (parentNode) {
+      const { name } = relExpr;
+      const { relations } = parentNode;
+      const nodes = (relations[name] = relations[name] || []);
+      this.index = nodes.push(this) - 1;
+    } else {
+      this.index = null;
+    }
+    this.dataPath = getDataPath(this);
+
```
lehni: @koskimas I don't understand why, but this part needed to be moved below the call of `getTypes()` for the tests not to break... Took me a while to figure it out.

lib/queryBuilder/graphInserter/DependencyGraph.js
koskimas: There is no need to fight `prettier`. The whole point of using `prettier` is to accept its style choices and just go with them.

koskimas: No need to change this back though.. I just meant in the future, no need to do stuff like this

lib/queryBuilder/graphUpserter/UpsertNode.js
koskimas: Actually I think this may be wrong. There can be multiple `UpsertNode`s for one relation. The index may not be correct.

lehni: Oh damn... Is there a way to fix that in a way that the index points to the right place in the graph always?

lehni:
@koskimas f11bb14682253a89a3df265e3bf56e874bc2578f addresses the issue with the `dataPath` indices in upserts, at least partially. I will add a comment in the code itself for a case that I don't know how to handle.

lib/queryBuilder/graphUpserter/UpsertGraph.js
```diff
@@ -66,64 +66,74 @@ class UpsertGraph {
...
     });
 
     current.forEach(current => {
       const key = current.$propKey(idProp);
       const upsert = upsertById[key];
 
       if (!upsert) {
-        this.buildGraphSingle(modelClass, upsert, current, parentNode, relExpr);
+        // TODO: What to do about dataPath here?
+        this.buildGraphSingle(modelClass, upsert, current, parentNode, relExpr, null);
```
lehni: @koskimas I am now building the `dataPath` recursively, in the same way it is done for inserts. the problem here is that for a `current` model that has no corresponding `upsert` model, there isn't a correct index for it into the graph, since it doesn't actually show up in the graph. This shouldn't cause any validation errors, but what if it does? What kind of index should we give it? `-1`?

lehni:
Also, this is currently failing because of the changed `errors` keys... the failing tests need some more substantial rewriting though because they match multiple errors that now have different properties.

lehni:
I just fixed the existing tests in e45fcada1b5ff28eb24598e87659214706544199. This still needs new tests though.

lehni:
@koskimas also, I'm currently pushing separate commits to this PR so we can keep changing things without losing the history. Once we settle on an approach, I'll squash them all to one commit before you  merge.

lib/queryBuilder/graphUpserter/UpsertGraph.js
koskimas: These nodes create delete and unrelate operations and nothing is validated. I think -1 is fine here. It should never actually get used.

lehni: Cool, I'll adjust that then.

lehni:
Alright, now all that's left to do is write unit tests for it... : )

lib/queryBuilder/graphInserter/DependencyGraph.js
```diff
@@ -27,10 +27,10 @@ class DependencyGraph {
 
     if (Array.isArray(models)) {
       for (let i = 0, l = models.length; i < l; ++i) {
-        this.buildForModel(modelClass, models[i], null, null, this.allowedRelations);
+        this.buildForModel(modelClass, models[i], null, null, this.allowedRelations, '');
```
koskimas: Empty string should be an empty array instead right?

lehni: Yes! That's an oversight, with no consequences, because [...emptyString] = [] : )

lib/queryBuilder/graphInserter/DependencyGraph.js
lehni: Oh I see. I can change it if you like, no problem

lehni:
@koskimas that's a random fail in Node 7, it should all pass

lehni:
Weird, this is the error... Why is the sequence differently there? And would that be an error?

screen shot 2018-01-09 at 01 01 09

koskimas:
Yeah, that's a random fail. I'll rerun the tests.

koskimas:
Are you done with this? Some test to `insertGraph` test suite would be nice. Then I think I'm happy with this.

lehni:
Yes I'm done with it except for the missing tests. I shall get on it ASAP, but this week is fairly busy. Maybe I can whip something up tonight.

lehni:
The `ValidationError` format is not going to change in other ways then, yes?

lehni:
@koskimas do these work for you? 060478f790f4eceb7d30dee8f41a7f1ca8c4bba7

koskimas:
Good tests, thanks! If you resolve the conflicts, I'll merge this.

lehni:
@koskimas I've merged the conflicts and pushed a new squashed commit. Let's see how the tests will do on this one.

lehni:
Another random fail on Node v9... All others pass, I think it'll turn green once you rerun it.

koskimas:
Yeah, it must be a random fail since you'd need to be a wizard to only break node 9 😄 

Thank you so much for working with this!

lehni:
You're welcome! More than happy to help out.

PR #687, opened 2017-12-31
Handle undefined returns from $beforeValidate()

lehni:
This simple PR detects the return of `undefined` from `$beforeValidate()` and uses the default `jsonSchema` in this case. The handler can return `null` if it explicitly wants to disable the use of a schema.

### DIFF
--- a/lib/model/AjvValidator.js
+++ b/lib/model/AjvValidator.js
@@ -48,7 +48,10 @@ class AjvValidator extends Validator {
     // rare case that the given model has implemented the hook.
     if (model.$beforeValidate !== model.$objectionModelClass.prototype.$beforeValidate) {
       ctx.jsonSchema = cloneDeep(ctx.jsonSchema);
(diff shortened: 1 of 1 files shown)


koskimas:
What's the use case for this?

lehni:
It's related to [this discsusion](https://gitter.im/Vincit/objection.js?at=5a4763b50163b028109308d8) on Gitter:
>  I was using `$beforeInsert()` to initialize a value that validation needed to already be set, but it looks like `$beforeInsert()` happens after validation. So the remedy seems to be to use `$beforeValidate()`, and then detect the case where the value isn't initialized yet in there. Or is there a better way to go about it?
> Also, `$beforeValidate()` expects a value for `jsonSchema` to be returned, and I just return the one that the method receives as an argument. I was wondering if this could be taken care of internally, by detecting undefined as the return value and using the default `jsonSchema` automatically in that case.

PR #688, opened 2018-01-01
Make QueryBuilder.traverse() aware of range()

lehni:
When `range()` is used before `traverse()`, the result is nested inside the result object, and `traverse()` starts traversing the wrong data structure.

This simple PR fixes this case. If this is deemed useful, I will write unit tests for it.

### DIFF
--- a/lib/queryBuilder/QueryBuilder.js
+++ b/lib/queryBuilder/QueryBuilder.js
@@ -585,7 +585,9 @@ class QueryBuilder extends QueryBuilderBase {
     }
 
     return this.runAfter(result => {
-      this.resultModelClass().traverse(modelClass, result, traverser);
+      // Range operations nest the result data inside the result object.
+      const data = this.has(RangeOperation) ? result.result : result;
+      this.resultModelClass().traverse(modelClass, data, traverser);
       return result;
     });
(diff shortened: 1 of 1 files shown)


lib/queryBuilder/QueryBuilder.js
```diff
@@ -585,7 +585,9 @@ class QueryBuilder extends QueryBuilderBase {
     }
 
     return this.runAfter(result => {
-      this.resultModelClass().traverse(modelClass, result, traverser);
+      // Range operations nest the result data inside the result object.
+      const data = this.has(RangeOperation) ? result.result : result;
+      this.resultModelClass().traverse(modelClass, data, traverser);
```
lehni: @koskimas with #685, this could become `this.has('range')`

lehni:
@koskimas I just checked what other methods use `runAfter()`, and it seems there are two other places where this could be used: `pluch()` and `throwIfNotFound()`. Maybe we could build a little private helper function (`getResultArray(builder, result)`) that returns the actual result array in case `range()` is used?

lehni:
Hmm, I think that's the better idea. I close this and create a new PR.

PR #689, opened 2018-01-01
Make RunAfterOperation aware of range()

lehni:
When `range()` is used before a `RunAfterOperation` (e.g. through `traverse()`), the received result is already nested inside the `range()` result object and needs to be unwrapped again for further processing.

This PR makes `RunAfterOperation` aware of the situation where it is preceded by a `RangeOperation`, and knows how to handle the result.

If this is deemed useful, I will write unit tests for it.

### DIFF
--- a/lib/queryBuilder/QueryBuilderBase.js
+++ b/lib/queryBuilder/QueryBuilderBase.js
@@ -4,6 +4,7 @@ const QueryBuilderOperationSupport = require('./QueryBuilderOperationSupport');
 const isSqlite = require('../utils/knexUtils').isSqlite;
 
 const KnexOperation = require('./operations/KnexOperation');
+const RangeOperation = require('./operations/RangeOperation');
 const SelectOperation = require('./operations/SelectOperation');
 const WhereRefOperation = require('./operations/WhereRefOperation');
 const ReturningOperation = require('./operations/ReturningOperation');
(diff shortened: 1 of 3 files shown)


koskimas:
This is too hacky. `QueryBuilderOperations` should remain as self-contained as possible and should not need to know about other operations and especially not depend on their position in the query. There are some cases currently where that is not true, but they are cases where it just couldn't be avoided. This issue can be avoided by ordering query builder calls correctly.

This would also break the code of people who use `runAfter` with `range` and assume the input is a `Range` object and not an array of items.

lehni:
I thought you would say this. My problem is that I can't really change the sequence in this case, as I'd like to apply default pick / omit settings in `execute()`, if no other were set. And that code always runs last, so always after `range()` and others were called.

I guess I have to explore other options. Would you consider adding some API that would allow me to rearrange the existing `_operations`, e.g. find `range()` and move it to the end?

lehni:
OK, I think I have found a simple workaround. I can't use `QueryBuilder.traverse()`, but I can still use `Model.traverse()`:

```js
      return this.runAfter(result => {
        const data = result && this.has(/^range$/)
          ? result.result
          : result
        this.resultModelClass().traverse(this.modelClass(), data, model => {
          if (this._include) {
            model.$pick(this._include)
          }
          if (this._exclude) {
            model.$omit(this._exclude)
          }
        })
        return result
      })
```

I still think `this.has('range')` would be nicer and faster than `this.has(/^range$/)`...

PR #696, opened 2018-01-05
Do not patch primary ids in upsertGraph()

lehni:
As outlined on [Gitter](https://gitter.im/Vincit/objection.js?at=5a4f937f232e79134dccea2d):

> It looks like the primary id values get reassigned in `upsertGraph()`. I get a lot of statements that looks like this: `update "model_name" set "id" = $1, ... where "model_name"."id" = $X`, with $1 and $X having the same value.

This patch fixes this. It also results in two failing tests, but I have a hunch these may not actually be errors, but a change of behaviour that needs to be tested for differently now.

As always, just a suggestion.

### DIFF
--- a/lib/queryBuilder/operations/UpsertGraphOperation.js
+++ b/lib/queryBuilder/operations/UpsertGraphOperation.js
@@ -154,7 +154,10 @@ class UpsertGraphOperation extends InsertGraphOperation {
 
         // The models were created by the parent class `InsertGraphOperation` with a
         // skipValidation flag. We need to explicitly call $validate here.
-        node.upsertModel.$validate(node.upsertModel, {patch: patch});
+        const json = node.upsertModel.$validate(node.upsertModel, {patch: patch});
+        // Remove the id properties from the json. Note that this does not not modify
+        // node.upsertModel, since $validate() returns a clone.
+        json.$omit(node.upsertModel.constructor.getIdPropertyArray());
 
         if (node.parentNode) {
           // Call patch through parent's $relatedQuery to make things like many-to-many
@@ -163,7 +166,7 @@ class UpsertGraphOperation extends InsertGraphOperation {
             .$relatedQuery(node.relationName)
             .childQueryOf(builder, true)
             .findById(node.upsertModel.$id())
-            [patch ? 'patch' : 'update'](node.upsertModel);
+            [patch ? 'patch' : 'update'](json);
         } else {
           query = node.upsertModel
             .$query()


koskimas:
It should be possible to update the primary key. However `upsertGraph` shouldn't update ids that haven't changed. I've already implemented that. I don't know when that has broken.

lehni:
Do you need a test case for replication?

koskimas:
A test case would be really helpful!

lehni:
@koskimas just use this for `main()` in `reproduction-templates.js`, and you'll see it happening:

```js
async function main() {
  await createSchema();

  ///////////////////////////////////////////////////////////////
  // Your reproduction
  ///////////////////////////////////////////////////////////////

  const jennifer = await Person.query().insertGraph({
    firstName: 'Jennifer',
    lastName: 'Lawrence',

    pets: [
      {
        name: 'Doggo',
        species: 'dog'
      }
    ]
  });

  await Person.query()
    .upsertGraph({
      id: jennifer.id,
      firstName: 'Jen'
    })
    .debug();

  const jen = await Person.query()
    .findOne({firstName: 'Jen'})
    .eager('pets');

  chai.expect(jen.pets[0].name).to.equal('Doggo');
}
```
Output:
```js
{ method: 'update',
  options: {},
  timeout: false,
  cancelOnTimeout: false,
  bindings: [ 1, 'Jen', 1 ],
  __knexQueryUid: 'ac5cd5dc-7ed0-4a3a-b782-2119b3cb790c',
  toNative: [Function: toNative],
  sql: 'update `Person` set `id` = ?, `firstName` = ? where `Person`.`id` = ?' 
}
```

koskimas:
Oh, that's expected. I thought you meant you get update statements with __only__ an id property.

koskimas:
Actually we could still avoid that. I'll see what I can do.

lehni:
I don't think it's possible to reassign ids in an `upsertGraph()` operation because the ids are used to identify the models that need changing. Therefore I thought that removing the ids from the data before each `patch()` / `update()`opdate operation should be the right way to go about it?

koskimas:
Relations don't have to use id columns. Relations can be from any column to any column. Also you should be able to update root models.

koskimas:
I have a fix for this almost ready. I'll post here when I have something for you to test.

koskimas:
@lehni I tried to add both optimizations you mentioned in gitter 
0dd4039

lehni:
Cheers, thanks! Will test now.

PR #699, opened 2018-01-06
Koa example

jtlapp:
Port of `example/express-es7` to koa.


jtlapp:
@lehni, I believe I noticed that you were working with koa. Would you mind giving my [middleware stack](https://github.com/Vincit/objection.js/pull/699/files#diff-6df55493de2a26fcdba9094340782b9e) a once-over?

jtlapp:
Seems to be a problem with Travis. Maybe run it again?

koskimas:
No it's not a travis problem. You need to run `npm run prettier` to fix the style.

lehni:
@jtlapp looks good to me!

I have a small difference in my `errorHandler()` that leads to all errors being rendered as JSON. Basically, when nothing else can be rendered, it outputs `err.message` with a fall-back as an error object:

```js
ctx.body = {
  message: err.message || 'An error has occurred.'
}
```

jtlapp:
Great, thanks @lehni. I think I'll leave the error handler as is, because I copied that forward from the express-ts example (see [here](https://github.com/Vincit/objection.js/blob/master/examples/express-es7/src/app.js#L33)).

lehni:
I changed it for consistency because every other error from Objection results in JSON being output.

jtlapp:
Okay, I don't have enough experience with objection to know which error output is best. @koskimas?

jtlapp:
@koskimas, prettier is happy with this PR as is. What's wrong this time?

jtlapp:
Okay @lehni, the error output is indeed wonky, but I believe the problem is at [`ctx.app.emit()`](https://github.com/Vincit/objection.js/pull/699/files#diff-6df55493de2a26fcdba9094340782b9eR26). The server log is dumping stack traces when I issue bad requests. I'm investigating.

I think I also prefer that the errors always be JSON, so I'll do that too.

jtlapp:
The problem is that the error handler needs to distinguish between validation errors and internal errors. Internal errors should indeed dump stack traces.

jtlapp:
I made the error handling [more robust](https://github.com/Vincit/objection.js/pull/699/files#diff-6df55493de2a26fcdba9094340782b9eR21). The express-es7 example should probably be modified to do something similar.

Also, why doesn't ValidationError set `err.message` to a readable string? Because of the local language problem? It's really weird to find serialized JSON in `err.message`.

P.S. Travis is failing on prettier's own prettying. Not sure what else I can do.

koskimas:
You need to run prettier just before committing. Thats all there is to it. Are using yarn? If so you may need to update the lock file. Maybe it has an old prettier version in it.

jtlapp:
I locally checked in 5c25406 prior to prettying, prettied, then checked in 6241fac. It's reporting the error on the latter, but maybe it coughs if any commit prior to push isn't prettied?

I just previously ran `npm install -g prettier`. I have prettier v1.9.2 installed, which npmjs says is the most recent version.

Navigating the details above, I found [this error](https://travis-ci.org/Vincit/objection.js/jobs/325894213#L3323), which seems to be the test case "integration tests postgres Model insertGraph queries .query().insertGraphAndFetch() should insert a model with relations and fetch the inserted graph."

koskimas:
objection doesn't use the global prettier, but the local one. The `insertGraphAndFetch` test seems to be failing randomly. I'll fix that.

koskimas:
This seems to be 99% equal to the es7 example. I don't think this should be added. It's up to me to keep this example updated once it's merged and I don't want to add another thing on my todolist. As I mentioned in gitter, a full-blown koa project with a good architecture authentication etc. could be useful since there is no real project example.

jtlapp:
That's fine. It was a good exercise for me, revealing that objection requires special error handling.

I think you're still missing an example that fully leverages TS instead of just minimally porting ES7, as `express-ts` does. I'm working on one for my own benefit, will share, no need to include it.

Thanks for considering this PR!

PR #708, opened 2018-01-09
Added missing relation types to express-ts

jtlapp:
This PR includes the following mods to the `express-ts` example:

(1) Added missing eager relation properties to Person.
(2) Made all eager relation properties optional.
(3) Added parentId to Person.
(4) Added ownerId to Animal.

None of these oversights caused errors in the example, because the end points where not directly referencing these properties, just passing them back to the client.

I only propose these mods to demo the general case.

In order to avoid circular dependencies, which Typescript disallows, I had to create interfaces for all of the models and put them all in the same file, `models/interfaces.ts`. Each models implements its interface, and `models/Person.ts` re-exports the Address interface.

It's not ideal in the general case to dump interfaces for all interrelated models into the same file, but I don't see how else to do it. If there's a better practice to demo, please do tell me!

What think you, @mceachen? Does anything need fixing? Thanks!

P.S. I have a mocha test suite that will work on `express-ts` with a few additional `get` end points. I made it for a koa-ts port I've been working on. I verified the above edits by diffing with the prior output of `example-requests.sh`.


jtlapp:
Ah poo. Forgot to pretty it. Just a sec...

jtlapp:
Travis was happy despite prettier finding things to change in the PR. I pushed the prettier output.

examples/express-ts/src/models/Animal.ts
```diff
@@ -1,13 +1,16 @@
...
-export default class Animal extends Model {
+export default class Animal extends Model implements Interfaces.Animal {
   readonly id: number;
-  owner: Person;
+  ownerId: number | null;
   name: string;
   species: string;
 
+  // Optional eager relations.
+  owner?: Interfaces.Person;
```
mceachen: yup, that's better. 👍

examples/express-ts/src/models/interfaces.ts
```diff
@@ -0,0 +1,42 @@
+import {Model} from 'objection';
```
mceachen: If you add this, please wrap it with a big jsdoc disclaimer that it's completely optional. I think people will most likely use the express-ts template as an example of best-practices, and I've found this approach to be decidedly non-DRY while not adding additional type safety.

I'd remove it, personally.

jtlapp: I'm not sure what you mean by this. I don't see how you can reference eager relation properties in a type-safe way without using these interfaces. If you try to do it without interfaces, TS barfs on circular dependency. Unless you can show me another way.

Yeah, interfaces are always non-DRY.

jtlapp: Ah scheisse. I just stripped out the interfaces and now it's working despite the cyclic dependency. I don't know what changed. I'll push it.

jtlapp: I've had a steep learning curve with Typescript. It has to be the most complex language I've ever tackled. Even C++ was more predictable.

jtlapp:
It's now working on my end, despite removing interfaces and now having cyclic dependency (an import loop). I don't understand why it wasn't working before.

Would you mind making sure it works for you too? I'm kinda weirded out that (apparently) what once wasn't working now is.

jtlapp:
Hey @mceachen. I seem to have found another type problem. It occurs in multiple places, but consider [one of them](https://github.com/Vincit/objection.js/blob/master/examples/express-ts/src/api.ts#L149):

```
// api.ts

    const actors = await movie.$relatedQuery('actors');
```

`actors` is of type `Model[]`, which doesn't give access to Person properties.

If we instead change this to the following:

```
// proposed api.ts

    const actors = await movie.$relatedQuery<Person>('actors');
```

Now `actors` is of the correct type `Person[]`.

I don't understand why the typings don't properly infer this. Ideally we'd find a way to make the typings smarter, instead of having to pass a type parameter. I do recall reading somewhere that if no type parameter is passed to a generic function, the type defaults to `any`.

jtlapp:
Another thought. If a client can supply the eager string, then the eager relation properties could potentially have all sorts of nested properties, to an undetermined level of nesting.

So perhaps the properties should have be of type `any`?

```ts
  parent?: Person;
  children?: Person[];
  pets?: Animal[];
  movies?: Movie[];
```

P.S. I find types extraordinarily complicated in Typescript, but I guess that's just a consequence of trying to type (and infer types) in a prototypal language.

jtlapp:
To summarize, these are the outstanding questions:

1. How should we handle the variable nature of eager relation types? Two possible options: (a) declare such properties as `any` or `any[]`; or (b) declare them for the first level of nesting and require the app to re-open the type for declaration merging to add more.
2. How should we get `$relatedQuery()` to return anything other than `Model`? Two possible options: (a) pass a type parameter to `$relatedQuery<MyModel>()`; and (b) find a way to fix the typings. I was unable to fix the problem with typecasting.
3. BONUS: Why isn't Typescript barfing on the circular dependencies of the current PR (sans interfaces)? There are import loops.

mceachen:
Re: 1: I've found the current typings work fine.

Re: 2: that's already doable with a cast, but I realized with TS 2.1 we could do a bit better: I just submitted #709 to make this automagic with `keyof`.

Re: 3: you only get an import loop if the definition is non-deferrable. I have bi-directional references to many models in my tree, and don't have any issues.

Re: PS: Try scala.

jtlapp:
> PS: Try scala.

Kotlin's next for me, after I finish up my current Java project. (My TS/JS project has no end.) My sojourns to understand TS typing occasionally land me in Scala or Kotlin land. People complain that Scala gives devs too much style freedom, making code difficult to inherit and maintain. I may have a glimpse of that seeing TS's type definition flexibility.

PR #718, opened 2018-01-11
Switch to Ajv errorDataPath: 'property' configuration

lehni:
As mentioned today on [Gitter](https://gitter.im/Vincit/objection.js?at=5a57499ab48e8c3566c8fdd2), this PR switches to using Ajv's `errorDataPath: 'property'` configuration to always receive property data-paths in validation errors, leading to some nice simplifications and more consistent error messages.

### DIFF
--- a/lib/model/AjvValidator.js
+++ b/lib/model/AjvValidator.js
@@ -7,7 +7,7 @@ class AjvValidator extends Validator {
   constructor(conf) {
     super();
 
-    this.ajvOptions = Object.assign({}, conf.options);
+    this.ajvOptions = Object.assign({ errorDataPath: 'property' }, conf.options);
 
     // Create a normal Ajv instance.
     this.ajv = new Ajv(
@@ -15,14 +15,14 @@ class AjvValidator extends Validator {
         {
           useDefaults: true
         },
-        conf.options
+        this.ajvOptions
       )
     );
 
     // Create an instance that doesn't set default values. We need this one
     // to validate `patch` objects (objects that have a subset of properties).
     this.ajvNoDefaults = new Ajv(
-      Object.assign({}, conf.options, {
+      Object.assign({}, this.ajvOptions, {
         useDefaults: false
       })
     );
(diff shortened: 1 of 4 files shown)


lehni:
One test is currently failing due to this issue here: https://github.com/epoberezkin/ajv/issues/671

Let's see how this goes first.

lehni:
@koskimas I've addressed the issue now in 765265325834807ce8836f7345471dab4bfc7686

Please have a look and let me know what you think about this.

lehni:
BTW, this was an error in the previous implementation:

https://github.com/Vincit/objection.js/pull/718/commits/765265325834807ce8836f7345471dab4bfc7686#diff-36a59a5729587eeda0ab504c598fe3e8L143

The path wasn't pointing to the additional property there, but was doing so at root level.

lib/model/AjvValidator.js
```diff
@@ -157,25 +157,19 @@ function parseValidationError(errors, modelClass, options) {
...
-
-    if (options.dataPath) {
-      key = `${options.dataPath.substring(1)}.${key}`;
-    }
-
+    const dataPath = `${options.dataPath || ''}${error.dataPath}`;
+    // Unknown properties are reported in `['propertyName']` notation,
+    // so replace those with dot-notation, see:
+    // https://github.com/epoberezkin/ajv/issues/671
+    const key = dataPath.replace(/\['([^']*)'\]/g, '.$1').substring(1);
```
koskimas: Doesn't this replace all brackets with dot notation? For example `foo['words lol'].bar` --> `foo.words lol.bar` which is an invalid path?

lehni: Note that the dataPaths all start with `.`and we remove it with `.substring(1)`

koskimas: The correct path for `foo['words lol'].bar` is `foo['words lol'].bar`, not `foo.words.lol.bar`.

koskimas: Keys can have spaces and in that case bracket notation must be used.

koskimas: Actually it seems that lodash accepts spaces with dot notation:
```js
_.get({a: {"b c": 1}}, "a.b c"); // --> 1
```

lehni: Oh I sorry, I didn't see the space because of the way the line broke.

lehni: that can be easily remedied. New version coming up.

lehni: This works:
```js
`foo['words lol'].bar['additional']`.replace(/\['([^' ]*)'\]/g, '.$1')
```

lehni: Which version do you prefer?

lehni: Also, are you sure that Objection relations or properties can have spaces in them?

lehni: The only difference is the added space in `[^']` → `[^' ]`. So either scenario can be addressed easily. I've pushed the version that handles the space in e8ec684982e2b16b6ae7319a9130eff82a2ab30f.

PR #722, opened 2018-01-12
Remaining validation typings tests

jtlapp:
I think this does it for #664, except for testing ModelClass methods, which the tests don't otherwise seem to do. (Hitting the sack now, out of the loop for next 12 hours.)

cc @mceachen


tests/ts/examples.ts
```diff
@@ -12,6 +13,27 @@ const { lit, raw, ref } = objection;
 
 class CustomValidationError extends Error {}
 
+class CustomValidator extends objection.Validator {
```
mceachen: I think we might want to add a quick comment here saying this is not example code for people to read to understand the API, but just the minimal stubs to exercise the typings, perhaps?

mceachen: Done.

mceachen:
There was a merge conflict with #753. I merged this with that PR, and resubmitted that PR, so we can close this.

Thanks for the PR, @jtlapp !

PR #729, opened 2018-01-17
Fix mergeEager() for child expressions with namedFilters

lehni:
As mentioned on [Gitter](https://gitter.im/Vincit/objection.js?at=5a5f9104b48e8c3566f10cf8):

I've developed a habit of defining `namedFilters` recursively throughout graphs. .e.g if I have three nested models that relate to each  other, and they all define a filter named `someFilter`, I use eager statements in each to load just the next level of relations, e.g.:
```js
const namedFilters = {
  someFilter: query => query.mergeEager('relationName(someFilter)')
}.
```

This works well, but as soon as I combine it with another filter that also defines a `mergeEager()` statement with child filters on the same children, the eager statements don't seem to get merged, and only the one that is applied last appears to be used as the eager statement on the child expression. If I flatten the statement and remove the recursive filter calls, then the merging appears to work.

This PR fixes this issue by allowing `mergeEager()` to merge arrays within the children expression by the use of array concatenation, thus correctly merging the `args` properties of the children expressions.

### DIFF
--- a/lib/queryBuilder/RelationExpression.js
+++ b/lib/queryBuilder/RelationExpression.js
@@ -1,4 +1,5 @@
-const merge = require('lodash/merge');
+const mergeWith = require('lodash/mergeWith');
+const union = require('lodash/union');
 const clone = require('lodash/clone');
 const values = require('lodash/values');
 const cloneDeep = require('lodash/cloneDeep');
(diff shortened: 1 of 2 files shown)


lehni:
I just realized that this that this statement also doesn't merge the `args` arrays correctly:
```js
merged.args = merge(merged.args, expr.args);
```
The result is just the value of `expr.args`. I am going to address this in this PR as well.

lehni:
instead of `concat()` we should probably use `_.union()` to avoid duplicates. Fix coming up...

lehni:
I knew you where going to say that :D 

I will try to come up with one, but it's quite tricky for me. Where is the right place for it?

lehni:
Working on this, I just realised that the `RelationExpression` code internally sometimes uses `this._filters`, and sometimes the getter for it named `this.filters`. What's the reason for doing both? Wouldn't it be better to always use `_filters` internally for performance?

lehni:
@koskimas are these tests good enough? They both fail without the change:
https://github.com/Vincit/objection.js/pull/729/commits/76064e1e1ba20417568237bf816552788c675627#diff-a5fa5cd1d839edc12779b0b64a8d35a1R753

koskimas:
Yes, those are enough. Thank you once again!

koskimas:
Are you done with this? Can I merge it?

lehni:
Yes, go ahead! Thanks!

Issue #724, opened 2018-01-13
Graph typings are incompatible with eager relation properties


jtlapp:
UPDATE: This is actually a Typescript-specific usage problem whose solution is not currently modeled in the examples.

The typings for `insertGraph()` and `upsertGraph()` don't allow graphs that include eager relation properties.

The problem is that these methods take `Partial<T extends Model>` as input parameters, but this only makes the eager relation properties optional and not also their properties, recursively. In particular, they expect an eager relation property to have all the methods of Objection.Model.

Here are some examples and their errors:

```ts
import * as Objection from 'objection';

class Person extends Objection.Model {
  id?: number;
  firstName: string;
  lastName: string;

  // Commenting the following out doesn't help, as then movies wouldn't be valid in Partial<T>
  movies?: Movie[];
}

class Movie extends Objection.Model {
  id?: number
  name: string;
}

Person
  .query()
  .insertGraph({
    firstName: 'Jennifer',
    lastName: 'Lawrence',

    movies: [{ // ERROR: missing all of Model's properties
      // id: 32 (adding this wouldn't help)
      name: 'American Hustle'
    }]
  });

Person
  .query()
  .insertGraph({
    firstName: 'Jennifer',
    lastName: 'Lawrence',

    movies: [{ // ERROR: missing all of Model's properties
      id: 32
      // name: 'American Hustle' (adding this wouldn't help)
    }]
  }, {
    relate: true
  });

Person
  .query()
  .insertGraph([{
    firstName: 'Jennifer',
    lastName: 'Lawrence',

    movies: [{ // ERROR: does not match movies property
      "#id": 'silverLiningsPlaybook'
      name: 'Silver Linings Playbook'
    }]
  }, {
    firstName: 'Bradley',
    lastName: 'Cooper',

    movies: [{ // ERROR: does not match movies property
      "#ref": 'silverLiningsPlaybook'
    }]
  }]);

Person
  .query()
  .upsertGraph({
    id: 1,
    firstName: 'Jonnifer',

    movies: [{ // ERROR: missing all of Model's properties
      id: 1
      //name: 'American Hustle' (adding this wouldn't help)
    }, { // ERROR: missing all of Model's properties
      id: 1253
      // name: 'Passengers' (adding this wouldn't help)
    }]
  });
```

I've been exploring ways to improve these typings in #714, but have not found anything that works other than the `any` type.

[...]

Issue #726, opened 2018-01-14
Example Typescript wrapper for Objection


jtlapp:
Okay @koskimas and @mceachen, I'm ready to share the project. It works, but there a few more things I want to do.

I've ported the `express-ts` example to [an example I'm calling `ts-wrapper`](https://github.com/jtlapp/objection.js/tree/ts-wrapper-example/examples/ts-wrapper). The README page at that link provides an overview. (`src/repl.js` is there only because I haven't moved it to `test.js` yet.)

In short, I'm demonstrating a way to encapsulate the Objection.js implementation behind Typescript interfaces to minimize the amount of code that depends on Objection.js. It happens to demonstrate this abstraction with a koa frontend, but I'm not wedded to that.

I'm sharing it for any feedback you might have and for you to consider whether it (or any part of it) might be useful to include within the Objection repo. No skin off my back either way, as I did it mainly to help me learn Objection and figure out a way to integrate Objection into my project.

I plan to write more tests, because doing so helps me understand how things work and occasionally reveals typings problems or my misunderstandings of the `express-ts` example. The mocha tests are currently just a port of `express-ts/example-requests.sh`, but I plan to test my repo API more directly with a port of `express-es7/client.js` to mocha tests.

I also have a number of unresolved TBDs to investigate. You can ignore all that.

Thanks!

mceachen:
I looked through your repo.

IMHO, I think the typescript example should match as closely as possible to @koskimas' esnext example. Indeed, just a couple PRs ago, I cloned the esnext tree and replaced the ts tree with the esnext code with the minimal changes necessary to make TypeScript happy in strict mode.

Introducing interfaces and services and brokers layers to example code adds both conceptual and implementation weight that, frankly, makes Objection appear to be harder to use than it actually is. I don't favor importing this.

mceachen:
That said, I think replacing the example-requests.sh with a mocha test suite would be wonderful (and we should push that over to the esnext tree as well, as a validation test before npm publishing).

jtlapp:
Thanks @mceachen!

> the typescript example should match as closely as possible to @koskimas' esnext example.  Indeed, just a couple PRs ago, I cloned the esnext tree and replaced the ts tree...

Where's this esnext tree? I search through the repo just yields some examples of static initializers.

> Introducing interfaces and services and brokers layers... makes Objection appear to be harder to use than it actually is.

I agree on that point.

[...]

PR #735, opened 2018-01-18
Mocha test for express-ts with a few type mods

jtlapp:
Here's my mocha test suite adapted for the `express-ts` example, per [this issue comment](https://github.com/Vincit/objection.js/issues/726#issuecomment-357834865).

I had to add a few GET requests to the server API for verifying activity.

I also changed eager relation types from `Model` to `Partial<Model>` to demonstrate how this should be done without type mismatches, which should resolve issue #724.

Lastly, the new `$relatedQuery` typings only help in the case where the method is called on a model of the expected return type. If the return type is different, you have to type-parameterize the `$relatedQuery` call to the expected type. I modified the server API to demo that as well.

cc @mceachen


jtlapp:
Any advice on how to avoid all those pull request commit notices?

I've been pulling updates to my fork master via the github UI, then pulling them down to my local repo. Will these commit notices not appear if I do this in reverse, pulling from the upstream master and pushing to my github fork?

jtlapp:
Also, it should be easy to adapt the other examples to also work with this mocha test.

examples/express-ts/test.js
```diff
@@ -0,0 +1,357 @@
...
+  }
+  for (const property in expected) {
+    assert.deepEqual(actual[property], expected[property], `property ${property}`);
+  }
+}
+
+function clearDatabase(knex) {
+  const tables = ['Person', 'Movie', 'Animal', 'Person_Movie'];
+  const jobs = [];
+  tables.forEach(table => {
```
mceachen: Do you know about `.map`? This whole method is a one-liner:

```js
return Promise.all(['Person', 'Movie', 'Animal', 'Person_Movie'].map(table => knex(table).del()))
```

jtlapp: Yeah, I guess it didn't come to mind. I can do it in one line, but it might be clearer in two:

```js
const tables = ['Person', 'Movie', 'Animal', 'Person_Movie']
return Promise.all(tables.map(table => knex(table).del()))
```

Thanks for the reminder! I still mostly think in Java.

examples/express-ts/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "clean": "rimraf example.db lib",
     "compile": "tsc",
+    "test": "mocha test.js",
```
mceachen: Not a merge blocker, but I think it'd be nice to automatically clean and compile before running the tests. If you add `npm-run-all` as a dev dependency, we can do this:

```json
    "pretest": "run-s clean compile"
```

(and while we're at it, change `start` to have a `prestart` instead of the `&&` which may not work on windows boxen under non-bash shells).

If you don't want to do this diff, I can do it after this PR merges. Your call.

examples/express-ts/src/api.ts
```diff
@@ -109,7 +117,7 @@ export default (router: express.Router) => {
       throw createStatusCodeError(404);
     }
 
-    const pet = await person.$relatedQuery('pets').insert(req.body);
+    const pet = await person.$relatedQuery<Animal>('pets').insert(req.body);
```
mceachen: Did you have to add these because we're using an older version of objection?

examples/express-ts/src/models/Animal.ts
```diff
@@ -9,7 +9,7 @@ export default class Animal extends Model {
   species: string;
 
   // Optional eager relations.
-  owner?: Person;
+  owner?: Partial<Person>;
```
mceachen: Why Partials? That's surprising to me.

examples/express-ts/test.js
```diff
@@ -0,0 +1,357 @@
+const axios = require('axios');
```
mceachen: 🥇 thanks for writing this! 🎉

examples/express-ts/src/api.ts
jtlapp: Depends on how old. Your recent typings change for `$relatedQuery` gets its type from the model its called on. That would yield a Person in this case. We need an Animal.

examples/express-ts/src/models/Animal.ts
jtlapp: I'm confused about your confusion. This is our conclusion from #724. It's not strictly needed in this particular example, but not having it demoed anywhere in the examples lead me on quite a goose chase trying to find typings that generally work with eager relations.

examples/express-ts/test.js
jtlapp: You're welcome! Thanks for writing `express-ts` and the `.sh` I based this on!

examples/express-ts/package.json
jtlapp: You'd need to build and run the server in a separate process too.

I'm no script whiz. You might want to do this to get things the way you want it.

examples/express-ts/src/api.ts
jtlapp: You'd have to make the typings infer the return type from the provided relation name.

!!!!!!

examples/express-ts/src/models/Animal.ts
mceachen: I was surprised that you used a Partial in the model specification. I expected them to only exist in querybuilder responses.

jtlapp: It was your suggestion, and it made things work:
https://github.com/Vincit/objection.js/issues/724#issuecomment-357461867

jtlapp: I notice that it's not being tested in `examples.ts`. Were we to test it there, we'd probably make the relation properties all partials in the tests too.

mceachen:
@koskimas I've run this path with #753 locally and everything looks OK. Are you OK with me squashmerging this?

Issue #734, opened 2018-01-18
Add base classes for relate and unrelate operations


koskimas:
As discussed in gitter

lehni:
@koskimas I think I can take care of this one.

lehni:
#736 is implementing this

PR #736, opened 2018-01-18
Implement new QueryBuilder.has*() and is*() checks

lehni:
As discussed, this PR adds `has(str), isFind(), isInsert(), isUpdate(), isDelete(), isRelate(), isUnrelate(), hasWheres(), hasSelects(), hasEager()`

It also addresses issue #734, since that's a requirement for this to work.

- This is work in progress, submitted for review, feedback, and further work.
- Unit tests are currently missing.
- API and docs changes are currently missing.

I will add some comments on code lines in the commit itself.

### DIFF
--- a/lib/queryBuilder/QueryBuilder.js
+++ b/lib/queryBuilder/QueryBuilder.js
@@ -15,6 +15,8 @@ const FindOperation = require('./operations/FindOperation');
 const DeleteOperation = require('./operations/DeleteOperation');
 const UpdateOperation = require('./operations/UpdateOperation');
 const InsertOperation = require('./operations/InsertOperation');
+const RelateOperation = require('./operations/RelateOperation');
+const UnrelateOperation = require('./operations/UnrelateOperation');
 
 const InsertGraphAndFetchOperation = require('./operations/InsertGraphAndFetchOperation');
 const UpsertGraphAndFetchOperation = require('./operations/UpsertGraphAndFetchOperation');
@@ -365,12 +367,60 @@ class QueryBuilder extends QueryBuilderBase {
     return this._resultModelClass || this.modelClass();
   }
 
+  isFind() {
+    return !(
+      this.isInsert() ||
+      this.isUpdate() ||
+      this.isDelete() ||
+      this.isRelate() ||
+      this.isUnrelate()
+    );
+  }
+
+  isInsert() {
+    return this.has(InsertOperation);
+  }
+
+  isUpdate() {
+    return this.has(UpdateOperation);
+  }
+
+  isDelete() {
+    return this.has(DeleteOperation);
+  }
+
+  isRelate() {
+    return this.has(RelateOperation);
+  }
+
+  isUnrelate() {
+    return this.has(UnrelateOperation);
+  }
+
+  hasWheres() {
+    return this.has(QueryBuilderBase.WhereSelector);
+  }
+
+  hasSelects() {
+    return this.has(QueryBuilderBase.SelectSelector);
+  }
+
+  hasEager() {
+    return !!this._eagerExpression;
+  }
+
   isFindQuery() {
-    return !this._operations.some(op => op.isWriteOperation) && !this._explicitRejectValue;
+    console.warn(
+      `isFindQuery is deprecated. Use isFind instead. This method will be removed in version 2.0`
+    );
+    return this.isFind();
   }
 
   isEagerQuery() {
-    return !!this._eagerExpression;
+    console.warn(
+      `isEagerQuery is deprecated. Use hasEager instead. This method will be removed in version 2.0`
+    );
+    return this.hasEager();
   }
 
   toString() {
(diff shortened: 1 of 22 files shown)


lib/queryBuilder/QueryBuilder.js
```diff
@@ -365,12 +367,63 @@ class QueryBuilder extends QueryBuilderBase {
     return this._resultModelClass || this.modelClass();
   }
 
+  isFind() {
+    return (
+      !this._explicitRejectValue &&
```
lehni: @koskimas why is this check for `!this._explicitRejectValue` here? I've removed it to see if things change, and all tests still pass...

lib/relations/belongsToOne/BelongsToOneUnrelateOperation.js
```diff
@@ -11,6 +11,27 @@ class BelongsToOneUnrelateOperation extends BelongsToOneRelateOperation {
     this.ids = [ids];
     return true;
   }
+
+  queryExecutor(builder) {
+    // NOTE: This is identical to BelongsToOneRelateOperation.queryExecutor()
```
lehni: We could also put a part of this code into a helper function and reuse it?

tests/unit/queryBuilder/QueryBuilder.js
```diff
@@ -1881,8 +1881,12 @@ describe('QueryBuilder', () => {
   });
 });
 
+const factoryBuilder = new QueryBuilder(Model);
+
 function createFindOperation(builder, whereObj) {
-  let method = new QueryBuilderOperation('find');
+  const method = factoryBuilder._findOperationFactory(builder);
+
+  method.onBefore2 = method.onAfter2 = null;
```
lehni: Getting these tests to work again needed some trickery, since they were not actually using the right `QueryBuilderOperation` subclass. I'm accessing the private `_findOperationFactory()` & co. methods, and then erasing `method.onBefore2 = method.onAfter2 = null;`. Not ideal, but I couldn't figure out another way without having to export private classes...

lib/queryBuilder/QueryBuilder.js
```diff
@@ -365,12 +367,63 @@ class QueryBuilder extends QueryBuilderBase {
...
+  }
+
+  hasEager() {
+    return !!this._eagerExpression;
+  }
+
   isFindQuery() {
-    return !this._operations.some(op => op.isWriteOperation) && !this._explicitRejectValue;
+    console.warn(
+      `isFindQuery is deprecated. Use isFind() instead. This method will be removed in version 1.5`
```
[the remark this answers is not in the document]
koskimas: I'm afraid it's 2.0 since the next version is 1.0 and there can be no breaking changes before 2.0.

lib/relations/belongsToOne/BelongsToOneUnrelateOperation.js
koskimas: Nah, don't bother. It's not that many lines

lib/queryBuilder/QueryBuilder.js
```diff
@@ -365,12 +367,63 @@ class QueryBuilder extends QueryBuilderBase {
...
 
+  isFind() {
+    return (
+      !this._explicitRejectValue &&
+      !(
+        this.isInsert() ||
+        this.isUpdate() ||
+        this.isDelete() ||
+        this.isRelate() ||
+        this.isUnrelate()
```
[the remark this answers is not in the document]
koskimas: The five loops hurt my inner premature optimizer, but this way it's cleaner. `isFind` is only called once per (normal) query, so it shouldn't matter. Let's leave it like this

lib/queryBuilder/QueryBuilder.js
koskimas: I have no idea. Let's remove it if no test fail

lehni:
@koskimas I've addressed your feedback and pushed a new version now. I think with docs and tests this should be good to go?

lehni:
@koskimas should we add check for `reject()` / `resolve()` also, e.g. `isRejected()` / `isResolved()`?

koskimas:
And documentation is also missing

lehni:
Of course! Just wanted to settle on an implementation first.

koskimas:
Oh, sorry! I didn't notice the

>  I think with docs and tests this should be good to go

😄

lehni:
No worries : ) Did you see this question? I think it could be useful:

https://github.com/Vincit/objection.js/pull/736#issuecomment-358912245

koskimas:
Those might be confused with the promise state (did the query fail/succeed). If we add something like that it would have  to be something like `isExplicitlyResolved` but I don't think those are really useful.

lehni:
@koskimas this is the whole shebang in one commit now. It took about twice as long to write the tests and docs than do the actual change : )

Here are the unit tests: https://github.com/Vincit/objection.js/pull/736/commits/a209c24c27969aa9575a8545c4dcbeef9145eb33#diff-989941cbe3736417a2af56417e714c6dR762

I think they're quite detailed. Please take a look and let me know your thoughts.

koskimas:
>  It took about twice as long to write the tests and docs than do the actual change

Yup, that's what I've learned too 😄 

The tests and docs look perfect, thanks! By the way, If you want to add yourself as a contributor in package.json, fell free to do so.

lehni:
@koskimas cool, thanks! Yes that'd be nice! I'll do so it in a PR.

PR #739, opened 2018-01-19
Make fact that $relatedQuery instances are stored on owner more apparent.

newhouse:
@koskimas given the similarities of $relatedQuery and $loadRelated, I thought it was important to highlight this fact in right in the main description of the method. I don't think this is overkill or misleading, and hopefully you agree.

### DIFF
--- a/doc/includes/API.md
+++ b/doc/includes/API.md
@@ -6822,7 +6822,7 @@ const builder = model.$relatedQuery(relationName, transactionOrKnex);
 ```
 
 > Fetch all models related to a model through a relation. The fetched models are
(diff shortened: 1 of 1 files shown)


koskimas:
Good idea. Maybe we could also mention the [`relatedFindQueryMutates`](http://vincit.github.io/objection.js/#relatedfindquerymutates) and  [`relatedInsertQueryMutates`](http://vincit.github.io/objection.js/#relatedinsertquerymutates) configs for changing that behaviour?

newhouse:
done.

koskimas:
Thanks!

Issue #742, opened 2018-01-21
Result type pass-through errors in typings


jtlapp:
Hey @mceachen. [Sami suggested](https://github.com/Vincit/objection.js/issues/717#issuecomment-359277051) that I open a new issue for this, though it's a continuation of the sorts of problems in #717. Boy did I learn a lot studying your typings!

I'm sharing from my fork before making any PRs:

- [These new typings tests](https://github.com/Vincit/objection.js/compare/master...jtlapp:pass-thru-errors) include a number of errors mostly related to not passing the result value along the chain of query builder calls. The only tests in here that do work are those testing `throwIfNotFound()`, which I've added because they helped me test my mods.
- [These revised typings](https://github.com/Vincit/objection.js/compare/master...jtlapp:pass-thru-typings) that resolve the above new tests. The type parameterization of QueryBuilder in a few of the Typescript tests also had to change.

I made the following sorts of changes to the typings, going from memory:

- Query builders now pass the expected result type along the chain. The declaration is now `QueryBuilder<QM, RM, RV>`, where `QM` = queried model, `RM` = candidate result model or model array, and `RV` = actual result value. `QM` is different from `RM` because `RM` can be either `QM` or `QM[]`. `RV` can be `QM`, `QM[]`, `QM | undefined`, and `number`.
- The query builder interfaces went through quite an evolution but they eventually came back to something like you originally had. The names have changed to identify the functionality they offer, so that may be understood without having to look up the interface. However, I recognize there is a drawback to not naming things according to their SQL correlates (e.g. insert, update, delete, etc.), because odds are that the typings will evolve according to their correlates. I can deal with either approach but wanted to present an alternative. I can definitively say that the approach I'm illustrating would have helped me a lot scanning through the typings.
- Generic parameters all have suggestive names. The `T` that originally appeared throughout threw me for a loop because I was confused about whether that was meant to represent a model or a QueryBuilder or a transaction or what. It took examining the typings at length to understand what it was intended to mean.
- Page query builders now support the `runAfter()` method.

The following remains to be done. I've given them a preliminary look and have questions about both. It would be helpful to collaborate.

- Possibly accommodate any QueryBuilder result type returned by callbacks, per [Sami's comment](https://github.com/Vincit/objection.js/issues/717#issuecomment-359276710). It may be that just the eager [modify/filter callbacks](https://github.com/jtlapp/objection.js/blob/pass-thru-typings/typings/objection/index.d.ts#L190) need a type `any`.
[...]

PR #744, opened 2018-01-22
Typings revised for passing result type down the query builder chain

jtlapp:
Typings revision per our conversation, @mceachen, for #742. Also had to prettier it up.

I think we left one topic unresolved: defaulting `QueryBuilder<M>` to yielding an `M[]` instead of an `M`, as things stand.

Not the problem: ~~I think the main problem with changing it to `M[]` is that, because we can only default unspecified parameters on the right, we'd end up specifying all the parameters `QueryBuilder<RM, RV, QB>` or `QueryBuilder<RM, QB, RV>` in many typings occurrences that are now just singlets or couplets. Putting `RV` last would reduce that some, but it might be a bit confusing to see the type in the form `QueryBuilder<return, queried, return>`, since the return types aren't adjacent.~~

Possible problem: We can infer `M[]` from `M` but not `M` from `M[]`.

UPDATE: I created an [experimental branch offshoot](https://github.com/jtlapp/objection.js/tree/model-array-default) of this PR that defaults to `Model[]`, but I don't have `$relatedQuery()` working.


tests/ts/examples.ts
```diff
@@ -285,8 +285,8 @@ qb = qb.where(Person.raw('random()', 1, '2', raw('3')));
 qb = qb.alias('someAlias');
 
 // query builder hooks
-qb = qb.runBefore(async (result: any, builder: objection.QueryBuilder<Person>) => {});
-qb = qb.runAfter(async (result: Person[], builder: objection.QueryBuilder<Person>) => {});
+qb = qb.runBefore(async (result: any, builder: objection.QueryBuilder<Person, Person[]>) => {});
+qb = qb.runAfter(async (result: Person[], builder: objection.QueryBuilder<Person, Person[]>) => {});
```
koskimas: One should return stuff from thses. Whatever you return from `runBefore` gets passed to the next `runBefore` function as the first argument. That's impossible write typings for, but return value should be required.

Whatever you return from `runAfter` ends up being the return value of the query (in addition to being the input for the next `runAfter`). This we probably could write typings for?

tests/ts/examples.ts
```diff
@@ -584,6 +593,107 @@ Person.query().where(ref('Model.jsonColumn:details'), '=', lit({ name: 'Jennifer
...
+  .findOne('raw sql')
+  .select('firstName')
+  .throwIfNotFound();
+const findOneWhereIn: Promise<Person | undefined> = Person.query()
+  .findOne('raw sql')
+  .whereIn('status', ['active', 'pending']);
+const findOneWhereInThrow: Promise<Person> = Person.query()
+  .findOne('raw sql')
+  .whereIn('status', ['active', 'pending'])
+  .throwIfNotFound();
```
koskimas: You cannot give a string for `findOne`. Also none of the methods accept raw sql as a string. Either `objection.raw` or `knex.raw` needs to be used.

typings/objection/index.d.ts
```diff
@@ -911,101 +860,101 @@ declare namespace Objection {
...
-    columns: Select<T>;
-    column: Select<T>;
-    from: Table<T>;
-    into: Table<T>;
-    table: Table<T>;
-    distinct: Distinct<T>;
+  type ColumnRef = string | Raw | Reference | QueryBuilder<any, any[]>;
+  type TableName = string | Raw | Reference | QueryBuilder<any, any[]>;
+
+  interface QueryInterface<QM, RM, RV> {
```
koskimas: I know `QueryInterface` was not added in this commit, but I have always wondered what't the purpose of it? Only one class ever implements it. Couldn't we just move these methods over to `QueryBuilder` and eliminate the whole interface?

tests/ts/examples.ts
koskimas: I don't know if it should be explicitly prevented, but having these examples is misleading.

tests/ts/examples.ts
```diff
@@ -584,6 +593,107 @@ Person.query().where(ref('Model.jsonColumn:details'), '=', lit({ name: 'Jennifer
...
+const deleteWhereIn: Promise<number> = Person.query()
+  .delete()
+  .whereIn('id', [1, 2, 3]);
+const deleteThrow: Promise<number> = Person.query()
+  .delete()
+  .throwIfNotFound();
+const deleteByIDThrow: Promise<number> = Person.query()
+  .deleteById(32)
+  .throwIfNotFound();
+
```
koskimas: Are there tests for these cases:

```js
// first turns array result into single
const x1: Promise<Model> = qb.where().first();
// the location of `first` doesn't matter
const x2: Promise<Model> = qb.first().where();
// returning turns `number` result back to `M[]`
const x3: Promise<Model[]> = qb.where().patch().returning();
// the position of `returning` doesn't matter (as long as it comes after the method that turns type into non-model)
const x4: Promise<Model> = qb.delete().returning().first().where();
```

tests/ts/examples.ts
jtlapp: I'll change the `findOne()` parameters to something valid. They don't matter for the purpose of this test, so might as well pick something that's possible.

tests/ts/examples.ts
jtlapp: I wondered about that. I'll see what I can do.

tests/ts/examples.ts
jtlapp: I'll check and add them if not. I want more `returning()` test cases as well. We put `returning()` on the base query builder, so it's always accessible.

typings/objection/index.d.ts
jtlapp: Are the QueryInterface methods all borrowed from Knex, while the QueryBuilder methods added by Objection? If so, it might help with maintenance to keep them separate. But then we might want to rename QueryInterface to make that clearer.

I do find the names QueryBuilder and QueryBuilderBase confusing. Both names suggest that they're the top-most query builder interface, but QueryBuilder extends QueryBuilderBase. At least in this new merging of query builders, every QueryBuilder instance implements QueryBuilderBase, so we could merge these two interfaces.

tests/ts/examples.ts
jtlapp: I'm noticing that my new examples are only testing the static methods. I wonder if I should duplicate them all for the instance methods.

mceachen: I'd tried to hit both in the past. If you could add them, that'd be great.

jtlapp: Maybe we don't need to dup many tests. After the first call to `Model.query()` or `model.$query()`, we're testing the same query builder methods. I'm thinking that we can provide coverage either way and then throw in a few token tests for other methods of getting query builders. The only internals we're taking advantage of is our knowing that they're all using the same query builder types.

Less code duplication makes for easier maintenance. And less judging! What do you think?

typings/objection/index.d.ts
mceachen: If you can pull it off and not break existing tests, go for it: less is more.

tests/ts/examples.ts
jtlapp: Okay, it seems that the purpose of these methods is to allow code to do an out-of-order prepending or appending of query actions to the query chain.

If the query chain were written expecting `runBefore` to change the result type that the first query method acts on, the type for the first method in the chain would be wrong, and the dev would have to typecast it. The typings only work if the final output of the `runBefore` series has the same type as the input to the first query method, which is now `Model | Model[]`.

I'm unable to find a way to incorporate `runAfter` result types. It would require adding a fourth type parameter to QueryBuilder and passing it from query builder to query builder through the typings. Call this type `AV`. We'd want `execute` to yield type `AV` when defined and `RV` otherwise, but AFAIK, Typescript cannot conditionally select among types. We could initialize `AV` to `never` so that `RV | never` evaluates to `RV`, but `RV | AV` would never evaluate to just `AV` because `RV` would never be `never`.

Unless there's a Typescript trick for doing this that I'm unaware of.

tests/ts/examples.ts
```diff
@@ -284,9 +284,23 @@ qb = qb.where(raw('random()', 1, '2'));
...
+  async (result: any, builder: objection.QueryBuilder<Person, Person[]>) => 88
+);
+
+const runAfterPerson: Promise<Person> = qb
+  .first()
+  .throwIfNotFound()
+  .runAfter(async (result: any, builder: objection.QueryBuilder<Person>) => 88);
+const runAfterPersons: Promise<Person[]> = qb.runAfter(
+  async (result: any, builder: objection.QueryBuilder<Person, Person[]>) => 88
+);
```
jtlapp: @koskimas, perhaps you can review these runBefore/runAfter tests? I'm mainly ensuring that these methods **do not** affect the query builder type chain, at least not in-place.

tests/ts/examples.ts
```diff
@@ -584,6 +607,132 @@ Person.query().where(ref('Model.jsonColumn:details'), '=', lit({ name: 'Jennifer
...
+    .throwIfNotFound();
+  const person = await staticQB;
+  staticQB = person.$query();
+
+  // Confirm that every query() type is a $query() type
+
+  let instanceQB = person.$query();
+  instanceQB = staticQB;
+}
+
```
jtlapp: @mceachen, I think I managed to avoid code duplication entirely, as well as make no assumptions about implementation. What do you think of this?

typings/objection/index.d.ts
jtlapp: I got answers. The QueryBuilder and QueryBuilderBase types may correspond to classes of those names in the implementation. If these types do indeed correspond to these classes, we may want to keep the names the same to facility maintenance.

OTH, as a person perusing just the public API -- including the typings -- I think I'd prefer calling QueryBuilderBase and QueryBuilder by the names of AbstractQueryBuilder and QueryBuilder, to tell me their roles and relationship at a glance.

Also, I was unable to merge these two interfaces because the return types of `throwIfNotFound()` cannot be generalized more strictly than `any`.

I did however move `returning()` and `runAfter()` to QueryBuilderBase. I think `runAfter()` may have originally been query-builder-specific in order to map the output of the query builder to the input of `runAfter()`. But because `runAfter()` callbacks can be chained, we can't hardcode its input type. I changed it to `any`.

jtlapp:
@mceachen, I experimented with defaulting query builders to using `Model[]`, and the only thing I don't have working is `$relatedQuery()`. I figured that's your baby, you'd want to trouble-shoot it.

I made a [`model-array-default` branch](https://github.com/jtlapp/objection.js/tree/model-array-default) off of the pending `pass-thru-typings` PR. For your convenience:

- [the typings](https://github.com/jtlapp/objection.js/blob/model-array-default/typings/objection/index.d.ts)
- [the tests](https://github.com/jtlapp/objection.js/blob/model-array-default/tests/ts/examples.ts)

mceachen:
@koskimas I'll have time to review this tomorrow, can you hold off on releasing 1.0.0 until tomorrow afternoon? It'd be nice if we got this refactoring into 1.0.0.

jtlapp:
Also, I may have a way to improve type support of model fields -- using partials in cases where models are currently returned whole but are actually partially returned, not using partials in cases where they are currently accepted when build-time-required model fields are actually required.

I might be able to do it in a backward-compatible way, but it's not entirely clear to me yet. 

It would be nice to hold off on 1.0 to let me continue working with the typings without feeling hamstrung for backward-compatibility. I'm racing as fast as I can.

cc @koskimas, @mceachen

mceachen:
@jtlapp I've been trying out this PR against my personal project, and the only issue I have is QueryBuilder's default of `RV=RM`, I really think it should be `RV=RM[]`. `.query()`, for example, returns all rows (and should, arguably, drive the default).

I also tried removing the `RM`, and rediscovered what you'd found that it was required to make `.returning` and like recover the return value, so I think that's defensible.

@koskimas, @jtlapp do you think it's reasonable to make QueryBuilder default return an array?

mceachen:
I'm also trying to flip the RM, RV generics so you don't have to do `<Person, Person, Person[]>` which makes me sad.

jtlapp:
@mceachen, I have all that fixed in the [model-array-default branch](https://github.com/jtlapp/objection.js/tree/model-array-default), per my above comment. The only thing not backward-compatible in that branch is `$relatedQuery()`, which I think is not quite working as you expect even in the current master.

jtlapp:
And yes, I agree that the default should be a model array, given that you have to go out of your way in the query builder chain to get anything else. The only drawback is that you have to do the unexpected parameterization `QueryBuilder<Model, Model>` to get a single model out -- it just looks redundant.

jtlapp:
If you want, I can push the model array default solution to this PR, so that we're evaluating that or no pass-through-result-types at all.

jtlapp:
I attempted to merge in @mceachen's mods but ended up screwing up the PR. #753 replaces this PR.

jtlapp:
@mceachen and I discussed supporting partials. It seems like a reasonable thing to do, but possibly non-trivial to do right. We're punting for 1.0. I think we're done with this investigation.

PR #749, opened 2018-01-22
Fix parsing relationships in case source data is already Model instance

### LINKED ISSUE #748: Model.fromJson is dropping relations from entities
Objection v 0.9.4

modelValidate.validate is removing all fields related to relationships here:
```
    // Strip away relations and other internal stuff.
    json = clone(json, true, true);
```

But then json that was generated this way is passed in modelSet to set relationships:

```
  if (!options.skipParseRelations) {
    parseRelationsIntoModelInstances(model, json, options);
  }
```

Even though at this point there is clearly no fields to create relationships from as those were removed during shallow/stripInternals cloning of the original JSON.

I presume the intention is to pass original, non-modified JSON to parseRelationsIntoModelInstances?

kibertoad:
fixes #748

### DIFF
--- a/lib/model/modelSet.js
+++ b/lib/model/modelSet.js
@@ -23,12 +23,13 @@ function setJson(model, json, options) {
   }
 
   json = model.$parseJson(json, options);
+  const originalJson = json;
   json = model.$validate(json, options);
 
   model.$set(json);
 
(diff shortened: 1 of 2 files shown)


koskimas:
Thanks!

Issue #717, opened 2018-01-11
QueryBuilder.relate() and .patch() yield the wrong type


jtlapp:
The typings have the `relate()` method of QueryBuilder returning type `this` and thus yielding a Model. However, `relate()` actually yields either a numberor an object pairing IDs to their corresponding reference fields. I don't know what the result is when providing multiple relations.

UPDATE: QueryBuilder `patch()` has the same problem -- should yield a number.

cc @mceachen

mceachen:
> relate() actually yields either a number (an ID) or an object pairing IDs to their corresponding reference fields

What leads you to that assertion? According to the [docs](http://vincit.github.io/objection.js/#relate), it should return a Promise to the number of related rows, not IDs.

@koskimas should `relate` return something like `QueryBuilderUpdate<this>` so the async return value is numeric?

jtlapp:
Ah, okay that may be correct about the number of related rows -- I was trying to infer that. The `express-ts` example returns the following, which is the only one I've tested:

```ts
{
  movieId: silverLinings.id,
  personId: bradley.id
}
```

jtlapp:
Yeah, `express-ts` returns the same thing. The above code is from my mocha test suite. Here's the raw output:

```
{"movieId":1,"personId":2}
```

jtlapp:
Is the output of `relate()` ever informative? In the above case, `movieId` and `personId` are actually handed to objection, which spits it back out.

jtlapp:
QueryBuilder `patch()` has the same problem, yielding a model type instead of a number (the number of modified rows).

koskimas:
@jtlapp `patch` typings do return a number. There's even this test in the test file for it:

```js
const rowsPatched: Promise<number> = qb.patch({});
```

You need the number to know if anything was updated when you do something like this:

```js
await Person.query().patch({age: 100}).where('age', '<', 100);
```

What kind of query throws if the where condition doesnät match any rows? Are you using `throwIfNotFound`?

koskimas:
`relate` return value depends on the actual operation. For `HasMany` and `BelongsToOne` relations `relate` is an update operation and the number of affected rows is returned. For `ManyToManyRelation` relate is an insert, and in that case the inserted model is returned (just like with `insert`). We could return a hardcoded `1` in that case too, but that would be less that useless.  By returning the inserted model, you can do this `.relate(foo).returning('*')` and get back whatever the database did for the inserted row (default values, generated values etc.).

[...]

Issue #501, opened 2017-08-31
Improved (database) errors


koskimas:
I would love to add a cross-db error API but as I have mentioned before it would be best implemented as a separate library. Then it could possibly be moved to knex at some point.

Currently errors thrown from db for example in case of foreign key constraint violation differ from db to db and are usually difficult to reason with. I would love to have something like:

```js
class DbError extends Error { ... }

class ConstraintViolationError extends DbError {
  get column() { ... }
  get table() { ... }
  get constraintName() { ... }
}

class ForeignKeyConstraintViolationError extends ConstraintViolationError {
  ...
}

// etc.
```
If someone is looking for an open source project, here's one that I think would benefit the whole node.js DB scene.

**EDIT** Check out  [db-errors](https://github.com/Vincit/db-errors) library I'm working on.
**EDIT 2** There's now [a db-errors plugin for objection](https://github.com/Vincit/objection-db-errors)

koskimas:
@fl0w suggested an approach where we would only add an unique key to the existing errors thus making this a non-breaking change.

The only problem I can think of is that the errors thrown by databases pretty much already contain all meaningful property names like code, message, type etc. At least postgres errors do, and that's enough. Our property would need to be something that doesn't conflict with any of the existing properties. Maybe we could use errorSymbol = Symbol() and something like

const { errorSymbol, errorCodes } = require('db-error');

function onError(error) {
  if (error[errorSymbol].code === errorCodes.ConstraintViolation) {
    console.log(error[errorSymbol].column)
    console.log(error[errorSymbol].constraintName)
  }
}

koskimas:
Actually the above approach has at least one severe downside: The custom information would not appear in the stack trace or error message. So uncaught errors would be as confusing as always.

capaj:
IMHO this issue should be placed in knex rather than here. This is a feature lacking on their side. They already normalise API for all the supported DBs, it should be doable to normalize errors as well.

BTW I'd definitely prefer the former approach of extending errors. Breaking changes are always a pain, but for our app we have a central place where we handle common SQL so it would be quite easy to refactor it.

elhigu:
Here is some discussion from knex about this error normalization:

https://github.com/tgriesser/knex/issues/2191
https://github.com/tgriesser/knex/issues/522 

I have also vague feeling that there was more discussion about it, but I couldn't find any other issues. 

[...]

PR #753, opened 2018-01-23
Pr 744 try 2

mceachen:
See PR #744

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -154,7 +154,12 @@ class Movie extends objection.Model {
 async () => {
   // Another example of strongly-typed $relatedQuery without a cast:
   takesPeople(await new Movie().$relatedQuery('actors'));
+
+  // If you need to do subsequent changes to $relatedQuery, though, you need
+  // to cast: :\
+  takesMaybePerson(await new Movie().$relatedQuery<Person>("actors").first())
 };
+
 const relatedPersons: Promise<Person[]> = new Person().$relatedQuery('children');
 const relatedMovies: Promise<Person[]> = new Movie().$relatedQuery('actors');
 
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -19,13 +19,14 @@ class Person extends objection.Model {
   children: Person[];
   pets: Animal[];
   comments: Comment[];
+  movies: Movie[];
 
   static columnNameMappers = objection.snakeCaseMappers();
 
   examplePersonMethod = (arg: string) => 1;
 
   petsWithId(petId: number): Promise<Animal[]> {
-    return this.$relatedQuery<Animal>('pets').where('id', petId);
+    return this.$relatedQuery('pets').where('id', petId);
   }
 
   fetchMom(): Promise<Person | undefined> {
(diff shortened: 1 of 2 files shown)


mceachen:
To recap: Joe refactored the QueryBuilder to capture the incoming and outgoing model types, so many methods on QueryBuilder that used to only be correct at the end of the call chain can now work in the middle of the chain.

Note that this *does contain some breaking changes*, in that a number of interfaces were collapsed into the main `QueryBuilder` interface, and code that may have compiled in the past needs explicit casting or will expose prior defects. I've tried this branch against my project (which uses ~75% of the objection API surface area), and it did require a couple updates, but the changes to my project were minimal.

jtlapp:
Oh, don't forget that this resolves #742 and partially resolves #717. I never looked at how to accommodate the varying `relate()` output types of #717.

mceachen:
@koskimas I've run this with the new tests from #735 and everything passes. Are you OK merging this?

koskimas:
Looks good to me! Great job with this 👍

jtlapp:
Just FYI, these new typings break my [rigging for Typescript support of custom query builders](https://github.com/jtlapp/temp-qb). The rigging is not viable for publication, but the break does send me off looking for another way to get custom query builders working. The problem appears to have to do with defaulting the result type to `Model[]`. I don't know if the fix is in my rigging or in the way the typings handle the defaulting.

Investigating...

elhigu:
In my current project these lines started failing: 

```
throw new ValidationError({docType: 'Invalid document type'});

file: 'file:///Users/mikaelle/Projects/Vincit/xxx/index.ts'
severity: 'Error'
message: 'Argument of type '{ name: string; }' is not assignable to parameter of type 'CreateValidationErrorArgs'.
  Object literal may only specify known properties, and 'name' does not exist in type 'CreateValidationErrorArgs'.'
at: '92,34'
source: 'ts'
code: '2345'
```

jtlapp:
I solved my problem in the custom query implementation. Each query builder class must specify the default model array. E.g.:

```ts
export class SessionQueryBuilder<QM, RM=QM[], RV=RM> extends QueryBuilder<QM, RM, RV> {
  session(session: Session) {
    return this.mergeContext({
      session: session
    });
  }
}
```

So all is good on that front. I'll still look at how to get support for custom query builders into the typings proper.

jtlapp:
>In my current project these lines started failing:
>
>throw new ValidationError({docType: 'Invalid document type'});

@elhigu, I'm thinking that the relevant mods are from the work @koskimas did on #501.

koskimas:
@elhigu Those are 1.0 breaking changes.

koskimas:
I just realized that the new validation error types don't allow custom error types or even a `Generic` or `Custom` type to use for own validation errors. Maybe something like that should be added. It's not related to this PR. I'll create another one if I come up with a solution.

jtlapp:
@mceachen, transactions are yielding `Promise<any>`. Not fully understanding how they are supposed to behave, it's hard for me to think about how we might do better than this.

jtlapp:
Type inference works fine with `Objection.transaction()`, but there's still something odd.

The typings declare a transaction to be the following but never uses the `T` type:

```ts
  export interface transaction<T> {
```

Here's the only place the decl is used, on the Objection namespace, apparently having no effect:

```ts
  export const transaction: transaction<any>;
```

This state of affairs also exists in the typings prior to the result type pass-through mods.

(I may have mistakenly interpreted this as meaning that transactions only return type `any`.)

mceachen:
@jtlapp let's not conflate other typings with this PR. I've merged this just now, and we can talk transactions over hangouts and determine what we should do there.

PR #758, opened 2018-01-24
Write docs and tests for QueryBuilder.has()

lehni:
@koskimas I forgot to add this in #736

I hope the test with a random list of operators is good enough?

### DIFF
--- a/doc/includes/API.md
+++ b/doc/includes/API.md
@@ -3000,6 +3000,33 @@ boolean|true if the query defines any eager expressions.
 
 
 
+#### has
+
+```js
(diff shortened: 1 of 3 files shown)


koskimas:
Perfect, thanks! By the way, its perfectly ok to use `.forEach` and other array functions nowadays if you prefer them. I've been using plain for loops in the hot path (and the habbit has leaked outside the hot path) because older nodes had some performance problems with those functions and closures.

lehni:
Good to know! I'll keep that in mind for future tests.

PR #764, opened 2018-01-27
Minor simplifications in eager() and mergeEager()  and mergeAllowEager()

lehni:
Looking through the eager code, I found this potential for some minor code stream-lining / simplifications.

### DIFF
--- a/lib/queryBuilder/QueryBuilder.js
+++ b/lib/queryBuilder/QueryBuilder.js
@@ -229,15 +229,7 @@ class QueryBuilder extends QueryBuilderBase {
   }
 
   eager(exp, filters) {
-    this._eagerExpression = exp || null;
-
-    if (isString(this._eagerExpression)) {
-      this._eagerExpression = parseRelationExpression(this._modelClass, this._eagerExpression);
-    }
-
-    if (filters) {
-      this._eagerExpression.filters = filters;
-    }
+    this._eagerExpression = parseRelationExpression(this._modelClass, exp, filters);
 
     checkEager(this);
     return this;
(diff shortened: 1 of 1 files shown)


lib/queryBuilder/QueryBuilder.js
```diff
@@ -229,15 +229,7 @@ class QueryBuilder extends QueryBuilderBase {
...
-    this._eagerExpression = exp || null;
-
-    if (isString(this._eagerExpression)) {
-      this._eagerExpression = parseRelationExpression(this._modelClass, this._eagerExpression);
-    }
-
-    if (filters) {
-      this._eagerExpression.filters = filters;
-    }
+    this._eagerExpression = parseRelationExpression(this._modelClass, exp, filters);
```
lehni: @koskimas the `isString()` check here isn't really necessary because `RelationExpression.parse()` already does the same.

newhouse:
@lehni I'm on the go right now so can't dig in, but can you see if there's anything in the `allowEager` and `mergeAllowEager` areas that could  benefit from this same change? It may already waterfall down to them, but I can't look now. I just know that I updated some code in the `mergeAllowEager` method that was related to this stuff a while back and thought it was worth a look to see if they could/would also benefit from this.

lehni:
@newhouse yes you're right indeed! I can add this to the PR. Thanks for bringing it to my attention!

lehni:
@newhouse 5fad00d172065e851181a354152b6f961b91554a includes that change now. Thanks!

koskimas:
Thanks!

lib/queryBuilder/QueryBuilder.js
```diff
@@ -1096,9 +1078,13 @@ Object.defineProperties(QueryBuilder.prototype, {
...
 });
 
-function parseRelationExpression(modelClass, expr) {
+function parseRelationExpression(modelClass, exp, filters) {
   try {
-    return RelationExpression.parse(expr);
+    const expr = RelationExpression.parse(exp);
+    if (filters) {
+      expr.filters = filters;
+    }
```
lehni: @koskimas we could consider moving this part to `RelationExpression.parse()` with a second filters argument. Maybe it's better to do the setting of `filters` there than here, since it's a property of `RelationExpression`

lehni: @koskimas oops, this came after the merge... But I think it's worth considering.

PR #767, opened 2018-01-29
RelatedTableAlias: generate correct aliases for tables with schema

### LINKED ISSUE #766: LeftJoin incorrectly processes tables with schema
objection/lib/relations/Relation.js:

```
  relatedTableAlias(builder) {
    const tableRef = builder.tableRefFor(this.relatedModelClass);
    return `${tableRef}_rel_${this.name}`;
  }
```

This code generates invalid (at least in PostgreSQL) alias in case tableRef includes schema part (e. g. 'app.users').

Here is the fixed variant:

```
  relatedTableAlias(builder) {
    const tableRef = builder.tableRefFor(this.relatedModelClass);
    const tableRefNoSchema = /[^.]*$/.exec(tableRef)[0];
    return `${tableRefNoSchema}_rel_${this.name}`;
  }
```

igor-savin-ht:
fixes #766

### DIFF
--- a/lib/relations/Relation.js
+++ b/lib/relations/Relation.js
@@ -79,8 +79,9 @@ class Relation {
 
   relatedTableAlias(builder) {
     const tableRef = builder.tableRefFor(this.relatedModelClass);
+    const tableRefWithoutSchema = tableRef.replace('.', '_');
 
-    return `${tableRef}_rel_${this.name}`;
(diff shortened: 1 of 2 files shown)


igor-savin-ht:
@koskimas Comments on whether it is better to replace '.' with '_' or drop the prefix entirely are most welcome.

PR #773, opened 2018-01-31
Fixes #772

### LINKED ISSUE #772: typings don't work with typescript 2.7
This is the error:

```
tests/ts/examples.ts(41,3): error TS2564: Property 'firstName' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(42,3): error TS2564: Property 'lastName' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(43,3): error TS2564: Property 'mom' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(44,3): error TS2564: Property 'children' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(45,3): error TS2564: Property 'pets' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(46,3): error TS2564: Property 'comments' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(47,3): error TS2564: Property 'movies' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(163,3): error TS2564: Property 'title' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(164,3): error TS2564: Property 'actors' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(165,3): error TS2564: Property 'director' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(214,3): error TS2564: Property 'species' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(229,3): error TS2564: Property 'comment' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(271,3): error TS2564: Property 'canAct' has no initializer and is not definitely assigned in the constructor.
tests/ts/examples.ts(292,72): error TS2345: Argument of type 'typeof Person' is not assignable to parameter of type 'ModelClass<Model>'.
  Type 'Person' is not assignable to type 'Model'.
    Types of property '$setJson' are incompatible.
      Type '(json: Pojo, opt?: ModelOptions | undefined) => Person' is not assignable to type '(json: Pojo, opt?: ModelOptions | undefined) => Model'.
        Type 'Person' is not assignable to type 'Model'.
```

@mceachen Could you take a look at this at some point?

mceachen:
* Updated all dependencies (except knex)
* Fixed optional/required model fields (TS 2.7 --strict support)
* TS 2.7 is more strict with genericized return values, but doesn't infer model types still, so many `<M>(this: M)` captures had to be used in lieu of prior `this`-returns.
* QM now always extends Model. I believe this is correct, right? Can you query on non-models?

### DIFF
--- a/examples/express-es6/package.json
+++ b/examples/express-es6/package.json
@@ -17,10 +17,10 @@
     "body-parser": "^1.18.2",
     "express": "^4.16.2",
     "express-promise-router": "^3.0.1",
-    "knex": "^0.14.2",
+    "knex": "^0.13.",
     "lodash": "^4.17.4",
     "morgan": "^1.9.0",
-    "objection": "^0.9.4",
+    "objection": "^1.0.0-rc.11",
     "sqlite3": "^3.1.13"
   }
 }
--- a/examples/express-es6/package.json
+++ b/examples/express-es6/package.json
@@ -13,14 +13,14 @@
   "author": "Sami Koskimäki",
   "license": "MIT",
   "dependencies": {
-    "axios": "^0.17.0",
+    "axios": "^0.17.1",
     "body-parser": "^1.18.2",
     "express": "^4.16.2",
-    "express-promise-router": "^2.0.0",
-    "knex": "^0.13.0",
+    "express-promise-router": "^3.0.1",
+    "knex": "^0.14.2",
     "lodash": "^4.17.4",
     "morgan": "^1.9.0",
-    "objection": "^0.9.0",
+    "objection": "^0.9.4",
     "sqlite3": "^3.1.13"
   }
 }
--- a/examples/express-es7/package.json
+++ b/examples/express-es7/package.json
@@ -24,14 +24,14 @@
   "author": "Sami Koskimäki",
   "license": "MIT",
   "dependencies": {
-    "axios": "^0.17.0",
+    "axios": "^0.17.1",
     "body-parser": "^1.18.2",
     "express": "^4.16.2",
-    "express-promise-router": "^2.0.0",
-    "knex": "^0.13.0",
+    "express-promise-router": "^3.0.1",
+    "knex": "^0.14.2",
     "lodash": "^4.17.4",
     "morgan": "^1.9.0",
-    "objection": "^0.9.0",
+    "objection": "^0.9.4",
     "sqlite3": "^3.1.13"
   },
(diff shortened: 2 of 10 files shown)


mceachen:
`name?: string` is a synonym for `(string | undefined)`.

`name: string` means the field will be set at construction time, and never
be undefined.

`name!: string` means consumers can expect the field to always be set by
the constructor, but that the compiler may not be able to determine that is
happening, and to not fall compilation if that happens.

mceachen:
`name?: string` is a synonym for `(string | undefined)`.

`name: string` means the field will be set at construction time, and never be undefined.

`name!: string` means consumers can expect the field to always be set by the constructor, but that the compiler may not be able to determine that is happening, and to not fall compilation if that happens.

examples/express-es6/package.json
```diff
@@ -13,14 +13,14 @@
...
   "license": "MIT",
   "dependencies": {
-    "axios": "^0.17.0",
+    "axios": "^0.17.1",
     "body-parser": "^1.18.2",
     "express": "^4.16.2",
-    "express-promise-router": "^2.0.0",
-    "knex": "^0.13.0",
+    "express-promise-router": "^3.0.1",
+    "knex": "^0.14.2",
```
igor-savin-ht: knex 0.14.2 has known problems with memory leaks, I wouldn't recommend updating to it before 0.14.3 is out.

examples/express-es7/package.json
```diff
@@ -24,14 +24,14 @@
...
     "body-parser": "^1.18.2",
     "express": "^4.16.2",
-    "express-promise-router": "^2.0.0",
-    "knex": "^0.13.0",
+    "express-promise-router": "^3.0.1",
+    "knex": "^0.14.2",
     "lodash": "^4.17.4",
     "morgan": "^1.9.0",
-    "objection": "^0.9.0",
+    "objection": "^0.9.4",
```
igor-savin-ht: why not use 1.0.0-rc.10 as elsewhere?

mceachen: 👍

examples/express-es6/package.json
mceachen: 👍

PR #784, opened 2018-02-06
Write docs and tests for QueryBuilder.clear()

lehni:
Similar to #758, here the docs and tests for QueryBuilder.clear()

### DIFF
--- a/doc/includes/API.md
+++ b/doc/includes/API.md
@@ -3027,6 +3027,33 @@ boolean|true if the query defines an operation that matches the given selector.
 
 
 
+#### clear
+
+```js
+queryBuilder.clear(selector);
+```
+
+```js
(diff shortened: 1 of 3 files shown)


kapouer:
Everybody's going to ask "does it clear select as well" ?

koskimas:
@kapouer I think you misunderstood what the method does.

```js
// Clears all calls to `where` method.
builder.clear('where')
// Clears all calls to `select` method.
builder.clear('select')
// Clears all calls to `where`, `orWhere`, `andWhere` etc. method.
builder.clear(/where/)
```

koskimas:
I think the documentation is clear on this.

lehni:
I could add an example to illustrate the use of a regular expression though?

PR #795, opened 2018-02-12
Do not fetch relations set to undefined in data

lehni:
When a value for a relation is set to `undefined` in data passed to `insertGraphAndFetch()` or `upsertGraphAndFetch()`, this relation should not be fetched after the insert / upsert operation.

A simple unit test is included to check for the correct behavior.


koskimas:
I'm currently rewriting this whole file. I'll include this fix in it.

lehni:
@koskimas I figured as much, but thought I could help with a unit test also.

PR #796, opened 2018-02-13
Doc: Add missing arg in runBefore() / runAfter()

lehni:
I found this omission today while playing around with the different query hooks. After looking into all occurrences and also the way these methods are being called, I'm fairly certain that this is their correct signature.

### DIFF
--- a/doc/includes/API.md
+++ b/doc/includes/API.md
@@ -2736,8 +2736,8 @@ const context = builder.context();
 Person
   .query()
   .context({
-    runBefore: (builder) => {},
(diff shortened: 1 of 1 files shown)


koskimas:
Did you test that its really like this?

lehni:
Yes I did!

lehni:
See here for the reason why:

https://github.com/Vincit/objection.js/blob/4db6fad8671c321e5317c57c8f6022ece964c02d/lib/queryBuilder/QueryBuilder.js#L1319

PR #798, opened 2018-02-15
Improve docs for runBefore() / runAfter()

lehni:
@koskimas I should have noticed this when creating #796, but here a follow-up PR that improves docs further.

We could consider making these `return result;` statements optional with a simple change to this line:

https://github.com/Vincit/objection.js/blob/4db6fad8671c321e5317c57c8f6022ece964c02d/lib/queryBuilder/QueryBuilder.js#L1319

->

```js
promise = promise.then(result => {
  const res = func.call(builder, result, builder);
  return res !== undefined ? res : result;
});
```

### DIFF
--- a/doc/includes/API.md
+++ b/doc/includes/API.md
@@ -2736,9 +2736,13 @@ const context = builder.context();
 Person
   .query()
   .context({
-    runBefore: (result, builder) => {},
-    runAfter: (result, builder) => {},
-    onBuild: (builder) => {}
+    runBefore: (result, builder) => {
+      return result;
+    },
+    runAfter: (result, builder) => {
+      return result;
+    },
+    onBuild: builder => {}
   });
 ```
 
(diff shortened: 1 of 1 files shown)


lehni:
Wait, that's not the only place that would need changing. These two would need similar changes:

https://github.com/Vincit/objection.js/blob/74a0d6f5189eba9e00ec8f4709e7c7f772c234d1/lib/queryBuilder/operations/RunBeforeOperation.js#L10
https://github.com/Vincit/objection.js/blob/74a0d6f5189eba9e00ec8f4709e7c7f772c234d1/lib/queryBuilder/operations/RunAfterOperation.js#L10

lehni:
@koskimas PS: If you think the above suggestion makes sense, I'm happy to create a PR and some tests for it.

koskimas:
Thanks! Let's not add the undefined handling you suggested. You would need to handle the case where undefined promise is returned too and that would complicate things:

```js
runAfter(async () => {
  await something();
  // No return value --> a promise resolving to `undefined` gets returned.
})
```

lehni:
That's a good point! I agree, let's keep it simple.

PR #799, opened 2018-02-15
Make joinModelClass public

lehni:
Update: This issues first was only about making `joinModelClass(knex)` work on relation relations that don't have any join models defined, but turned into a broader discussion about opening up access to the method and making it public.


lehni:
Looks like a random fail in Node 9...

Also, I just noticed that `joinModelClass()` is private. Could it be added to the public API? I have use for it, e.g. for cleaning join tables in unit tests.

koskimas:
I don't think joinModelClass function should be public. You need to pass knex instance to it, which is not the nicest API. You can use raw knex to clear the join table.

lehni:
The nice thing about being able to access `joinModelClass()` is that I don't have to hard-code the table names, and can have things operate on a meta-level. There is real value in that...

What if the private `joinModelClass(knex)` was renamed to `getJoinModelClass(knex)`, and we introduce a new, public `get joinModelClass()` accessor which returns `this.getJoinModelClass(this.ownerModel.knex())`? I think this would be generally useful, without causing harm.

koskimas:
> `this.getJoinModelClass(this.ownerModelClass.knex())`

👍 seems like a good idea. I wonder if there was a reason I didn't do it that way to begin with 🤔 
Well, if the tests pass, then I'm okay with that.

lehni:
Great to hear! I'm happy to work on the PR for it. Do you mean it could even be used internally instead of passing knex around?

koskimas:
Maybe if the tests work. I'm pretty sure it was a failing test that made me add the knex as an argument. I can't remember why it's needed. It has something to do with the fact that if the junction table model is not provided, it is generated.

lehni:
Ok, I will look into all this shortly!

lehni:
@koskimas good news: All tests seem to pass!

lib/relations/manyToMany/ManyToManyRelation.js
```diff
@@ -52,7 +52,7 @@ class ManyToManyRelation extends Relation {
     ctx = finalizeJoinModelClass(ctx);
 
     this.joinTableExtras = ctx.joinTableExtras;
-    this.joinTableModelClass = ctx.joinModelClass;
+    this.joinTableModelClass = ctx.joinTableModelClass;
```
lehni: @koskimas I renamed `ctx.joinModelClass` to `ctx.joinTableModelClass` so that is symmetric with the fields on `Relation`, and also so that it's easier to find actual uses of `Relation.joinModelClass`

lehni:
Let's see if this passes all tests on Travis. If this looks good to you,  I shall add some docs to it as well.

lehni:
All green! 🎉

lib/relations/Relation.js
```diff
@@ -70,12 +70,14 @@ class Relation {
...
-  joinModelClass(knex) {
-    if (knex && knex !== this.joinTableModelClass.knex()) {
-      return this.joinTableModelClass.bindKnex(knex);
-    } else {
-      return this.joinTableModelClass;
+  get joinModelClass() {
+    if (this.joinTableModelClass) {
+      const knex = this.ownerModelClass.knex();
+      if (knex !== this.joinTableModelClass.knex()) {
+        return this.joinTableModelClass.bindKnex(knex);
```
lehni: @koskimas couldn't this simply be `this.joinTableModelClass.knex(knex)` now instead?

lib/relations/Relation.js
```diff
@@ -74,7 +74,7 @@ class Relation {
     if (this.joinTableModelClass) {
       const knex = this.ownerModelClass.knex();
       if (knex !== this.joinTableModelClass.knex()) {
-        return this.joinTableModelClass.bindKnex(knex);
+        this.joinTableModelClass.knex(knex);
```
lehni: @koskimas this works indeed! All tests should still pass.

koskimas: This is a huge no-no! Think about parallel transactions and what will happen with them.

lehni: Oh ok, yeah I hadn't thought about that! 😬 I'll revert the commit then, makes a lot of sense. Surprised that no unit test caught it?

lehni:
Hmm is this another random fail? It only happens on Node 6, but is somehow related to Postgresql

lehni:
I've added docs to it now, and it all seems to pass again.

koskimas:
LGTM! Thanks

PR #800, opened 2018-02-15
typos in docs

burningTyger:
not sure why the last line is changed??


koskimas:
Functions can be virtual attributes too. The function example is there for a reason.

Issue #360, opened 2017-04-18
QueryBuilder.toString() and QueryBuilder.toSql() throw TypeError when using JoinEagerAlgorithm


pvatterott:
In objection 0.7.9, when calling `QueryBuilder.toString()` or `QueryBuilder.toSQL()`, objection will throw an exception from `RelationJoinBuilder.js`.

The following code snippet should be a minimal repro:
```javascript
try {
  const query = MyModel.query().eager('relation').eagerAlgorithm(MyModel.JoinEagerAlgorithm);
  console.log(query.toString()); // or toSQL();
} catch (err) {
  console.error(err);
}
```
throws the following exception:
```javascript
TypeError: Cannot read property 'columns' of undefined
        at RelationJoinBuilder.buildSelects (.../packages/models/node_modules/objection/lib/queryBuilder/operations/eager/RelationJoinBuilder.js:392:37)
        at RelationJoinBuilder.doBuild (.../packages/models/node_modules/objection/lib/queryBuilder/operations/eager/RelationJoinBuilder.js:293:12)
        at RelationJoinBuilder.build (...packages/models/node_modules/objection/lib/queryBuilder/operations/eager/RelationJoinBuilder.js:140:10)
        at JoinEagerOperation.onBeforeBuild (.../packages/models/node_modules/objection/lib/queryBuilder/operations/eager/JoinEagerOperation.js:81:22)
        at QueryBuilder.buildInto (.../packages/models/node_modules/objection/lib/queryBuilder/QueryBuilderOperationSupport.js:315:10)
        at _build (.../packages/models/node_modules/objection/lib/queryBuilder/QueryBuilder.js:1407:25)
        at QueryBuilder.build (.../packages/models/node_modules/objection/lib/queryBuilder/QueryBuilder.js:734:23)
        at QueryBuilder.toString (.../packages/models/node_modules/objection/lib/queryBuilder/QueryBuilder.js:536:17)
```

koskimas:
Whoops! Actually there is no way to return anything useful from `toString` of a join eager query since the table information (all columns and types) need to be fetched first. `toString` will work after the first query thought. I need to add a better error message there.

You can chain `.debug()` method to the query to get the sql printed out. Will that work?

pvatterott:
[...]

PR #817, opened 2018-02-23
Clarity for QueryBuilder.toString() vs .toSql() behavior

bw:
## Discussion and context

From looking at #360, I can see why [toString() was wrapped in a try-catch](https://github.com/Vincit/objection.js/commit/17ee8f1489a51ea8ea9681bacdceb5933a646c0d). In my personal opinion, I think that try-catching all exceptions is dangerous. Silently returning a phrase instead of the expected SQL is even more sketchy, but reading the thread in #360 makes it clear that most people are using toString() for logging purposes. I guess that's okay.

The problem is that `toSql()` is a direct alias of `toString()`, complete with catching every error. We use `toSql()` to pipe into a alternate database resolution layer. So in instances where we want an exception to be thrown, we must instead literally look for the phrase "This query cannot be built synchronously" to determine if an error has occurred.

## Proposed changes

This PR leaves the behavior of `toString()` the same, but will allow `toSql()` to raise exceptions as appropriate. To be more specific, it makes `toSql()` the primary method and `toString()` a method that wraps `toSql()` in a try-catch.

So the behavior is exactly the same for toString() and no changes need to be made to the code.

Without this PR, we are forced to  use `this.build().toString()` directly.

### DIFF
--- a/lib/queryBuilder/QueryBuilder.js
+++ b/lib/queryBuilder/QueryBuilder.js
@@ -435,14 +435,14 @@ class QueryBuilder extends QueryBuilderBase {
 
   toString() {
     try {
-      return this.build().toString();
+      return this.toSql();
     } catch (err) {
       return `This query cannot be built synchronously. Consider using debug() method instead.`;
     }
   }
 
   toSql() {
-    return this.toString();
+    return this.build().toString();
   }
 
   clone() {


koskimas:
You should never execute the SQL returned by toSql or toString. It interpolates all bindings into the string and skips escaping. This is also a breaking change and I cannot merge this before 2.0. I'd say you should find another way to do what you are trying to do. Maybe we could add a method that returns the template SQL with bindings array, like knex's toSQL?

bw:
@koskimas that's not what [the documentation](http://vincit.github.io/objection.js/#tosql) seems to suggest.

What we need is a way to return the final raw query so we can run it through our own system (which handles things like batching, etc). What's the best way to do that?

(Also-- in what way is it a breaking change? I thought the whole point of having both `toSql()` and `toString()` was to differentiate the behavior. It doesn't make sense that, in 1.0, `toSql()` maps to `toString()`.)

Thanks!

elhigu:
@bw The breaking change in this is that in 1.0 `toSql()` and `toString()` has been meant to be aliases to each other. After this change they are not functioning the same way anymore and documentation would need to be changed to mention that other one of the methods is now throwing an exception in some cases.

In knex there is `toSQL()` which returns sql with `?` style parameter bindings and `toSQL().toNative()` which returns SQL that uses native bindings (for example with postgresql $1, $2... etc.). Examples found here http://knexjs.org/#Interfaces-toSQL

If objection could provide API for returning knex query builder / throwing an exception if multiple queries are needed to run the objection builder, in that case knex's `toSQL().toNative()` could be used.

bw:
@elhigu OK, I understand now why this would be a breaking change. We can introduce a new method that might solve this problem better in the next release. (We'd be happy to help.)

In the meantime, it sounds like we should be using `build().toSQL().toNative()`? Can you advise on this?
What we are doing is executing the output on Postgres via another layer (complicated circumstances; there are a few other things wrapping it). So what we need is a valid, escaped, SQL as a string we can use elsewhere.

I would also suggest that the documentation should be updated to reflect the actual behavior of `toSQL()` and `toString()`. This might help prevent some trip-ups-- it took us quite some time to realize that `toSQL()` was literally returning the string "This query cannot be built synchronously...".

----

With the squad's approval (on a separate PR), I would like to update the docs for both methods from:

> Returns the SQL string. If this query builder executes multiple queries, only the first query's SQL is returned.

to the following (note also that the above description is incorrect; executing multiple queries _won't return only the first query's SQL_. It will just return a hardcoded string instead):

> **toSql()**: Returns the SQL string, via Knex as `this.build().toString()`. This may be useful for debugging or logging, but should not be used to create queries for database execution. In particular, bindings may not be escaped properly.
>
> In the current release, if the query builder attempts to execute multiple queries or throw any exception, **no error will throw** and instead the following string is returned:
>
> ```
> This query cannot be built synchronously. Consider using debug() method instead.
> ```
>
> Later versions of Objection may introduce a native way to retrieve an executable SQL statement, or handle this behavior differently. If you need executable SQL, consider `this.build().toSQL().toNative()`, which can [provide formatted bindings](http://knexjs.org/#Interfaces-toSQL).
>
> ----
>
> **toString()**: An alias for `toSql()`. In later versions of Objection, `toSql()`'s behavior may change to throw an exception for query builders that execute multiple queries, but `toString()` will ensure a valid string output for use in debug or logging.

koskimas:
>  that's not what the documentation seems to suggest.

The documentation could be clearer about this, but I thought it was obvious that an SQL string without a separate bindings should never be executed. Isn't this SQL injection 101? I'll add a note about this to the docs.

`toString` and `toSql` are aliases and that's pretty stupid considering they do different things in knex. We should probably deprecate `toSql`, make it clear that `toString` is only for debugging and add a `toNativeSql` (name can be debated) method that returns what `build().toSQL().toNative()` returns. `toNativeSql` will throw in cases where the query requires asynchronous actions (like `joinEager` method) but that could also be fixed by exposing the `Model.fetchDbMetadata` function that could be called during app initialization. After that function is called, queries like `joinEager` no longer require async stuff. The exception thrown from `toNativeSql` should make this fix clear to the user.

bw:
@koskimas: to be honest, I assumed that, as long as the proper query builder methods were used, the final result call to `toSql()` would be escaped. Again, this was all operating on the assumption that `toSql()` was rendering a SQL statement that could be executed.

bw:
As the proposed changes in this PR are not to be implemented, closing this. A documentation-update PR has been made as #819.

PR #827, opened 2018-02-28
combine strict compiler options

bennyn:
The "strict" option already enables `--noImplicitAny`, `--noImplicitThis`, `--alwaysStrict`, `--strictNullChecks`, `--strictFunctionTypes` and `--strictPropertyInitialization`, so there is no need to list them individually.

Reference: https://www.typescriptlang.org/docs/handbook/compiler-options.html

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,18 +1,13 @@
 {
   "compilerOptions": {
     "allowJs": false,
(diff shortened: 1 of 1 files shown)


koskimas:
Thank you!

PR #835, opened 2018-03-09
Remove unused variable in transaction docs

vladshcherbin:


koskimas:
It's there on purpose to show that whatever you return inside the transaction, is the return value of the transaction function.

PR #836, opened 2018-03-09
Fix typings for loadRelated static method

chamini2:


koskimas:
Thank you!

chamini2:
When could we get a release?

PR #858, opened 2018-03-30
ensureRelation returns null if relation is string

timhuff:
I haven't run the test suite on this change but `$setRelated` kept erroring for me until I made this change. I feel it must be a bug, though, because the way it was written, line 171 doesn't do anything. Is there not a test case for `$setRelated('name', model)`?

### DIFF
--- a/lib/model/modelSet.js
+++ b/lib/model/modelSet.js
@@ -169,9 +169,8 @@ function appendRelated(model, relation, models) {
 function ensureRelation(model, relation) {
   if (typeof relation === 'string') {
     relation = model.constructor.getRelation(relation);
-  } else {
-    return relation;
   }
+  return relation;
 }
 
 module.exports = {


koskimas:
Wow, nice catch. Apparently there is no test for this case. I'll add one. I'll merge this as soon as the tests pass.

timhuff:
I'm surprised this hasn't come up yet. Looks like it's causing [this test case](https://github.com/Vincit/objection.js/blob/master/tests/integration/misc/%23782.js) to fail, though.

koskimas:
That was a random fail. I just fixed it.

Yeah, super weird this hasn't come up.

timhuff:
yea, that's what I figured, too. When I ran the tests manually, it worked. I made a new commit to trigger a new run with Travis CI.

koskimas:
It works now. I'll merge this as you remove the dummy commit.

timhuff:
Should be good to go. Thanks!

koskimas:
Awesome, thanks! I'll write a test case and release a new patch version today.

PR #863, opened 2018-04-01
Add `name` property to custom errors

yenbekbay:
If you access the `stack` property of an error, you will get a string with the first line [formatted as `<error class name>: <error message>`](https://nodejs.org/api/errors.html#errors_error_stack). Here, `<error class name>` refers to the `name` property of the error, which by default is just `Error`.

Currently, whenever you try to print the stack of a custom objection.js error, you get the string that starts with `Error: `. I think it would be preferable to get the actual name of the error instead of just `Error`.


koskimas:
Good idea, thanks!

Issue #869, opened 2018-04-06
Cannot use 'in' operator to search for 'key' in undefined


igor-savin-ht:
EDIT BY @koskimas: This has been fixed in 1.1.3

Seems to be a regression in 1.1.2.

When calling JSON.stringify(entities); (which, e. g., happens when you are trying to send an array of Objection entities as a response from Express), an error is thrown.
Error seems to only manifest itself when entity in question has custom idColumn defined (in this particular case it was defined as
```
	static get idColumn() {
		return 'key';
	}
```
)

Error stacktrace:

TypeError: Cannot use 'in' operator to search for 'key' in undefined
    at assignJsonValue (C:\sources\lms-api\models\node_modules\objection\lib\model\modelToJson.js:79:33)
    at toExternalJsonImpl (C:\sources\lms-api\models\node_modules\objection\lib\model\modelToJson.js:51:5)
    at toJson (C:\sources\lms-api\models\node_modules\objection\lib\model\modelToJson.js:20:14)
    at Brand.$toJson (C:\sources\lms-api\models\node_modules\objection\lib\model\Model.js:204:12)
    at Brand.toJSON (C:\sources\lms-api\models\node_modules\objection\lib\model\Model.js:208:17)
    at JSON.stringify (<anonymous>)
    at eval (eval at stringify (C:\sources\lms-api\api\node_modules\express\lib\response.js:1117:14), <anonymous>:1:6)
    at stringify (C:\sources\lms-api\api\node_modules\express\lib\response.js:1117:14)
    at ServerResponse.json (C:\sources\lms-api\api\node_modules\express\lib\response.js:260:14)
    at ServerResponse.send (C:\sources\lms-api\api\node_modules\express\lib\response.js:158:21)
    at router.get (C:\sources\lms-api\api\controllers\brands.controller.js:44:7)
    at <anonymous>

Surprisingly, this works fine when single entity is being stringified.

koskimas:
I'll need a repro for this. I was not able to reproduce this.

igor-savin-ht:
@koskimas Will create a failing test ASAP

koskimas:
Silly me, this throws:

```js
const { Model } = require('./');

class Person extends Model {
  static get idColumn() {
    return 'key';
  }
}

const p1 = Person.fromJson({ key: 1 });
const p2 = Person.fromJson({ key: 2 });

JSON.stringify([p1, p2]);
```

@igor-savin-ht No need for repro anymore.

[...]

PR #870, opened 2018-04-06
Test to reproduce #869

kibertoad:
refs #869


kibertoad:
already reproduced and fixed

koskimas:
Thanks, but this is already fixed in 1.1.3

PR #882, opened 2018-04-13
 Types for tableMetadata and fetchTableMetadata 

pvatterott:
I'm not super confident in these, especially `FetchTableMetadataOptions.parentBuilder`, but this attempts to add types for the new functions exposed in ff9af5d by @koskimas

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -237,6 +237,21 @@ declare namespace Objection {
     [key: string]: any;
   }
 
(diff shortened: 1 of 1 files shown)


koskimas:
`parentBuilder` is private and shouldn't be in the typings. Other than that, looks good 👍

pvatterott:
Removed `parentBuilder` in 80e2568b6f40ee6bdb88368e65b251618ca0346d

PR #889, opened 2018-04-25
Adding beforeInsert to RelationMapping interface

hardeep:
Hi I noticed that the beforeInsert method defined in the documentation is missing from the typings file. This pull request will fix that.

Apologies for not creating an issue before sending the pull request. It was a blocker for me, so I needed to fix it for my own needs.


koskimas:
Nice catch, Thank you! The signature of that function is actually something like

```ts
 beforeInsert?: <T extends Model>(model: T, ctx: QueryContext) => Promise<any> | void;
```

The first argument is the model instance that is being inserted. I don't really know if that is the best way to type that, since I don't use typescript. The `RelationThrough` can also have a `beforeInsert` hook, which could be added here too. A test case is also required in order for this to be merged. A simple usage example in examples.ts will suffice.

koskimas:
@hardeep Are you going to work on this or should I close this?

PR #894, opened 2018-05-01
Update modelClass typing in RelationThrough to align with RelationMapping

hanhsu:
- Update modelClass typing in RelationThrough to align with RelationMapping


koskimas:
Thanks!

Issue #895, opened 2018-05-01
Missing applyFilter() and eagerOptions() from TypeScript typings


gregl-flipp:

koskimas:
This issue seems to be missing words 😉

gregl-flipp:
@koskimas was just about to create a PR. Hopefully not much else describe aside from two methods being added :)

koskimas:
Fixed in #6c12c05

PR #896, opened 2018-05-01
Missing applyFilter() and eagerOptions() from TypeScript typings

gregl-flipp:
https://github.com/Vincit/objection.js/issues/895

- add the two missing methods to TypeScript typings


koskimas:
Thank you!

Issue #782, opened 2018-02-06
Cannot upsert anything under a node that is going to be related in upsertGraph.


muffinsparticus:
Greetings!

I've encountered an issue that's blocking me from moving forward with objection.js at the moment. 

I have a DataModel (A) that has a many to many relationship with another DataModel (B) that has a many to many relation to another DataModel (C). Whenever I perform an upsert with using DataModel A objection always tries to insert the relation between BxC into the proper join table when the relation exists. 

Is this a bug with the upsert graph and if not is there a way to prevent objection from inserting duplicate relations? It seems to handles the AxB relation just fine.

P.S. It also doesn't seem to properly unrelate the BxC the relation when it doesn't exist.

koskimas:
Hi,

unfortunately I have no idea what you mean by that descriptioin. `upsertGraph` is so complex, that your description can mean a hundred different things. Could you use the `reproduction-template.js` file in the repo root and create a simple example of the case you are talking about?

koskimas:
Also could you try the newest 1.0 RC version if you haven't already. It has a bunch of upsertGraph fixes.

muffinsparticus:
Thank you very much for the response! 

The project is breaking when I install version 1.0.0-RC12 and I can confirm it works with version 0.9.4. I'm currently using gulp to pipe the data models through babel to compile it to ES5 javascript. Yet when I try to run the code, after compiling, it gives me the following error message:

{ errorMessage: 'Unexpected token {',
      errorType: 'SyntaxError',
      stackTrace: 
       [ 'Module._compile (module.js:373:25)',
         'Object.Module._extensions..js (module.js:416:10)',
         'Module.load (module.js:343:32)',
         'Function.Module._load (module.js:300:12)',
         'Module.require (module.js:353:17)',
         'require (internal/module.js:12:17)',
         'Object.<anonymous> ( main.js:9:18 )',
         'Module._compile (module.js:409:26)',
         'Object.Module._extensions..js (module.js:416:10)' ] }

Are there any new requirements with version 1.0.0-RC12 that I should be aware of? Ex: new babel dependencies to install. I can provide more info about the build process if necessary.

In the meantime I'll try to get a test example ready using the repo-template.

koskimas:
There's nothing new that should or could break babel.

muffinsparticus:
Thank you for the reply!

I've attached a copy of an example using the repo file as requested below: 
[...]

PR #910, opened 2018-05-08
Fix/upsert graph 782

johannessjoberg:
This PR is aimed to solve this issue presented here: https://github.com/Vincit/objection.js/issues/782

In this discussion @koskimas presented 3 criteria to merge a potential PR: 

- [x] All current tests must pass (this should be obvious)

- [x] The new feature must be tested

- Integration tests are added for verifying that relating node with children is working for `ManyToMany`, `BelongsToOne` and `HasMany` relations
- Unit tests are added to verify that options object is correctly transformed depending on path from root
- Integration test for verifying that options are respected when relating node with children 
- [x] Must support all current options with boolean and relation path values. Especially remember to consider the case where relation paths point to models that are under the model that is about to be related.
- See `transformOptionsFromPath` function and unit tests


lib/queryBuilder/operations/UpsertGraphOperation.js
```diff
@@ -176,6 +178,23 @@ class UpsertGraphOperation extends InsertGraphOperation {
...
+
+    return Promise.map(
+      upsertRecursivelyNodes,
+      node => {
+        return node.upsertModel.constructor
+          .query()
+          .upsertGraph(node.upsertModel, transformOptionsFromPath(node.opt, node.relPathFromRoot))
+          .childQueryOf(builder, true);
+      },
+      { concurrency: this.graph.rootModelClass.concurrency }
```
igor-savin-ht: Considering that knex is intending to eventually drop Bluebird usage, is it safe relying on it being there? @koskimas

lib/utils/transformOptionsFromPath.js
```diff
@@ -0,0 +1,32 @@
...
+    const transformedOpt = transformOptionFromPath(options[key], relPathFromRoot);
+    if (transformedOpt) {
+      transformedOptions[key] = transformedOpt;
+    }
+  });
+
+  return transformedOptions;
+};
+
+const transformOptionFromPath = (opt, relPathFromRoot) => {
```
igor-savin-ht: since this is a public function, probably having jsdoc would be a good idea

koskimas: ? This is not public

koskimas: And jsdoc isn't used anywhere in objection codebase

igor-savin-ht: @koskimas Right, my mistake. Comment still applies, but should be at "transformOptionsFromPath" instead.

koskimas: That's not public either. Only stuff mentioned in the docs is considered public

lib/utils/transformOptionsFromPath.js
```diff
@@ -0,0 +1,32 @@
+const transformOptionsFromPath = (options, relPathFromRoot) => {
```
igor-savin-ht: Probably would be a good idea to add jsdoc since this is a public function. Also, any particular reason why this is defined as a constant arrow function? In other parts of objection.js I see exported functions defined simply as "function xyz (args)"

lib/utils/transformOptionsFromPath.js
igor-savin-ht: Fair point.

koskimas:
Oh, It's not travis' problem. Your tests fail for some reason. Objection simply swallows the actual error and shows the useless one seen in travis :grimacing: There is something wrong with your migrations I'd quess. I just tested and the master's test run without any problem.

lib/utils/transformOptionsFromPath.js
koskimas: I agree, the same style could be used for these for consistency

johannessjoberg: All right, got some different habits :) I'll fix that. Perhaps directly export the function?

johannessjoberg:
Thanks @koskimas. I'll get on that rebase and try to figure out what's wrong.

lib/queryBuilder/operations/UpsertGraphOperation.js
johannessjoberg: 👍 It's also used in similar fashion for `delete`, `relate` and `update` operation currently.

johannessjoberg:
If seems as if travis does like me like me much :| 

Builds pass for Node 9 prev commit, this commit for Node 6. But for others Node versions: `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`

koskimas:
@johannessjoberg The last one was just a random travis glitch. I restarted the tests and they finished successfully

igor-savin-ht:
@koskimas What is left before this can be merged?

koskimas:
@igor-savin-ht I haven't reviewed the code yet.

lib/queryBuilder/graphUpserter/UpsertNode.js
```diff
@@ -327,6 +348,22 @@ function hasChanges(currentModel, upsertModel) {
...
+function hasRelationsInUpsertModel(upsertModel) {
+  const relationArray = upsertModel.constructor.getRelationArray();
+  const changingProps = [];
+
+  for (let i = 0, l = relationArray.length; i < l; ++i) {
+    const relation = relationArray[i];
+    const upsertRelated = upsertModel[relation.name];
+
+    if (upsertRelated) {
+      changingProps.push(upsertRelated);
```
koskimas: The array is not needed. You could just return true here and remove the array right?

testUtils/TestSession.js
```diff
@@ -189,26 +216,78 @@ class TestSession {
...
+              .index();
+            table
+              .integer('model2Id')
+              .unsigned()
+              .notNullable()
+              .references('id_col')
+              .inTable('model2')
+              .onDelete('CASCADE')
+              .index();
           })
```
koskimas: Why did you change all these? Do you have different settings for `prettier` or something?

koskimas: Oh, I think my prettier script doesn't touch this file at all. This is fine :+1:

koskimas:
@johannessjoberg Everything looks good to me. I added a note about a small improvement (one array creation could be avoided). Other than that I'm ready to merge this. Awesome job, excellent tests, clean code :100:

koskimas:
@johannessjoberg Actually the one nitpick I found is so small that I don't want to bother you with it. Thanks again!

johannessjoberg:
Glad to hear everything looks good and thank you for the kind words! Always glad to be able to help out :) 

When do you think this will be released?

koskimas:
Soon, maybe even today.

johannessjoberg:
Awesome! 👍

koskimas:
@johannessjoberg 1.1.9 is out including this PR

johannessjoberg:
Great! Thanks @koskimas :)

PR #915, opened 2018-05-12
Add composite id typing for patchAndFetchById

woochi:
As stated in the [documentation](http://vincit.github.io/objection.js/#composite-keys) and visible in the code, `patchAndFetchById` supports finds with composite ids. However, currently the TypeScript typings only indicate that single column ids are supported.

This PR updates the typings to be in line with the implementation and documentation.


koskimas:
Thanks!

PR #928, opened 2018-05-18
Overview for API#models and tabbed-example component

rivertam:
I had to use a combination of vue components and a plugin for the
vuepress default theme. Vuepress's default config has the power to be
configured heavily, but it's a little weird. If I just used the vue
components in the md directly, it would take the children of the
components (which should be turned into codeblocks) and make that just
one text component, so it wouldn't get processed by markdown-it. But if
I used markdown-it-container, the HTML replacement occurs before
runtime.

It's possible using <ClientOnly> would have worked, but I didn't try.
This way feels more extensible, though slightly gross.

### DIFF
--- a/newDoc/api/model.md
+++ b/newDoc/api/model.md
@@ -1,6 +1,35 @@
 # class Model
 
-## static tableName
+> Subclasses of this class represent database tables.
+
+## Overview
+
+#### Model Lifecycle
+
+For the purposes of this explanation, let’s define three data layouts:
+
+1. `database`: The data layout returned by the database.
+1. `internal`: The data layout of a model instance.
+1. `external`: The data layout after calling model.toJSON().
+
+Whenever data is converted from one layout to another, converter methods are called:
+
+1. `database` -> [`$parseDatabaseJson`](/TODO/$parseDatabsaeJson) -> `internal`
+1. `internal` -> [`$formatDatabaseJson`](/TODO/$formatDatabaseJson) -> `database`
+1. `external` -> [`$parseJson`](/TODO/$parseJson) -> `internal`
+1. `internal` -> [`$formatJson`](/TODO/$formatJson) -> `external`
+
+So for example when the results of a query are read from the database the data goes through the [`$parseDatabaseJson`](/TODO/$parseDatabaseJson) method. When data is written to database it goes through the [`$formatDatabaseJson`](/TODO/$formatDatabaseJson) method.
+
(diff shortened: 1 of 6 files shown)


rivertam:
`<tabbed-component>` is very heavily based on https://github.com/spatie/vue-tabs-component.

This was my first productive foray into vue programming. I don't like it so far. :stuck_out_tongue:

But vuepress seems quite nice

newDoc/api/model.md
```diff
@@ -1,15 +1,60 @@
...
+For the purposes of this explanation, let’s define three data layouts:
+
+1. `database`: The data layout returned by the database.
+1. `internal`: The data layout of a model instance.
+1. `external`: The data layout after calling model.toJSON().
+
+Whenever data is converted from one layout to another, converter methods are called:
+
+1. `database` -> [`$parseDatabaseJson`](/TODO/$parseDatabsaeJson) -> `internal`
+1. `internal` -> [`$formatDatabaseJson`](/TODO/$formatDatabaseJson) -> `database`
```
koskimas: These links can already be filled I think. We know where the functions will be. In this case /api/model.html#parsedatabasejson or simple #parsedatabasejson since we are in the same file.

newDoc/api/model.md
```diff
@@ -1,15 +1,60 @@
...
+1. `external` -> [`$parseJson`](/TODO/$parseJson) -> `internal`
+1. `internal` -> [`$formatJson`](/TODO/$formatJson) -> `external`
+
+So for example when the results of a query are read from the database the data goes through the [`$parseDatabaseJson`](/TODO/$parseDatabaseJson) method. When data is written to database it goes through the [`$formatDatabaseJson`](/TODO/$formatDatabaseJson) method.
+
+Similarly when you give data for a query (for example [`query().insert(req.body))`](/TODO/insert) or create a model explicitly using [`Model.fromJson(obj)`](/TODO/fromjson) the [`$parseJson`](/TODO/$parseJson) method is invoked. When you call [`model.toJSON()`](/TODO/tojson) or [`model.$toJson()`](/TODO/tojson) the [`$formatJson`](/TODO/formatJson) is called.
```
koskimas: Same with the links in this file.

koskimas:
Awesome job, thank you! I think the links could be written out with the same amount of work than it takes to write `/TODO/`. I'll merge this now, but maybe you could use the final links in the next PR if you still have time to work on this.

rivertam:
@koskimas Yep good point. I actually wasn't reading the docs closely enough to realize they linked to the same file. Will definitely address next session.

PR #938, opened 2018-05-26
Simple parameter type conversion when using page

alansikora:
Hey!

I'm not sure if this is desirable, but I feel like pagination parameters often come from querystrings and maybe we should make sure those parameters are integer before doing calculations that will result in very strange results.

Thanks for Objection.js, it is a core module of my project and I love using it everyday.


koskimas:
Good idea! Thanks!

PR #949, opened 2018-06-05
add warning about knexSnakeCaseMappers in migrations

heisian:
Added a warning, but at this point I'm thinking it might even be prudent to remove the entire section about using `knexSnakeCaseMappers` in the `knexfile`.. unless we want to advise using a separate `knexfile` or config for migratons.

Ref [Knex.js issue #2644](https://github.com/tgriesser/knex/issues/2644)

### DIFF
--- a/doc/includes/RECIPES.md
+++ b/doc/includes/RECIPES.md
@@ -257,6 +257,11 @@ to camel case. You still need to use snake case in `relationMappings` and querie
 and their variants still take objects in camel case. The reasoning is that objects passed to those methods usually come
 from the client that also uses camel case.
 
+**NOTE**: Beware of using `knexSnakeCaseMappers` in migrations - Knex does not perform a conversion when using a
+property accessor for `max_batch` - a query result that tells Knex the latest migration batch number. If you use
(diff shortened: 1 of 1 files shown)


heisian:
Something like...
```node
const Knex = require('knex');
const { knexSnakeCaseMappers } = require('objection');

const knex = Knex({
  client: 'postgres',

  connection: {
    host: '127.0.0.1',
    user: 'objection',
    database: 'objection_test'
  }

  // Merge `postProcessResponse` and `wrapIdentifier` mappers.
  ...knexSnakeCaseMappers()
});

// DO NOT USE THIS AS A KNEXFILE FOR MIGRATIONS VIA THE CLI
```

heisian:
Eslint failing in tests but only one file changed in comparison to master? Perhaps the eslint rules were updated but not kept in sync w/ what's checked in tests?

koskimas:
prettier probably got updated and something changed. We should probably add a lock file just because of prettier.

koskimas:
Tests fail. Rebase master, run prettier and try again.

koskimas:
Thank you!

PR #985, opened 2018-06-28
API Reference: add Relation class

heisian:
The trend seems to be description, then example, then details.. @koskimas LMK if this is a good format before I do other pages.. seems like an almost 1:1 transfer is desired here?


heisian:
ah sorry about that, I didn't see your review from 8 days ago, please re-open

PR #1020, opened 2018-07-28
#1002: Implement fail-fast logic for relation resolution

### LINKED ISSUE #1002: "forEachChildExpression" does not fail fast
`objection/lib/queryBuilder/RelationExpression.js`
`forEachChildExpression`

Imagine case where we have 
```
allRelations = {
   childPages
}
```
and deeper down we have
```
node.$relation === 'child_pages'
```

Then we do
```
const relation = allRelations[node.$relation] || null;
```

Currently this will result in null being passed further, which typically will result in non-trivial to debug error deeper down. Are there any valid cases where null is an acceptable resolution outcome? 

Current effect of this is that during .relate operation on misnamed relationships you get
`TypeError: Cannot read property 'relatedModelClass' of null\n    at forEachExpr (/Users/igor.savin/WebstormProjects/core-services/node_modules/objection/lib/queryBuilder/operations/eager/RelationJoinBuilder.js:483:46)\n    at RelationExpression.forEachChildExpression`

kibertoad:
fixes #1002


kibertoad:
@koskimas Identical check in WhereInEagerOperation doesn't seem to be triggered anymore (at least judging by coverage report) since we are catching this early; should I remove it?

lib/queryBuilder/RelationExpression.js
```diff
@@ -185,9 +186,15 @@ class RelationExpression {
       for (let i = 0, l = childNames.length; i < l; ++i) {
         const childName = childNames[i];
         const node = this[childName];
-        const relation = allRelations[node.$relation] || null;
-        const childExpr = new RelationExpression(node, 0);
+        const relation = allRelations[node.$relation];
+        if (!relation) {
+          throw new ValidationError({
```
koskimas: I don't think this class should throw a `ValidationError` directly, but some other error and the catched can decide which error is thrown out from objection. Let's keep this class as low-level as possible.

kibertoad: > the catched can decide which error is thrown out from objection

Could you please clarify this part? Should all places that call this code have catch block for rethrowing specific errors?

koskimas: It should say "catcher" instead of "catched". I just mean that we should create some custom error and throw that and then the code that catches the error should convert it to a `ValidationError`. This doesn't work if the user has overwritten the `Model.createValidationError` method.

koskimas: It's called from two places and the other one doesn't have any error handling currently. You can just leave it as is. So really, it's just a matter of moving that code to a new place.

kibertoad: Pushed changes, hopefully I understood your idea correctly :)

PR #1022, opened 2018-07-28
Add caching to travis config

kibertoad:


koskimas:
How does the cache work? How long are the files cached? How does this work when a dependency is updated in package.json?

kibertoad:
From Travis documentation: "Before the build, we check if a cached archive exists. If it does, we download it and unpack it to the specified locations.

After the build we check for changes in the directory, create a new archive with those changes, and upload it to the remote storage."

So npm install will still be executed, but would take less time (only incremental changes) , hypothetically.

koskimas:
Ok, thanks!

PR #1024, opened 2018-07-29
Add missing RelationThrough#beforeInsert' and Join typings

rgtk:
http://vincit.github.io/objection.js/#relationmappings

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -213,7 +213,7 @@ declare namespace Objection {
     to: string | Reference | (string | Reference)[];
     modelClass?: ModelClass<any> | string;
     extra?: string[] | object;
-    beforeInsert?: (model: Model, context: QueryContext) => void;
+    beforeInsert?: (model: Model, context: QueryContext) => Promise<any> | void;
   }
 
   export interface RelationMapping {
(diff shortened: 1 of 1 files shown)


typings/objection/index.d.ts
```diff
@@ -198,6 +213,7 @@ declare namespace Objection {
     to: string | Reference | (string | Reference)[];
     modelClass?: ModelClass<any> | string;
     extra?: string[] | object;
+    beforeInsert?: (model: Model, context: QueryContext) => void;
   }
```
koskimas: This method can be async, so maybe the return value should be `void | Promise<any>` or something? I'm not sure how other async hooks have been typed.

koskimas:
It seems that you have duplicates of your old commits in this PR. Please rebase master to remove those.

rgtk:
Updated and added another missing typing.

typings/objection/index.d.ts
```diff
@@ -213,6 +213,7 @@ declare namespace Objection {
     to: string | Reference | (string | Reference)[];
     modelClass?: ModelClass<any> | string;
     extra?: string[] | object;
+    beforeInsert?: (model: Model, context: QueryContext) => Promise<any> | void;
```
mattleff: This should be `Promise<void> | void;`

rgtk:
Feedback fix applied.

PR #1031, opened 2018-08-03
Added andWhereRaw alongside whereRaw/orWhereRaw

wubzz:
While trying to incorporate objection into a rather large project I noticed `.andWhereRaw` had not been added in objections querybuilder.

Calling `.whereRaw` works just as well, but for the sake of consistency I figured it would be best to add `.andWhereRaw` as well.


koskimas:
Thank you!

PR #1038, opened 2018-08-10
TypeScript: Allow non-static values for 'patch' and 'update'

### LINKED ISSUE #1036: TypeScript: 'patch' input type does not allow 'Raw' values
`patch` takes `Partial<T>` which expects a subset of `T`. This means it does not allow for any `Raw` values in the corresponding objects, unless they are defined within `T`.

Based on the definition of `Partial<T>`

```ts
type Partial<T> = { [P in keyof T]?: T[P] }
```

I suggest the following for `patch` and `update`:

```ts
type Update<T> = { [P in keyof T]?: T[P] | Raw}
```

Does that sound reasonable?

xsc:
The following dynamic values are now allowed to be passed as column
values to `patch` and `update`:

- Knex.Raw
- Reference
- QueryBuilder

Fixes #1036.


koskimas:
Thanks!

PR #1040, opened 2018-08-14
Correct TS definition for graph options interfaces

mattleff:
Matches docs [here](http://vincit.github.io/objection.js/#upsertgraphoptions).


koskimas:
Thanks!

PR #1054, opened 2018-09-03
Fix a few typos in docs

tamlyn:
- plaque -> plague
- write queries to the documents -> write queries for the documents
- is ran -> is run


koskimas:
Thanks

PR #1067, opened 2018-09-11
Use up-to-date nyc instead of deprecated istanbul

kibertoad:
https://github.com/gotwarlost/istanbul is pretty much dead, all the active development is happening in https://github.com/istanbuljs/istanbuljs/tree/master/packages/istanbul-lib-coverage and nyc is the most convenient way to consume it.


koskimas:
Thanks Igor!

Issue #978, opened 2018-06-26
ajv option v5 is depreceated


vonagam:
Option `v5` (used in [validator](https://github.com/Vincit/objection.js/blob/e220e51374a439b8b5dfd82b85f08b49d37ece32/lib/model/Model.js#L339) and shown in documentation for "Custom validation") is [no longer](https://github.com/epoberezkin/ajv/commit/69ac3d18c7c5f587725e203c675c623578e49182) used in ajv.

It was replaced with [$data](https://github.com/epoberezkin/ajv#data-reference) option, some methods were moved to [ajv-keywords](https://github.com/epoberezkin/ajv-keywords).

kibertoad:
@vonagam Can you provide more detailed mapping? setting $data to true actually breaks an Objection.js ajv test.

vonagam:
Well, your way was right. There is no mapping, it should be that easy.

But i am too getting errors from test for 411.

While investigating i found out that order of arguments have mismatch here: [1](https://github.com/Vincit/objection.js/blob/14d00cc8915839bc480e8825dff5e5b5fe9a9137/lib/model/AjvValidator.js#L55) and [2](https://github.com/Vincit/objection.js/blob/master/tests/integration/misc/%23411.js). So you end up replacing `jsonSchema` with `json`. Unfortunately, it does not solve the real cause of the problem.

I also found out that [`compileNormalValidator(jsonSchema)`](https://github.com/Vincit/objection.js/blob/14d00cc8915839bc480e8825dff5e5b5fe9a9137/lib/model/AjvValidator.js#L139) does not return a validating function with property `schema` set to passed `jsonSchema` as it should do. That's why the validation fails.

To understand why the property set wrongly i would have to debug this more seriously, but i personally do not have time for this and currently do not use ajv for validation myself. Sorry that i couldn't help.

PR #1068, opened 2018-09-11
Use correct option for a newer ajv version

kibertoad:
refs https://github.com/Vincit/objection.js/issues/978


kibertoad:
This is broken, closing for now

PR #1084, opened 2018-09-20
Add objection-transactional-tests to list of plugins

lautis:
When testing applications using Objection, it's nice to be able to run those tests in isolation and in parallel. Adds objection-transactional-tests to the list of plugins for others to find if they have the same use case.


koskimas:
Thanks!

PR #1117, opened 2018-10-17
fix: Ensure resultSize resolves to a number.

### LINKED ISSUE #1116: resultSize returns a string
This may be related to the driver, I was using mysql when I came across this, but in my case calling `resultSize` on a query builder is returning a string. Considering the [fallback in the method](https://github.com/Vincit/objection.js/blob/master/lib/queryBuilder/QueryBuilder.js#L525) is to return the number `0`, I think this is a bug. 

To compound the issue for me, the [TS typing](https://github.com/Vincit/objection.js/blob/master/typings/objection/index.d.ts#L849) shows it returns a `Promise<number>`.

If passing the value through `parseInt` would not be considered backwards breaking, I'd be happy to open a PR.

mastermatt:
Fixes #1116.

Passing a number to `parseInt` is safe so this won't break any calling
code that is already manually doing the casting.

### DIFF
--- a/tests/unit/queryBuilder/QueryBuilder.js
+++ b/tests/unit/queryBuilder/QueryBuilder.js
@@ -729,15 +729,16 @@ describe('QueryBuilder', () => {
   });
 
   it('resultSize should create and execute a query that returns the size of the query', done => {
-    mockKnexQueryResults = [[{ count: 123 }]];
+    mockKnexQueryResults = [[{ count: '123' }]];
     QueryBuilder.forClass(TestModel)
       .where('test', 100)
       .orderBy('order')
+      .limit(10)
+      .offset(100)
       .resultSize()
       .then(res => {
         expect(executedQueries).to.have.length(1);
         expect(res).to.equal(123);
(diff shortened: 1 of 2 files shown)


tests/unit/queryBuilder/QueryBuilder.js
```diff
@@ -729,15 +729,16 @@ describe('QueryBuilder', () => {
...
     QueryBuilder.forClass(TestModel)
       .where('test', 100)
       .orderBy('order')
+      .limit(10)
+      .offset(100)
       .resultSize()
       .then(res => {
         expect(executedQueries).to.have.length(1);
         expect(res).to.equal(123);
-        // resultSize should cancel the groupBy call since it doesn't affect the outcome.
```
mastermatt: Removed this comment because a `GROUP BY` on the underlying query would affect the outcome of a `COUNT`.

koskimas: Yeah it should have said "orderBy" instead.

Issue #1115, opened 2018-10-16
fn is not a true shortcut for Model.knex().fn.


mastermatt:
I was thrown for a loop today when I couldn't use `MyModel.fn.now()`, but I could use `MyModel.knex().fn.now()` and the [docs](http://vincit.github.io/objection.js/#fn) say that it's just a shortcut.

The issue is that in [`Model#fn`](https://github.com/Vincit/objection.js/blame/master/lib/model/Model.js#L505) is a static method while in [`knex`](https://github.com/tgriesser/knex/blob/d4a3387c56c0ec93ad554074a514d4300cea64ef/src/util/make-knex.js#L116-L121) it's a getter.

The solution for me is to use `MyModel.fn().now()`.

I think `Model#fn` should be made a getter as well, but at the very least the docs should be updated to be less ambiguous.

koskimas:
Weird, that's been there since v0.1 and no one has brought this up before. I'll update the docs. We can't really change the method before 2.0 because it's a breaking change.

mastermatt:
Can this issue be linked under the [2.0](https://github.com/Vincit/objection.js/projects/1) project? I've never used GH projects, not sure myself.

koskimas:
Fixed by #1119

PR #1119, opened 2018-10-17
proposal: Convert Model.fn to a getter.

mastermatt:
Relates to #1115;

Turning `fn` into a getter better reflects the intention of shortcutting
to `knex.fn`, however, would be a breaking change if made alone.

This proposed patch converts it to a getter, but the returned
`knex#FunctionHelper` instance is callable and returns itself.
The effect is that `Model.fn.now()`, `Model.fn().now()`, and
`Model.knex().fn.now()` all produce the same result.

### DIFF
--- a/lib/model/Model.js
+++ b/lib/model/Model.js
@@ -502,9 +502,19 @@ class Model {
     return raw.apply(undefined, arguments).toKnexRaw(this);
   }
 
-  static fn() {
-    const knex = this.knex();
-    return knex.fn;
+  /**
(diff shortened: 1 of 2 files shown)


mastermatt:
Note: the current [TS typings](https://github.com/Vincit/objection.js/blob/master/typings/objection/index.d.ts#L396) make it seem as if `Model.fn.now()` is already the correct method and prohibits `Model.fn().now()` from being transpiled. Forcing TS users to use `Model.knex().fn.now()`.

koskimas:
Nice! Thank you.

Issue #1121, opened 2018-10-20
applyModifier('unkown')  returns internal exception


lehni:
This relates to my work on #1120: I just noticed that when calling `applyModifier('unkown')` on a query builder, a seemingly internal ` ModifierNotFoundError` is thrown without any error message set on it. Is this intentional? Shall I address it in the upcoming PR for #1120?

lehni:
I just noticed that the problem goes further: `ModifierNotFoundError.modiferName` is `undefined` instead of the actual unknown modifier in those situations. I also addressed this in #1122

PR #1122, opened 2018-10-20
Improve error message thrown for unknown modifiers

lehni:
This PR addressed the problem in #1121

When calling `applyModifier('unkown')` on a query builder, a seemingly internal ` ModifierNotFoundError` is thrown without any error message set on it.

This change gives that error a human readable message.

### DIFF
--- a/lib/utils/createModifier.js
+++ b/lib/utils/createModifier.js
@@ -2,7 +2,7 @@ const { asArray, isString, isFunction, isPlainObject } = require('./objectUtils'
 
 class ModifierNotFoundError extends Error {
   constructor(mofifierName) {
-    super();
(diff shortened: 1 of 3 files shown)


lib/utils/createModifier.js
```diff
@@ -17,9 +17,11 @@ function createModifier({ modelClass, modifier, modifiers, args }) {
     let modify = null;
 
     if (isString(modifier)) {
-      modifier = modifiers[modifier] || modelModifiers[modifier];
+      modify = modifiers[modifier] || modelModifiers[modifier];
```
lehni: @koskimas this was replacing `modifier` here with `undefined`, leading to wrong exception also that mentioned `undefined` instead of the actually missing modifier name.

Issue #1120, opened 2018-10-19
Support a catch-all modifier hook


lehni:
I'd like to propose to add a way for a model class to provide a catch-all modifier for the case when a requested modifier isn't defined / found. Something along these lines:

```js
class MyModel extends Model {
  ...

  static modifierNotFound(builder, modifier) {
    // Do something with the modifier and the query, and return the builder if it was
    // handled, nullish otherwise, indicating Objection should still throw an error.
    const { properties } = this.getJsonSchema()
    if (properties && modifier in properties) {
      return builder.select(modifier)
    }
  }
}
```

Optionally it could also just return a boolean.

I think this could open the door to many more creative uses of eager expressions and modifiers, and would be a pretty great feature that could be added with very little overhead. I'd be happy to create a PR for it.

@koskimas what do you think?

koskimas:
Yes, let's do it! A PR would be great!

lehni:
Very cool! Do you like the naming?

lehni:
I think returning boolean is cleaner probably?

koskimas:
I would have probably just added a default implementation that throws the error and overriding the method would automatically cause no error to be thrown. Boolean is fine too.

lehni:
I can do that, yes! But what if my overriding implementation doesn't handle a given modifier and would like to throw the sam error?

I was thinking yesterday after doing this that instead of directly handling it, it could return a function if a given modifier should be handled.

koskimas:
In that case it can delegate to the default implementation `super.noModifierFound(builder, modifier)`. Why would it return a function?

lehni:
I can do that, yes! But what if my overriding implementation doesn't handle a given modifier and would like to throw the sam error?

I was thinking yesterday after doing this that instead of directly handling it, it could return a function if a given modifier should be handled.

lehni:
Because it wouldn't necessarily receive the `builder` then, just the modifier, and somebody could do their own `modifier` store, similar to `modifiers` which are also functions... Just a thought for more flexibility. and you're right, delegation would solve that. I change it.

lehni:
I've rewritten this now as discussed. I needed to move `ModifierNotFoundError` to model and things got a bit more complicated in result. I'm not sure I've done everything how you would address it. PTAL.

PR #1123, opened 2018-10-20
Implement modifierNotFound() catch-all hook

lehni:
This PR relates to #1120 

It adds a way for a model class to provide a catch-all modifier for the case when a requested modifier isn't defined / found.

~~It is still work in progress, as its tests depend on #1122 to be merged, and docs still need to be written once we agree on the API surface.~~

```js
class MyModel extends Model {
  ...

  static modifierNotFound(builder, modifier) {
    // Do something with the modifier and the query, and return `true` if it was
    // handled, falsy otherwise, indicating Objection should still throw an error.
    const { properties } = this.getJsonSchema()
    if (properties && modifier in properties) {
      builder.select(modifier)
      return true
    }
  }
}
```

### DIFF
--- a/doc/includes/API.md
+++ b/doc/includes/API.md
@@ -2418,7 +2418,7 @@ Type|Description
 
 
 
-#### applyFilter
+#### applyModifier
 
 Applies modifiers to the query builder.
 
@@ -2434,6 +2434,13 @@ modifier|string|The name of the modifier, as found in [`modifiers`](#modifiers).
 Type|Description
 ----|-----------------------------
 [`QueryBuilder`](#querybuilder)|`this` query builder for chaining.
+
+
+
+
+#### applyFilter
+
+An alias for [`applyModifier`](#applymodifier)
 
 
 
@@ -6182,6 +6189,40 @@ row|Object|A database row.
 Type|Description
 ----|-----------------------------
 [`Model`](#model)|The created model instance
+
+
+
+
+#### modifierNotFound
+
+```js
+class BaseModel extends Model {
+  static modifierNotFound(builder, modifier) {
+    const { properties } = this.jsonSchema
+    if (properties && modifier in properties) {
+      builder.select(modifier)
+    } else {
+      super.modifierNotFound(builder, modifier)
+    }
+  }
+}
+```
+
+Handles modifiers that are not recognized by the various mechanisms that can specify
+them, such as [`modify`](#modify) and [`applyModifier`](#applymodifier), as well as
+the use of modifiers in eager expressions (see [`RelationExpression`](#relationexpression))
+and in relations (see [`RelationMapping`](#relationmapping).
+
+By default, the static `modifierNotFound()` hook throws a `ModifierNotFoundError` error.
+If a model class overrides the hook, it can decide to handle the modifer through the passed
+`builder` instance, or call the hook's definition in the super class to still throw the error.
+
+##### Arguments
+
+Argument|Type|Description
+--------|----|-------------------
+builder|[`QueryBuilder`](#querybuilder)|The query builder on which to apply the modifier.
+modifier|string|The name of the unknown modifier.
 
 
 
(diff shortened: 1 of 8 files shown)


lehni:
*Update:* Since #1122 was merged, I've adjusted this PR to activate all tests. This is good to from my side now.

lehni:
Now with docs too: 6853f7465bf83d2ef68613f5d845b3a10d7dfe64

doc/includes/API.md
```diff
@@ -6186,6 +6193,45 @@ Type|Description
 
 
 
+#### modifierNotFound
+
+```js
+class BaseModel extends Model {
+  static modifierNotFound(builder, modifier) {
+    const { properties } = this.getJsonSchema()
```
koskimas: getJsonSchema() is not a public method.

koskimas: getJsonSchema() is not a public method.

lib/queryBuilder/operations/eager/RelationJoinBuilder.js
```diff
@@ -584,7 +584,7 @@ function createModifierQuery({ builder, modelClass, expr, modifiers, relation })
         modifiers
       });
     } catch (err) {
-      if (err instanceof ModifierNotFoundError) {
+      if (err instanceof modelClass.ModifierNotFoundError) {
         throw modelClass.createValidationError({
```
lehni: @koskimas is it better to import the error class or reference it like here?

lib/utils/createModifier.js
```diff
@@ -1,12 +1,5 @@
 const { asArray, isString, isFunction, isPlainObject } = require('./objectUtils');
 
-class ModifierNotFoundError extends Error {
-  constructor(mofifierName) {
-    super(`Unable to determine modify function from provided value: "${mofifierName}".`);
-    this.mofifierName = mofifierName;
```
lehni: @koskimas I've spotted a spelling mistake here: `mofifierName`. I've corrected this in the new location of the class.

lib/utils/createModifier.js
```diff
@@ -41,6 +34,5 @@ function createModifier({ modelClass, modifier, modifiers, args }) {
 }
 
 module.exports = {
-  createModifier,
-  ModifierNotFoundError
+  createModifier
```
lehni: @koskimas I'm not sure if you want to keep exporting this as a named export now that there's only one export in the file?

doc/includes/API.md
```diff
@@ -6186,6 +6193,46 @@ Type|Description
...
+--------|----|-------------------
+builder|[`QueryBuilder`](#querybuilder)|The query builder on which to apply the modifier.
+modifier|string|The name of the unknown modifier.
+
+##### Return value
+
+Type|Description
+----|-----------------------------
+boolean|`true` if the unknown modifier was handled, `false` otherwise, indicating an error should be thrown.
+
```
koskimas: Isn't this now wrong?

lehni: Yes indeed, I forgot to remove that. When there's no return value at all, I just remove the list, yes?

koskimas:
@lehni Some tests still fail.

lehni:
@koskimas should be fixed now!

lib/queryBuilder/operations/eager/RelationJoinBuilder.js
```diff
@@ -575,16 +575,16 @@ function createModifierQuery({ builder, modelClass, expr, modifiers, relation })
...
     try {
-      modifier = createModifier({
-        modifier: modifierName,
-        modelClass: relation.relatedModelClass,
-        modifiers
-      });
+      modifier(modifierQuery);
     } catch (err) {
-      if (err instanceof ModifierNotFoundError) {
+      if (err instanceof modelClass.ModifierNotFoundError) {
```
lehni: @koskimas this is why the tests were failing: The exceptions are now thrown when using the modifier, not when creating it.

lib/relations/Relation.js
```diff
@@ -347,24 +355,12 @@ function createRelationProperty(ctx, refString, propName) {
...
-    }
-  } else {
-    modify = () => {};
-  }
+  const modify =
+    modifier &&
+    createModifier({
+      modifier,
+      modelClass: ctx.relatedModelClass
+    });
```
lehni: @koskimas not sure you like writing it like this?

koskimas: That's fine.

PR #1178, opened 2018-12-19
RelationMapping TS interface

ericmuyser:
`filter` should support string and object per documentation
`beforeInsert` should be supported

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -220,7 +220,7 @@ declare namespace Objection {
     relation: Relation;
     modelClass: ModelClass<any> | string;
     join: RelationJoin;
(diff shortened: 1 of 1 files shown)


typings/objection/index.d.ts
```diff
@@ -221,7 +221,8 @@ declare namespace Objection {
     modelClass: ModelClass<any> | string;
     join: RelationJoin;
     modify?: (queryBuilder: QueryBuilder<any>) => QueryBuilder<any>;
-    filter?: (queryBuilder: QueryBuilder<any>) => QueryBuilder<any>;
+    filter?: ((queryBuilder: QueryBuilder<any>) => QueryBuilder<any>) | string | object;
```
koskimas: `filter` is an alias for `modify`. `modify` should have the same type.

koskimas:
Thanks!

PR #1184, opened 2018-12-25
Disable optional increment and decrement amount

skatpgusskat:
When I use `increment` function, I set amount as undefined, then error occur likes below.
```
Error: Undefined binding(s) detected when compiling RAW query: `abc` + ?
(abc is my column name)
```

So, I think the function `increment(columnName: string, amount?: number)` should be `increment(columnName: string, amount: number)`, update amount from optional to normal.


koskimas:
Tests fail.

koskimas:
@skatpgusskat Are you going to fix the tests, or should I close this?

PR #1181, opened 2018-12-21
Use strict mode

ratson:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

Tested on my Mac with Node v11.5, it is slightly faster or identical, so I think it is good idea to use strict mode for all sources.

without 'use strict'
```
$ mocha --slow 60000 --timeout 60000 --reporter spec --recursive perf

  Performance tests
    Model methods (85 models in the dataset)
      runtime: 5067 ms, 0.317 ms / run
      ✓ 16000 fromJson calls for the dataset (1360000 individual models) [goal 10000 ms]
      runtime: 2587 ms, 0.043 ms / run
      ✓ 60000 $toJson calls for the dataset (5100000 individual models) [goal 10000 ms]
      runtime: 3331 ms, 0.028 ms / run
      ✓ 120000 $toDatabaseJson calls for the dataset (3000000 individual models) [goal 10000 ms]
      runtime: 3310 ms, 0.028 ms / run
      ✓ 120000 $toDatabaseJson calls for the dataset with $omitFromDatabaseJson (3000000 individual models) [goal 10000 ms]
      runtime: 5316 ms, 0.053 ms / run
      ✓ 100000 $clone calls for the dataset (8500000 individual models) [goal 10000 ms]
      runtime: 2805 ms, 0.014 ms / run
      ✓ 200000 traverse calls for the dataset [goal 10000 ms]
```

with 'use strict'
```
$ mocha --slow 60000 --timeout 60000 --reporter spec --recursive perf

  Performance tests
    Model methods (85 models in the dataset)
      runtime: 4983 ms, 0.311 ms / run
      ✓ 16000 fromJson calls for the dataset (1360000 individual models) [goal 10000 ms]
      runtime: 2597 ms, 0.043 ms / run
      ✓ 60000 $toJson calls for the dataset (5100000 individual models) [goal 10000 ms]
      runtime: 3291 ms, 0.027 ms / run
      ✓ 120000 $toDatabaseJson calls for the dataset (3000000 individual models) [goal 10000 ms]
      runtime: 3119 ms, 0.026 ms / run
      ✓ 120000 $toDatabaseJson calls for the dataset with $omitFromDatabaseJson (3000000 individual models) [goal 10000 ms]
      runtime: 5077 ms, 0.051 ms / run
      ✓ 100000 $clone calls for the dataset (8500000 individual models) [goal 10000 ms]
      runtime: 2797 ms, 0.014 ms / run
[...]

PR #1189, opened 2019-01-07
Enforce use strict rule with ESLint

ratson:
Continue the effort in https://github.com/Vincit/objection.js/pull/1181


koskimas:
Thanks!

PR #1237, opened 2019-03-02
Add useLimitInFirst property to ModelClass type def

carlbennettnz:


koskimas:
@carlbennettnz Thanks for the PR! Your change doesn't pass the tests though. I'll merge this as soon as this is changed to pass the test.

carlbennettnz:
Looks like making the parameter optional fixed it

PR #1242, opened 2019-03-05
Add typings for range() without arguments

### LINKED ISSUE #1241: Typings do not support range() without arguments
The typings for `range()` do not allow calling it without arguments, even though that can be done if you specify `limit` and `offset` manually.

luckycadow:
Closes #1241

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -887,6 +887,7 @@ declare namespace Objection {
     resultSize(): Promise<number>;
 
     page(page: number, pageSize: number): QueryBuilder<QM, Page<QM>>;
+    range(): QueryBuilder<QM, Page<QM>>;
     range(start: number, end: number): QueryBuilder<QM, Page<QM>>;
     pluck(propertyName: string): this;
     first(): QueryBuilderYieldingOneOrNone<QM>;


typings/objection/index.d.ts
```diff
@@ -887,6 +887,7 @@ declare namespace Objection {
     resultSize(): Promise<number>;
 
     page(page: number, pageSize: number): QueryBuilder<QM, Page<QM>>;
+    range(): QueryBuilder<QM, Page<QM>>;
```
kibertoad: shouldn't start and end parameters just be marked optional for that?

luckycadow: If they were marked as optional then the typings would allow `.range(10)`, but that's not something objection will handle internally.

kibertoad: Makes sense.

koskimas:
Why would anyone call this function without arguments? Isn't that just the most confusing code ever expecially considering that objection doesn't document the default values? I don't think this should be merged. Instead, the default values should be removed.

luckycadow:
@koskimas From the documentation `range can be called without arguments if you want to specify the limit and offset explicitly` 
https://vincit.github.io/objection.js/#range

luckycadow:
I think the way `range()` works with arguments is more confusing.  It's easy to assume that `.range(limit, offset + limit)` would produce the same set of records as calling `.limit()` and `.offset()` with the same values.

koskimas:
Oh, I didn't remember that feature 🤐

PR #1254, opened 2019-03-17
Update knex explicitly to bust cache

kibertoad:


kibertoad:
OK, this brings Objection.js up-to-date as far as Knex is concerned :)

package.json
```diff
@@ -68,7 +68,7 @@
     "knex": "0.x"
   },
   "devDependencies": {
-    "@types/knex": "^0.14.5",
+    "@babel/polyfill": "^7.2.5",
```
kibertoad: Knex throws an error of polyfill dependency is not available now since it relies on it for Node 6 support since 0.16.x

package.json
```diff
@@ -68,7 +68,7 @@
     "knex": "0.x"
   },
   "devDependencies": {
-    "@types/knex": "^0.14.5",
```
kibertoad: Types are included with knex itself now.

package.json
```diff
@@ -80,7 +80,7 @@
     "expect.js": "^0.3.1",
     "fs-extra": "^7.0.1",
     "glob": "^7.1.3",
-    "knex": "0.x",
+    "knex": "0.16.4-next2",
```
kibertoad: Disregard the "next" status of it, it's definitely the best available Knex version at this point.

capaj: too bad that the 0.16.4 isn't out yet

package.json
koskimas: Crap! We can only remove node 6 support in 2.0 since we follow semver. We'll just have to live with this dependency.

koskimas:
Nice, thank you! So was it travis cache that prevented the update? Or what?

package.json
koskimas: Oh, it's a devDependency. Well that's ok I quess. Do all knex users need to install that?

koskimas: The knex `lib` is littered with these `var _assign = require('babel-runtime/core-js/object/assign');` so I quess everyone needs to always install that. Why isn't it just a dependency?

koskimas: Never mind. I just hadn't built knex in a while :man_facepalming:

kibertoad:
> So was it travis cache that prevented the update? Or what?

I think so, yes. Most likely any kind of modification of package.json would force knex version to also be updated, so probably we can get back to using 0.*, but I'm somewhat concerned about false sense of security that gives in cases where package.json is not being modified for some time.

package.json
kibertoad: >Do all knex users need to install that?

Only those who are still using Node 6. But they get pretty explicit error message if they don't, so hopefully it's an easy fix on their side. It is sorta kinda somewhat breaking change, but Node 6 is getting deprecated soon anyway.

koskimas:
Let's just remove the cache! It's not that useful anyway

PR #1258, opened 2019-03-26
Bugfix: Include related virtualAttributes in toJSON() calls

### LINKED ISSUE #1252: Pick return virtual attributes?
Would be sweet if ```pick``` could work similar to ```virtualattributes``` http://vincit.github.io/objection.js/#virtualattributes wihout the need to defined virtualAttributes and call ```.json()```

```ts
class Person extends Model {
  firstName: string;
  lastName: string;

  fullName() {
    return `${this.firstName} ${this.lastName}`;
  }
}

Person.query().pick(Person, ['id', 'fullName']).first();
```
or even a way to append virtual attributes per query. Something like
```ts
Person.query().append(Person, ['fullName']).first();
```

Thanks!

venables:
**Bug**: When calling toJSON on a model with related objects, the related objects do not include their virtualAttributes.  

**Cause**: Due to the change introduced in https://github.com/Vincit/objection.js/commit/55303773da82e196b732803f112118fcc3bee9b2 (to close #1252), calls to `toJSON` set an array of virtualAttributes for `opt.virtuals`.  This array is then passed to child/related object `toJSON` calls.  This causes the related models to believe they were called with `toJSON({ virtuals: [<parent attributes>] })`, thus excluding their own virtualAttributes.

For example, if a `Post` belongs to one `User`, and the `User` has a virtual `fullName`,  calling `post.toJSON()` would not include `user.fullName`

I can confirm `v1.6.3` works as expected, and `v1.6.4` introduced the bug.

**Solution:**
This PR partially reverts https://github.com/Vincit/objection.js/commit/55303773da82e196b732803f112118fcc3bee9b2 to allow for virtualAttributes to be calculated on a per-model basis by default.  

Another solution would be to entirely revert 55303773da82e196b732803f112118fcc3bee9b2

### DIFF
--- a/lib/model/modelToJson.js
+++ b/lib/model/modelToJson.js
@@ -8,7 +8,7 @@ function toJson(model, optIn) {
   const modelClass = model.constructor;
 
   const opt = {
-    virtuals: getVirtuals(optIn, modelClass),
+    virtuals: getVirtuals(optIn),
     shallow: isShallow(optIn),
     omit: getOmit(optIn, modelClass),
     pick: null,
@@ -40,13 +40,13 @@ function toDatabaseJson(model, builder) {
   return mergeQueryProps(model, json, opt.omitFromJson, builder);
 }
 
-function getVirtuals(opt, modelClass) {
+function getVirtuals(opt) {
   if (!opt) {
-    return modelClass.getVirtualAttributes();
+    return true;
   } else if (Array.isArray(opt.virtuals)) {
     return opt.virtuals;
   } else if (opt.virtuals) {
-    return modelClass.getVirtualAttributes();
+    return true;
   } else {
     return null;
   }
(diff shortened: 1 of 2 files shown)


koskimas:
Oh, crap! I thought there was a test for this case, but it uses the same model class for the related items. I'll merge this or some other fix as soon as I get to my computer.

feross:
I believe I also experienced the breakage that this PR addresses.

venables:
@koskimas Considering this bug was introduced in a patch release, most apps will pick up this bug due to npm defaults.

What are your thoughts on reverting the original change to fix the apps in the wild, and re-relasing as a minor version bump?

koskimas:
Yeah, I'll release a new version right now. Why would I revert the original change instead of fixing this bug?

koskimas:
By the way, thanks for fixing this @venables :bowing_man:

koskimas:
objection 1.6.5 is out

PR #1280, opened 2019-04-08
Allow relation lookup to accept Model classes

maxnordlund:
This allows `person.$relatedQuery(Movie).relate(50)`, to use the example from the docs. Using `instanceof Model` ensures all Model subclasses are supported. If you would have used `this`, only subclasses of that particular Model would work, which is not want you want. In the example above, that would mean only `Person` instances would be accepted, and not `Movie`s.

### DIFF
--- a/lib/model/Model.js
+++ b/lib/model/Model.js
@@ -637,6 +637,10 @@ class Model {
   }
 
   static getRelation(name) {
+    if (name instanceof Model) {
+      return name
+    }
+
     const relation = this.getRelationUnsafe(name);
 
     if (!relation) {


koskimas:
If there's such an example, it needs to be removed. I won't merge this.

koskimas:
Model can have multiple relations to same model class.

maxnordlund:
Hold on, all this does is make it possible to use a model class directly instead of it's relation name. The returned value is just the same, no?

maxnordlund:
If that is not the case, then the docs need to be updated to explain this in (more) detail.

koskimas:
What would `person.$relatedQuery(Movie).relate(50)` do if `Person` has two relationships to `Movie`?

koskimas:
Could you show me the place that suggest that this kind of thing is possible?

koskimas:
Also a model class is not an instance of any model. The implementation is completely wrong. Also you have zero tests for this and the current tests fail.

koskimas:
The first argument to `$relatedQuery` is the name of the **relation**. Not the model.

koskimas:
Also, `getRelation` returns a `Relation` instance, never a model class or an instance.

maxnordlund:
Ok let start from the top:

I was just writing some code for my company where I'm adding an user to a group, many-to-many. Following the example I figured it should be `group.$relatedQuery("users").relate(user)`. But then I thought it might be better to avoid hard coded strings, and use the model class itself `group.$relatedQuery(User).relate(user)`, only that doesn't work.

Thus, I made this PR as an suggestion. Unfortunately I didn't understand there are `Relation` objects floating about, and not just straight out models. So you're right that having multiple relations to the same model makes the behaviour undefined. Yes, it should have been `Object(name).prototype instanceof Model`, but that's why you have PRs and code review ;)

The behaviour I wanted was for using model classes as a shorthand for the _first and only_ relation to that class. This PR does not do that, and after thinking a bit more about it adds more complexity then IMHO it's worth.

---

On a final not I would like to stress the point that this PR is an suggestion, and not to be taken as gospel. It's my fault for not communicating that clearly. It seems like assumptions seldom hold. I also would like to make a suggestion to you as well. Being terse is a good thing, but for an outsider like that doesn't know the internals by heart, I need a bit more info to understand. I'm sure it wasn't your intention, but the comments felt downright hostile.

PR #1282, opened 2019-04-08
Add typings for modifiers

### LINKED ISSUE #1278: Missing typings for modifiers
Would it be possible to add typings for [`Model.modifiers`](https://vincit.github.io/objection.js/#modifiers)? They seem to be missing at the moment.

They could be something like:

```ts
export interface Modifiers {
   [name: string]: (builder: QueryBuilder<Model>) => QueryBuilder<Model>;
}

// ...

export class Model {
  static modifiers: Modifiers;
}
```

I'm not sure if the types can be narrowed somehow to only point to the "current" model...

---

If this issue is valid, I am happy to open a PR to correct the typings.

willsoto:
Closes #1278

---

I am honestly not sure if this is ideal given the discussion we had in #1278, but I couldn't get the existing typings to work. Would love some help there if it's possible to make them work.

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -94,6 +94,14 @@ class Person extends objection.Model {
     const itemMessage: string = errorItem.message;
     return new CustomValidationError('my custom error: ' + message + ' ' + itemMessage);
   }
+
+  static get modifiers() {
+    return {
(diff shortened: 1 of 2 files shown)


koskimas:
Thais!

PR #1325, opened 2019-05-09
Added type definition for aliasFor()

jc-white:
I noticed there was no type definition for aliasFor() so I have added it to the typings.


koskimas:
Thanks!

Issue #1292, opened 2019-04-12
Unable to pass two or more columns into .whereIn method due to types inconsistency.


emahnovets:
Hi guys.

I've find some inconsistency in objection types.
Knex supports array of column names to be passed into `.whereIn` method but in objection.js only one ColumnRef supported. 

https://knexjs.org/#Builder-whereIn
https://github.com/Vincit/objection.js/blob/master/typings/objection/index.d.ts#L1270

Example of query:
```
Response
      .query()
      .whereIn(['entity_id', 'entity_version'], subquery)
      .andWhere(options);
```

PR #1327, opened 2019-05-13
Adjust whereIn and whereInComposite typings to allow multiple columns

jbagio:
Closes https://github.com/Vincit/objection.js/issues/1292


koskimas:
Are you sure you can do things like `whereIn(['foo', 'bar'], [[1, 2], [3, 4], [5, 6]])` using knex? Also, could you add "tests" for this in the example.ts file?

jbagio:
Sure, I've added a (quite nonsensical) whereIn example, let me know if this is good enough.
There are currently no examples for whereInComposite (or whereComposite for that matter), would you want me to add those as well?

koskimas:
That's good enough, thanks!

Issue #1251, opened 2019-03-15
interface ModelOptions should have optional property `old`


ky-cheng:
According to https://vincit.github.io/objection.js/#modeloptions `old` object should be an optional attribute of `ModelOptions`. Happy to open a PR to add `old?: any;` to the ModelOptions interface.

koskimas:
Sure, I'd merge it :+1:

PR #1328, opened 2019-05-13
Add old property to ModelOptions interface

jbagio:
Closes https://github.com/Vincit/objection.js/issues/1251


koskimas:
Thanks!

PR #1332, opened 2019-05-14
Update typings for Union, Union All and Intersect

saumya1510:
Knex [unionAll](https://knexjs.org/#Builder-unionAll) and [intersect](https://github.com/tgriesser/knex/pull/3023) accept list of callbacks, builders or raw statements, same as union. I wanted to use the intersect method but realized it only accepted callback. So, updated the typings.


koskimas:
Thanks!

Issue #1342, opened 2019-05-21
typings objection, snakeCaseMappers & knexSnakeCaseMappers


cdupetit:
Hi,
In the definition file 'typings/objection/index.d.ts' : 
```
  const snakeCaseMappers: () => ColumnNameMappers;
  const knexSnakeCaseMappers: () => KnexMappers;

```
the options parameter is missing (see 'lib/utils/identifierMapping.js') : 
```
function knexSnakeCaseMappers(opt = {}) {}
function snakeCaseMappers(opt = {}) {}

```
The definition should be : 
```
  const snakeCaseMappers: (opt?: any) => ColumnNameMappers;
  const knexSnakeCaseMappers: (opt?: any) => KnexMappers;

```
Thanks.
Regards

koskimas:
PR merged

ps2goat:
When will this fix be released? It's not in 1.6.9, but it would be nice to have it.  Thanks for keeping the typing fixes coming!

PR #1343, opened 2019-05-21
snakeCaseMappers & knexSnakeCaseMappers parameters

cdupetit:
see #1342 issue

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -29,8 +29,13 @@ declare namespace Objection {
   const compose: Compose;
   const mixin: Mixin;
 
-  const snakeCaseMappers: (opt?: any) => ColumnNameMappers;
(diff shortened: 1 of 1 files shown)


typings/objection/index.d.ts
```diff
@@ -29,8 +29,8 @@ declare namespace Objection {
   const compose: Compose;
   const mixin: Mixin;
 
-  const snakeCaseMappers: () => ColumnNameMappers;
-  const knexSnakeCaseMappers: () => KnexMappers;
+  const snakeCaseMappers: (opt?: any) => ColumnNameMappers;
+  const knexSnakeCaseMappers: (opt?: any) => KnexMappers;
```
koskimas: The options should be typed. You should create an interface like `SnakeCaseMapperOptions` and use that as the argument.

cdupetit: Just done.

PR #1344, opened 2019-05-21
Consistently handle ModifierNotFoundError in relations

lehni:
@koskimas we talked about this today on Slack. This seems like the cleanest solution to me, as there are parts in the library that rely on being able to distinguish `ModifierNotFoundError`  from other `Error` objects.

### DIFF
--- a/lib/relations/Relation.js
+++ b/lib/relations/Relation.js
@@ -130,6 +130,10 @@ class Relation {
       builder.resolve([]);
     }
 
+    return this.applyModify(builder);
+  }
+
+  applyModify(builder) {
     try {
       return builder.modify(this.modify);
     } catch (err) {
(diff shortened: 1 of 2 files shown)


lib/relations/Relation.js
```diff
@@ -151,14 +155,14 @@ class Relation {
...
 
-    if (relatedSelect.isSelectAll()) {
+    if (relatedJoinSelect.isSelectAll()) {
       // No need to join a subquery if the query is `select * from "RelatedTable"`.
-      relatedSelect = aliasedTableName(relatedTable, relatedTableAlias);
+      relatedJoinSelect = aliasedTableName(relatedTable, relatedTableAlias);
     }
 
-    return builder[joinOperation](relatedSelect, join => {
+    return builder[joinOperation](relatedJoinSelect, join => {
```
lehni: @koskimas I've renamed these for consistency with the same code in `ManyToManyRelation`

PR #1362, opened 2019-06-04
Adds typings to orderBy to allow for multiple columns.

### LINKED ISSUE #1354: orderBy typing error when using array of columns
I am having an issue in TypeScript when using order by this way: `.orderBy([{ column: 'isConfigured', order: 'asc' }, { column: orderBy, order: ascDesc }]);` citing:

```
Argument of type '{ column: string; order: string; }[]' is not assignable to parameter of type 'ColumnRef'.
  Type '{ column: string; order: string; }[]' is not assignable to type 'string'.ts(2345)
```

The [Knex.js documentation](https://knexjs.org/#Builder-orderBy) says this should be possible. Am I missing something?

Thanks!

laginha87:
Closes #1354

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -886,6 +886,13 @@ const whereDelRetFirstWhere: Promise<Person | undefined> = qb
   .first()
   .where({ firstName: 'Mo' });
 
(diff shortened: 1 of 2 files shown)


koskimas:
Thanks!

PR #1380, opened 2019-06-18
Improve union/unionAll types

mordof:
### Description

Matching union/unionAll types to be closer to what the Objection/Knex api actually supports.

Previously, those methods on QueryBuilder didn't support supplying a QueryBuilder directly in the types. In plain Javascript, I've tested the calls itself, and union/unionAll support many arguments passed... each of which can be a QueryBuilder, or a callback. Also in this fashion, `wrap` argument can be provided last in the argument list as well.

I've included the capability to have up to 7 union arguments, and then an additional `wrap` boolean. Typescript doesn't allow having any amounts of arguments, then specifying the last item, unfortunately. the use of 7 amount of arguments is fairly common as far as I'm aware.

 #### Related issues

No related issues.

 ### Checklist

 - [x] Code conforms with the ESLint and Prettier rules (`npm eslint` passes
      and `npm prettier` has been applied)

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -170,6 +170,52 @@ async () => {
   takesMaybePerson(await Person.query().findOne({ lastName }));
 };
 
+// union/unionAll types
+
+async () => {
+  await Person.query().where({ lastName: 'finnigan' })
+    .union(
+      // supports callbacks, or querybuilders along-side each other.
+      Person.query().where({ lastName: 'doe' }),
+      (qb) => qb.table(Person.tableName).where({ lastName: 'black' }),
+    );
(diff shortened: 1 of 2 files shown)


mordof:
Updated description to match the template currently in a PR. Hopefully that's useful here - I didn't look over the template PR too closely.

mordof:
brought my forked repo up to date with master, then rebased my branch on master for two reasons:
1. to update it as there's been new commits
2. to trigger a new CI check, since it isn't my code's fault CI failed.

koskimas:
Thanks for this @mordof!

mordof:
You're welcome :)

PR #1391, opened 2019-06-26
Make graph options more strict

jure:
Verify that an option being passed is indeed `undefined`, a boolean or an array of strings.

Previously if you specified an option such as `noUpdate: '[files]'`, it would be gladly accepted and `!!`-ed. Being familiar with arrays within a single RelationExpression from e.g. `eager()` makes this likely to occur (e.g. https://github.com/elifesciences/elife-xpub/pull/2173/files#diff-5b518139d4766e3a75dc9c2cb1213c90R68), be hard to find and cause data loss.

The correct way to specify that you only don't want to update the `files` is `noUpdate: ['files']`, an array of strings.

This PR verifies that you're passing options of the correct type.

Perhaps there's a bigger type-checking for options discussion to be had, but this is what bit us, so I figured it would be an OK place to start. :)

### DIFF
--- a/lib/queryBuilder/graph/GraphOptions.js
+++ b/lib/queryBuilder/graph/GraphOptions.js
@@ -162,8 +162,14 @@ class GraphOptions {
 
     if (Array.isArray(option)) {
       return option.indexOf(node.relationPathKey) !== -1;
+    } else if (typeof option === 'boolean') {
+      return option;
+    } else if (option === undefined) {
+      return false;
     } else {
-      return !!option;
+      throw new Error(
+        `invalid option for ${optionName}: ${option}. please provide a boolean or array of strings.`
+      );
     }
   }
 }
--- a/lib/queryBuilder/graph/GraphOptions.js
+++ b/lib/queryBuilder/graph/GraphOptions.js
@@ -168,7 +168,7 @@ class GraphOptions {
       return false;
     } else {
       throw new Error(
-        `invalid option for ${optionName}: ${option}. please provide a boolean or array of strings.`
+        `expected ${optionName} option value "${option}" to be an instance of boolean or array of strings`
       );
     }
   }
--- a/tests/integration/upsertGraph.js
+++ b/tests/integration/upsertGraph.js
@@ -2526,6 +2526,29 @@ module.exports = session => {
             });
           });
         });
+
+        it('should throw a sensible error if an option with an invalid type is passed', done => {
+          Model1.bindKnex(session.knex)
+            .query()
+            .upsertGraph(
+              {
+                id: 1
+              },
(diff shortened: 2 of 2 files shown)


jure:
Pipeline is failing because it can't connect to the MySQL server:
```
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
The command "mysql -u root -e "CREATE USER objection"" failed and exited with 1 during .
```

jure:
Coveralls is still saying coverage dropped, but the line it's pointing to as missed is in a test, not in code. 🤷‍♂

jure:
Kind of difficult to search for examples of bugs out there due to this, but I gave it a shot anyway:

1. https://github.com/ashanren/webpack-generic/blob/b747e879757ffb1ba29d6402b8790704759d91fb/app/controllers/students.js#L63

2. https://github.com/astronomer/houston-api-legacy/blob/20205a7d4b6787b9f08e1a7ebb7b0cdcf998499e/src/services/workspace.js#L63

Worth to fix and address?

koskimas:
I think we can merge this! Thanks!

**About upsertGraph in general:**
I've noticed that people misuse and overuse it alot and I regret ever adding the feature. It goes against every principle of objection (query builder, you get what you write, easy to understand, explicit).

Alot of my time during the last year has gone to fixing weird corner case bugs in `upsertGraph` and I haven't had time to further develop the good features of objection. 

There's still a lot of work to be done with `upsertGraph` to make people avoid its pitfalls. The next thing I'm going to do is write a huge warning section in the docs where I discuss the numerous caveats in length to try to make people not use it for everything.

I'm happy to merge PR's that fix these things, but I won't be spending much time on `upsertGraph` myself, since I'm trying to get the 2.0 release together.

diversemix:
Thanks for the reply @koskimas ... My vote would certainly be to 'deprecate' this feature and move to a plugin. 
/cc @jure @malexsan1

koskimas:
@diversemix I don't know if we can actually remove it anymore at this point, even to extract is as a plugin. Many people have selected objection because of `upsertGraph` (which kind of breaks my heart :slightly_smiling_face:). 

**\<rant>**
It seems like a silver bullet on first glance. It seems to allow you to get the benefits of relational databases, with the simplicity of mongodb, but in reality, it combines the bad parts of both of them :smile: It's still useful for small things like updating a single relation with a parent object. 

A good example where I might still use it is updating the roles of a user with the user. In this case it's simple to understand and saves a lot of code, and also has the same performance. It all goes to shit with larger graphs, or when people don't even know what kind of graph they are passing in.
**\</rant>**

PR #1392, opened 2019-06-26
Add MySQL service to travis.yml

jure:
Recently MySQL isn't started by default in Travis and needs to be specified manually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,3 +33,6 @@ notifications:
 
 addons:
   postgresql: '9.5'
+
+services:
+  - mysql


jure:
Yeah, this was definitely it, MySQL isn't started by default anymore on Travis.

Node 8 and 9 are failing with an interesting single MySQL failure:

```
  3654 passing (3m)
  11 pending
  1 failing
  1) integration tests
       mysql
         GraphInsert tests
           many to many relation
             should insert an object using #ref
               should insert an object using #ref:
      AssertionError: expected 1000011 to equal 1000010
      + expected - actual
      -1000011
      +1000010
```

Otherwise, this seems good to go.

koskimas:
Thank you!

PR #1401, opened 2019-07-03
add mariadb connector

knoxcard:
add mariadb connector


elhigu:
mariadb driver is not supported by knex

knoxcard:
@rusher - knex is the next goal, I still need to test it out, everything should just fine, then I'll come back here and resubmit.

PR #1402, opened 2019-07-03
use spread operator over Object.assign consistently

christian-fei:
hello,

i was reading the doc and noticed that the spread operator was used more often than `Object.assign`.

converted two uses of `Object.assign` to `...`


koskimas:
Thanks!

Issue #1339, opened 2019-05-20
JoinRelation interface weakens type in method chain


Aksigera:
```
class Foo extends Model {
  prop!: number;
}

async function joinRelationDoesNotSupportChaining() {
  const foo1: Foo | undefined = await Foo.query().first();
  const foo2: Foo | undefined = await Foo.query().joinRelation('any_relation').first(); // won't compile
}
```
From `JoinRelation` interface signature `foo2` inferred type would be `Model | undefined`, but, if I'm correct, it's type would never be weaker than `Foo | undefined`.

So I suggest to change typings from 
```
  interface JoinRelation {
    <QM extends Model>(relationName: string, opt?: RelationOptions): QueryBuilder<QM, QM[]>;
  }
```
to
```
  interface JoinRelation {
    <TBase extends Model, QM extends TBase>(this:QueryBuilder<TBase>, relationName: string, opt?: RelationOptions): QueryBuilder<QM, QM[]>;
  }
```

mordof:
Ran across this. Unfortunately I don't have time to dig into it at the moment, but looking at the example change you've given @Aksigera I don't think you need `TBase`.

should work the same with:

```typescript
  interface JoinRelation {
    <QM extends Model>(this: QueryBuilder<QM>, relationName: string, opt?: RelationOptions): QueryBuilder<QM, QM[]>;
  }
```

As QM can then be derived directly from the QueryBuilder found in the `this` context.

Aksigera:
> Ran across this. Unfortunately I don't have time to dig into it at the moment, but looking at the example change you've given @Aksigera I don't think you need `TBase`.
> 
> As QM can then be derived directly from the QueryBuilder found in the `this` context.

I don't expect `joinRelation()` to return type of `this`, that's why I added `TBase`.
In your code, output of function is of the same type as, well, input. But **join** means expanding of type, so we have such hierarchy `Model <= TBase <= QM`, where `<=` means smaller or equal.

TBase can be changed to anything, I'm not familiar with naming conventions on this project

mordof:
adding a note here that this affects the other similar functions `rightJoinRelation`, etc. Going to tackle this issue this week

koskimas:
Fixed in 2.0

PR #1409, opened 2019-07-15
joinRelation types no longer weakend

mordof:
### Description

`joinRelation` and other similar methods returned a queryBuilder type with just `Model`. This fix preserves the actual base model the query is being made on.

e.g. `Person.query().joinRelation('children')` will return `QueryBuilder<Person, Person[], Person[]>` now (Person class in test types used for this example).

As much as I'd like to be able to actually strengthen the types returned (e.g.  Person returned with a guaranteed `children` prop as `Person[]` not optional), the style of using a string with its own language as the argument for eager/relational calls prevents typescript from being able to infer this at this time.

 #### Related issues

 - Related to https://github.com/Vincit/objection.js/issues/1339

 ### Checklist

- [x] Code conforms with the ESLint and Prettier rules (`npm eslint` passes and `npm prettier` has been applied)

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -134,6 +134,13 @@ const lastName = 'Lawrence';
 
 takesPersonQueryBuilder(Person.query());
 
+// ensure `joinRelation` and friends returns the correct QueryBuilder types
+takesPersonQueryBuilder(
+  Person.query()
+    .where({ lastName })
+    .joinRelation('children')
+);
+
 async () => {
   takesPeople(await Person.query().where('lastName', lastName));
   takesPeople(await Person.query().where({ lastName }));
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -347,7 +347,11 @@ declare namespace Objection {
   }
 
   interface JoinRelation {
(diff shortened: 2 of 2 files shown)


mordof:
@koskimas  Should I be leaving my related ticket references just as "Related to", or would you rather I have them as "resolves" or "fixes" so that github automatically closes the tickets on merge?

capaj:
@mordof objection does allow to pass an object to relations: https://vincit.github.io/objection.js/api/types/#relationexpression-object-notation

I believe we could strengthen the type if the object notation is passed. The only problem is my Typescript skills probably are not up to it yet. What about yours?

mordof:
@capaj  that's a good point. In those cases, the types can be strengthened. I think that'd be worth an improvement ticket, as I want to focus on the more critical type bugs for now.

The general idea behind how to do that is to check that the argument passed is an object, and that the keys extend keyof the model in the query builder. Then once those keys pass the checks of belonging to the model in question, you can have the return type query builder take a type of the original model, then use a union on the keys from the object to remove the optional modifier.

koskimas:
Thanks for this. These changes have been incorporated in the new v2.0 typings. Since 2.0 is just around the corner, I think we can close this.

Issue #856, opened 2018-03-27
Problem with $relatedQuery typings


chamini2:
If I have:

```ts

class One extends Model {
  public static get relationMappings(): RelationMappings {
    return {
      twos: {
        relation: Model.HasManyRelation,
        modelClass: Two,
        join: { from: 'ones.id', to: 'twos.one_id' }
      }
    };
  }

  public twos: Two[];
}

class Two extends Model {
  id: number;
}

function a() {
  const one = new One();
  one.$relatedQuery('twos')
  .insert([{ id: 1 }]);
}

```

It will throw at compile the following:

```
error TS2345: Argument of type '{ id: number; }[]' is not assignable to parameter of type 'Partial<Two[] & Model> | undefined'.
  Type '{ id: number; }[]' is not assignable to type 'Partial<Two[] & Model>'.
    Types of property 'push' are incompatible.
      Type '(...items: { id: number; }[]) => number' is not assignable to type '((...items: Two[]) => number) | undefined'.
        Type '(...items: { id: number; }[]) => number' is not assignable to type '(...items: Two[]) => number'.
          Types of parameters 'items' and 'items' are incompatible.
            Type 'Two' is not assignable to type '{ id: number; }'.
              Property '$id' does not exist on type '{ id: number; }'.
```

I think this problem was introduced in https://github.com/Vincit/objection.js/pull/773:

> QM now always extends Model. I believe this is correct, right? Can you query on non-models?

chamini2:
It does type check, although it still looks like a hack.

douggr:
Actually, `public twos: Two[];` should be `public twos: Two`. It's handled it internally somehow.

```ts
export default class User extends Model {
  public static readonly relationMappings = {
    emails: {
      join: {
        from: "user.id",
        to: "email.user_id",
      },
      modelClass: Email,
      relation: Model.HasManyRelation,
    },
  };

  public emails?: Email;
}
```

```ts
  User.query().findById(1).then((r) => {
    if (!r) {
      throw new Error("Not Found");
    }

    return r.$relatedQuery("emails")
      .insert([
        { email: Math.random().toString(36).substring(7) },
        { email: Math.random().toString(36).substring(7) },
        { email: Math.random().toString(36).substring(7) },
      ])
      .returning("*")
      .then((rs) => res.jsonp(rs));
  });

  // and loading... 
    User.query()
    .findById(1)
    .eager("emails")
    .then((r) => res.jsonp(r));

```

[...]

PR #1410, opened 2019-07-15
Types require less casting, as well as handles invalid (existing) props passed error

mordof:
### Description

The types for `$relatedQuery` (without the force cast) didn't handle type inference very well, also returning a QueryBuilder with types that made chaining subsequent calls problematic/error incorrectly.

With the changes in this PR, `$relatedQuery` (without force cast) no longer has complications for subsequent chained calls. It returns the correct QueryBuilder by using the property name, grabbing that property type from the class defined (supports property types being an array of models or singular), and ensures that type extends `Model`. This significantly reduces the situations needing force-casting for `$relatedQuery`.

A new addition to the types here is if a relation name is passed, which exists as a property on the model, but does not actually extend `Model`, `$relatedQuery` will return `never`. This way the types will alert the developer when they've specified something which isn't query-able.

This PR has two assumptions involved:
 - Any relation available for `$relatedQuery` will also be included in the types of the model class itself.
 - The type of the relation specified will have extended `Model`.

If any of the assumptions are undesirable by the developer, they can use the force-cast version of `$relatedQuery`. e.g: `await child.$relatedQuery<Person>('parent');` (though this case would work perfectly fine without the force cast unless types aren't defined).

 #### Related issues

 - Related to https://github.com/Vincit/objection.js/issues/856

 ### Checklist

- [x] Code conforms with the ESLint and Prettier rules (`npm eslint` passes and `npm prettier` has been applied)

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -58,7 +58,7 @@ class Person extends objection.Model {
   }
 
   fetchMom(): Promise<Person | undefined> {
-    return this.$relatedQuery<Person>('mom').first();
+    return this.$relatedQuery('mom').first();
   }
 
   async $beforeInsert(queryContext: objection.QueryContext) {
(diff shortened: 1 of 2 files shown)


koskimas:
Thanks for this. These changes have been incorporated in the new v2.0 typings. Since 2.0 is just around the corner, I think we can close this.

PR #1414, opened 2019-07-23
Add total number of pages to .page() return object

o8e:
This is pretty useful considering I have to manually calculate this each time I use the paging operation at the moment.


koskimas:
I don't think this needs to be added by objection. Also the implementation is broken in all possible ways.

PR #1423, opened 2019-07-27
Add missing typings

willsoto:
I wanted to get this PR open so there is visibility into the work being done. The new typings seem much cleaner than the old ones so I wanted to make sure I kept that cleanliness. _Please let me know if this is looking good so far._ If so, I will continue to work on it until `npm run test-typings` passes (or I get tired 😄 )

So far I've added support for:
- `whereBetween`
- `whereJsonSupersetOf`
- `whereJsonIsArray`
- `count`
- `static relatedQuery`
- `transaction`
- `relate`

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -436,6 +436,16 @@ declare namespace Objection {
     (insert: PartialModelObject<ModelType<QB>>[]): ArrayQueryBuilder<QB>;
   }
 
+  interface RelateMethod<QB extends AnyQueryBuilder> {
+    <RelatedModel extends Model>(
+      ids: MaybeCompositeId | Partial<RelatedModel> | Partial<RelatedModel>[]
+    ): SingleQueryBuilder<QB>;
+
+    <RelatedModel extends Model>(
+      ids: MaybeCompositeId | Partial<RelatedModel> | Partial<RelatedModel>[]
+    ): ArrayQueryBuilder<QB>;
+  }
+
   interface EagerMethod<QB extends AnyQueryBuilder> {
     (expr: RelationExpression<ModelType<QB>>, modifiers?: Modifiers): QB;
   }
@@ -699,6 +709,8 @@ declare namespace Objection {
 
     insert: InsertMethod<this>;
     insertAndFetch: InsertMethod<this>;
+
+    relate: RelateMethod<this>;
 
     eager: EagerMethod<this>;
     mergeEager: EagerMethod<this>;
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -221,6 +221,13 @@ declare namespace Objection {
     <M extends Model>(modelClass: ModelClass<M>): M['QueryBuilderType'];
   }
 
+  /**
+   * https://vincit.github.io/objection.js/api/types/#type-fieldexpression
+   */
+  type FieldExpression = string;
+
+  type JsonObjectOrFieldExpression = object | object[] | FieldExpression;
+
   type Selection<QB extends AnyQueryBuilder> = ColumnRef | AnyQueryBuilder | CallbackVoid<QB>;
 
   interface SelectMethod<QB extends AnyQueryBuilder> {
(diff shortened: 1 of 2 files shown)


koskimas:
Awesome! Thank you! :+1:

typings/objection/index.d.ts
```diff
@@ -352,6 +374,10 @@ declare namespace Objection {
     (column: string, amount?: number): QB;
   }
 
+  interface CountMethod<QB extends AnyQueryBuilder> {
+    (): QB;
```
koskimas: `count` can also take a `ColumnRef` or an array of them. http://knexjs.org/#Builder-count

koskimas: And apparently all kinds of other stuff I wasn't even aware of :smile:

koskimas:
@mordof You should also be aware of this PR

typings/objection/index.d.ts
```diff
@@ -352,6 +374,12 @@ declare namespace Objection {
     (column: string, amount?: number): QB;
   }
 
+  interface CountMethod<QB extends AnyQueryBuilder> {
+    (column?: ColumnRef, options?: { as: string }): QB;
+    (aliasToColumnDict: { [alias: string]: string | string[] }): QB;
+    (...columns: string[]): QB;
```
koskimas: I think this can also be `ColumnRef[]`

willsoto:
@koskimas Let me know if you'd like to merge this at any point so others can work on this without having to daisy-chain PRs. Just ping me so I can clean up the commits.

koskimas:
I believe @mordof is the only other person that will likely work on the new typings. Let's see what he says. We can definitely merge smaller pieces anyway. Just let me know when you want me to merge.

mordof:
@koskimas I haven't verified correctness of the types themselves added, but this doesn't overlap with anything I'm working on at this point.

Do you want me to verify correctness of the types specifically before this goes through as well? If you're comfortable with what's there, I'm ok with it going in.

koskimas:
@mordof I can take care of the PR reviews.

willsoto:
@koskimas I've add support for transactions and cleaned up my commits. According to `npm run test-typings` the last thing that needs to be typed is `relate` - I can make a follow-up PR for that since I don't think it will be too difficult.

Let me know if you have any feedback and I can take care of it. Otherwise, feel free to merge if you think it's good.

typings/objection/index.d.ts
```diff
@@ -1024,4 +1063,91 @@ declare namespace Objection {
...
+  /**
+   * Overloading is required here until the following issues (at least) are resolved:
+   *
+   * - https://github.com/microsoft/TypeScript/issues/1360
+   * - https://github.com/Microsoft/TypeScript/issues/5453
+   *
+   * @tutorial https://vincit.github.io/objection.js/guide/transactions.html#creating-a-transaction
+   */
+  export interface transaction<T> {
+    start(knexOrModel: knex | ModelClassSpecifier): Promise<Transaction>;
```
koskimas: `ModelClassSpecifier` is not allowed here. Only a model class is. You can copy paste most of this from the old typings.

willsoto: Fixed. One question I had was around the use of `V` - I can't tell what it is by the name but I assume it represents the type of the return value. Would you mind if I renamed it to something more meaningful? Maybe `ReturnValue`?

willsoto:
I added support for `relate` as well. It was pretty straightforward I hope...

typings/objection/index.d.ts
koskimas: By all means. Descriptive names are always better. I only use short type names `M` and `QB` so that they don't bloat the code.

willsoto:
@koskimas This should be good to go unless you have any more comments.

typings/objection/index.d.ts
```diff
@@ -1024,4 +1075,95 @@ declare namespace Objection {
...
+
+  /**
+   * Overloading is required here until the following issues (at least) are resolved:
+   *
+   * - https://github.com/microsoft/TypeScript/issues/1360
+   * - https://github.com/Microsoft/TypeScript/issues/5453
+   *
+   * @tutorial https://vincit.github.io/objection.js/guide/transactions.html#creating-a-transaction
+   */
+  export interface transaction<T> {
```
koskimas: The `T` type is unused and can be removed.

typings/objection/index.d.ts
```diff
@@ -1024,4 +1075,95 @@ declare namespace Objection {
...
+        boundModelClass1: MC1,
+        boundModelClass2: MC2,
+        boundModelClass3: MC3,
+        boundModelClass4: MC4,
+        boundModelClass5: MC5,
+        trx?: Transaction
+      ) => Promise<ReturnValue>
+    ): Promise<ReturnValue>;
+
+    <V>(knex: knex, callback: (trx: Transaction) => Promise<V>): Promise<V>;
```
koskimas: Shouldn't `V` be `ReturnValue` here too?

typings/objection/index.d.ts
willsoto: Fixed

typings/objection/index.d.ts
willsoto: Yup! Fixed

PR #1425, opened 2019-07-30
Separate select and delete queries for m2m unrelate operations

falkenhawk:
This is an attempt on working around ER_CANT_UPDATE_USED_TABLE_IN_SF_OR_TRG mysql error when executing a db trigger on a join table in m2m relation

- extract a subquery reading related ids to separate query run before the delete query for m2m unrelate operation


falkenhawk:
oh god. I was supposed to set up this PR on my fork only, not here 😓 I'm sorry.

Issue #1356, opened 2019-05-29
Add filter method on ManyToMany join table.


keepcosmos:
Is there any way to add `filter` options on ManyToMany `through`?

like,

```js
goodFriends: {
  relation: ManyToManyRelation,
  to: XX,
  from: YY,
  through: {
    from: 'XX.id'
    to: 'YY.id',
    filter: (builder) => { builder.where('relationships.type', 'good') }
  }
}
```

koskimas:
Not currently, but that could be useful addition. It could be useful solution to https://github.com/Vincit/objection.js/issues/837

ericmuyser:
This would allow `N:M` polymorphic relationships like in this answer: https://github.com/Vincit/objection.js/issues/19#issuecomment-389356230

But instead of the filter field being on the relation table, it would be on the join table, which is freaking awesome (and how I thought it worked until I kept running into this error and analyzed that answer further).

I have a generic exclusive arc table that allows me to build relationships on the fly. I put the `extra: ['relationKey']` on that table thinking it would work. It doesn't. So I'll need a `relationKey` on every single table. Sigh.

![image](https://user-images.githubusercontent.com/156768/59824461-1df72d00-936c-11e9-9a9b-001cc807d0eb.png)

ericmuyser:
I figured out how to the querying part, but I still need to set a value on the pivot table column for querying later. (Mentioned in #1405)

The problem is beforeInsert isn't executed below, because it's not inserting a new profile. But it is inserting a new pivot table row for the relation, which is where I need to set a column value.


```
    public static get relationMappings (): RelationMappings {
        return {
            owner: {
                relation: Model.HasOneThroughRelation,
                modelClass: Profile,
                filter: (qb) => {
                    qb.join('nodes as node_profiles', 'node_profiles.toProfileId', '=', 'profiles.id')
                    qb.where('node_profiles.relationKey', 'owner')
                },
                beforeInsert (model) {
                    (model as Node).relationKey = 'owner'
                },
                join: {
                    from: 'collections.id',
                    to: 'profiles.id',
                    through: {
                        from: 'nodes.fromCollectionId',
                        to: 'nodes.toProfileId'
                    }
                }
            }
        }
    }
```

lexa0303:
Any updates on this?

haydernativapps:
Any update about this issue?

sayhicoelho:
Any update? This feature is needed...

bbrizzi:
[...]

PR #1428, opened 2019-08-01
filters/modify on join table of m2m relations

falkenhawk:
addressing https://github.com/Vincit/objection.js/issues/1356

Allow specifying `modify` / `filters` for the join table in ManyToManyRelations.


falkenhawk:
@koskimas would you consider this for objection v3.0? I could try to rebase and add tests, but not sure if I could make it on time, depends when you would like to move out of v3 alpha stage

falkenhawk:
@koskimas I've added some tests and adjusted docs.

lib/queryBuilder/QueryBuilder.js
```diff
@@ -325,7 +325,7 @@ class QueryBuilder extends QueryBuilderBase {
         );
       } else if (op.constructor === FromOperation) {
         return op.table === tableName;
-      } else if (op.name === 'as' || op.is(FindOperation)) {
+      } else if (op.name === 'as' || op.is(FindOperation) || op.is(OnErrorOperation)) {
```
falkenhawk, line 328: I have no recollection though why I added this.

lehni: It's not wrong: If you have a query that only contains an `onError()` operation, it's still a select-all query.
As for why it was added, I'd guess you had some logic in your code base that always calls `onError()` o your queries and wanted to filter this out? So I do think this makes sense, but it's not really in the scope of this PR to add it?

lehni:
LGTM, thanks @falkenhawk!

As for the `OnErrorOperation` bit, I think it's not wrong to leave this in. Merging.

developer239:
The new API is not doing anything and both `filter` and `modify` are ignored.

Issue #1435, opened 2019-08-06
Wrong return type for `findById` passed with composite key


ronyclau:
# Type of issue
Bug in TS type definition

# Description / Steps to reproduce
In the TS type definition of `QueryBuilderBase`, only these two [overloads of `findById`](https://github.com/Vincit/objection.js/blob/master/typings/objection/index.d.ts#L734-L735) are present:

```typescript
{
  // ...
  findById(id: Id): QueryBuilderYieldingOneOrNone<QM>;
  findById(idOrIds: IdOrIds): this;
  // ...
}
```

Let's say a model has a composite primary key, the following use case would trigger a compile-time error due to the incorrect overload:
```typescript
class SomeModel extends objection.Model {
  static idColumn = ['idField1', 'idField2']

  // ...
}

SomeModel.query()
  .findById([1, 100])
  .then((model: SomeModel | undefined) => {
    //   ^^^^^ error TS2740
  });
```

In the above case, the only eligible overload is
```typescript
findById(idOrIds: IdOrIds): this;
```

`await`-ing / `.then()` on the return type of this overload only yields `SomeModel[]`.

# Expected result
The code should compile.

# Actual result
```
error TS2740: Type 'SomeModel[]' is missing the following properties from type 'SomeModel': id, type, articles, $beforeInsert, and 31 more.
```

koskimas:
Thanks! PR is merged now.

PR #1436, opened 2019-08-06
fix findById return type for composite key

ronyclau:
### Description
This problem lies in the TypeScript type definition only.

`QueryBuilder`'s method [`findById`](https://github.com/Vincit/objection.js/blob/master/typings/objection/index.d.ts#L734-L735) currently accepts both an [`Id`](https://github.com/Vincit/objection.js/blob/master/typings/objection/index.d.ts#L339) (for single-field key) and an [`Ids`](https://github.com/Vincit/objection.js/blob/master/typings/objection/index.d.ts#L341) (for composite key).

Both versions return `this` for method chaining, however, only the "single-field key" version is overloaded to return a `QueryBuilderYieldingOneOrNone<QM>`. Such overloading breaks this use case for composite key:

```typescript
class SomeModel extends objection.Model {
  static idColumn = ['idField1', 'idField2'];
  // ...
}

function someFunctionNeedsModel(model: SomeModel | undefined) {
  // do something
}

// With .then()
SomeModel.query()
  .findById([1, 1])
  .then(model => {
    someFunctionNeedsModel(model);
  });

// With Promise
someFunctionNeedsModel(await SomeModel.query().findById([1, 1]));
```

`findById` should always return a `QueryBuilderYieldingOneOrNone<QM>`.

This PR fixes the overload, adds related test cases, and fixes an existing test case which incorrectly uses `findById` (seems like an unfortunate typo that compiles).

#### Related issues

Relates to #1435

### Checklist

- [x] Code conforms with the ESLint and Prettier rules (`npm eslint` passes
      and `npm prettier` has been applied)

### DIFF
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -426,7 +426,7 @@ async () => {
 
 const maybePerson: Promise<Person | undefined> = qb.findById(1);
 
-const maybePeople: Promise<Person[]> = qb.findById([1, 2, 3]);
+const maybePeople: Promise<Person[]> = qb.findByIds([1, 2, 3]);
 
 // query builder knex-wrapping methods:
 
@@ -1032,4 +1032,33 @@ const plugin2 = ({} as any) as objection.Plugin;
   takesModelClass(MyPerson);
   takesPersonClass(MyPerson);
   takesModelSubclass(new MyPerson());
+};
+
+// Examples with composite key
+class Interview extends objection.Model {
+  interviewer!: Person;
+  interviewee!: Person;
+  interviewDate?: number;
+
+  static columnNameMappers = objection.snakeCaseMappers();
+
+  static idColumn = ['interviewer', 'interviewee'];
+}
+
+async () => {
+  // findById with composite key
+  const interview: Interview | undefined = await Interview.query().findById([10, 11]);
+
+  // findById with composite key, chained with other query builder methods
+  const interviewWithAssignedDate: Interview | undefined = await Interview.query()
+    .findById([10, 11])
+    .whereNotNull('interviewDate');
+
+  // findByIds with sets of composite key
+  const interviews: Interview[] = await Interview.query().findByIds([[10, 11], [11, 12], [12, 13]]);
+
+  // findByIds with sets of composite key, chained with other query builder methods
+  const interviewsWithAssignedDate: Interview[] = await Interview.query()
+    .findByIds([[10, 11], [11, 12], [12, 13]])
+    .whereNotNull('interviewDate');
 };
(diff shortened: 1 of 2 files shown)


mordof:
That's a really strange error on the Node 6 build...

tests/ts/examples.ts
```diff
@@ -1033,3 +1033,32 @@ const plugin2 = ({} as any) as objection.Plugin;
   takesPersonClass(MyPerson);
   takesModelSubclass(new MyPerson());
 };
+
+// Examples with composite key
+class Interview extends objection.Model {
+  interviewer!: Person;
```
mordof: technically doesn't affect the typescript, but wouldn't these need to be numbers?

mordof: That's also an interesting thought: `findById(s)` types for the ID should probably be enforced to match the `idColumn` property of the class.

ronyclau: > technically doesn't affect the typescript, but wouldn't these need to be numbers?

Right. Should've defined a relationship for this. I was just trying to utilize the existing model class.

> That's also an interesting thought: `findById(s)` types for the ID should probably be enforced to match the `idColumn` property of the class. (unrelated to this PR)

Tried, but didn't manage to.

When you retrieve a `QueryBuilder` from a model (`SomeModel.query()`), only the `Model` type is passed as the type parameter, instead of the `ModelClass` type.

https://github.com/Vincit/objection.js/blob/9669e9617bb3e0db7162724cbd20b917affd5f20/typings/objection/index.d.ts#L448

The case is the same when you construct a `QueryBuilder` with its static method `forClass`:

https://github.com/Vincit/objection.js/blob/9669e9617bb3e0db7162724cbd20b917affd5f20/typings/objection/index.d.ts#L659-L661

It seems that it is impossible to get the actual `ModelClass`, when what we have is just a generic type `QM in <QM extends Model>`.

So the problem is `QueryBuilder` accepts a `Model` instead of a `ModelClass`. However, if we change this requirement, I don't know whether it will break any existing code. Didn't try that far though.

Meanwhile, it is impossible to access the static property of a generic type parameter by lookup types. So dead end here.

ronyclau:
Completely clueless on the Node 6 build error. I am guessing the error will be gone when rebuilt.

koskimas:
@ronyclau Thanks for this PR! We are working on completely new typings in the `new-typings` branch and this is actually already fixed there. The new types won't be released until 2.0 though, so I'll merge this and release a new patch version of objection soon.

tests/ts/examples.ts
koskimas: No need to try to match the `idColumn`. That would be insane :smile:

ronyclau:
@mordof @koskimas Thank you for your review! :)

Issue #1431, opened 2019-08-03
Organizing TypeScript tests


willsoto:
As I was working on adding more types, I felt like the current `example.ts` "dumping ground" is not really ideal for ensuring that the types are correct and that there are as few regressions as possible.

I would like to spend some time refactoring that to have better visibility into:

- Existing test coverage with TS
- What hasn't been tested
- Example usage that TS users can reference when they run into issues (even potentially as a guide to best practices regarding Custom Query Builders, etc)

Here is the quick view of what I have in mind:

```
├── custom-query-builder
├── joins
├── mixins
├── query-examples
│   ├── basic-queries
│   │   ├── delete.ts
│   │   ├── find.ts
│   │   ├── insert.ts
│   │   └── update.ts
│   └── relation-queries
├── transactions
└── where
```

I see it as a bit of a mix between going through the docs and ensuring that at least all the documented cases are covered while also making sure the basic operators are covered (all the possible ways to join, all the possible where clauses, etc)

Does this seem reasonable? Very much open to feedback on this.

cc/ @mordof

koskimas:
Makes a lot of sense! Something like [dtslint](https://github.com/microsoft/dtslint) could also be useful.

mordof:
I very much agree with how you feel about the current state of the types as well.

About the only thing I'd like to pick on from what you said is this:
> Existing test coverage with TS

This could mean two things, and I want to confirm...
does this mean that you want to change the normal javascript tests over to being typescript?
or does this mean you want to clarify the typescript type tests  in a clean way that's not just in the example file.

If it's the latter, I'm also on board with that. If it's the former, I don't think that's worth working on at this point in time.

mordof:
A part about the examples/documentation... I'm wondering, can we have that in a different section compared to the typescript "tests", a location that we can actually include in official documentation generation for the page? Having a single root folder for everything typescript feels a bit yucky if we're handling tests, and example cases separated

willsoto:
Yes, it's the latter. I'd like to clarify the type tests in a clean way so we can visually grep what is covered.

> A part about the examples/documentation... I'm wondering, can we have that in a different section compared to the typescript "tests", a location that we can actually include in official documentation generation for the page?

[...]

PR #1437, opened 2019-08-06
Initial refactor of TS tests

willsoto:
Opening for visibility and feedback. Already noticing some gaps in coverage so I'll be trying to fix those as I work through these.

Ref #1431

### DIFF
--- a/tests/ts/query-examples/basic-queries/delete.ts
+++ b/tests/ts/query-examples/basic-queries/delete.ts
@@ -1,0 +1,28 @@
+import * as objection from '../../../../';
+import { raw } from '../../../../';
+
+class Person extends objection.Model {}
+
+(async () => {
+  await Person.query().deleteById(1);
+
+  const numDeleted: number = await Person.query()
+    .delete()
+    .where(raw('lower("firstName")'), 'like', '%ennif%');
+
+  console.log(numDeleted, 'people were deleted');
+
+  await Person.query()
+    .delete()
+    .whereIn(
+      'id',
+      Person.query()
+        .select('persons.id')
+        .joinRelation('pets')
+        .where('pets.name', 'Fluffy')
+    );
+
+  await Person.query()
+    .delete()
+    .whereExists(Person.relatedQuery('pets').where('pets.name', 'Fluffy'));
+})();
--- a/tests/ts/custom-query-builder.ts
+++ b/tests/ts/custom-query-builder.ts
@@ -1,9 +1,9 @@
 import { Model, QueryBuilder } from '../../';
 
 class CustomQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<M, R> {
-  ArrayQueryBuilderType: CustomQueryBuilder<M, M[]>;
-  SingleQueryBuilderType: CustomQueryBuilder<M, M>;
-  NumberQueryBuilderType: CustomQueryBuilder<M, number>;
+  ArrayQueryBuilderType!: CustomQueryBuilder<M, M[]>;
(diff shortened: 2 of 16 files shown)


tests/ts/query-examples/basic-queries/delete.ts
```diff
@@ -0,0 +1,28 @@
...
+class Person extends objection.Model {}
+
+(async () => {
+  await Person.query().deleteById(1);
+
+  const numDeleted: number = await Person.query()
+    .delete()
+    .where(raw('lower("firstName")'), 'like', '%ennif%');
+
+  console.log(numDeleted, 'people were deleted');
```
mordof: i don't think these will ever actually be run, will they?

mordof: hm i guess it doesn't matter if these files are purely for giving some examples

mordof:
There's already an `examples` folder. can we move these new examples into `examples/ts`?

koskimas:
@willsoto The idea here seems to be to copy the examples from the documentation? They are in no way comprehensive, but I guess that's a good place to start. I'd rather have real unit tests like `where.ts` that tests all the possible ways you can call `where/orWhere/andWhere/etc.`.

tests/ts/query-examples/basic-queries/delete.ts
willsoto: Yeah. For right now I'm just blindly copy-pasting from the example. I don't mind either way on these but they don't add any value, so might just remove.

willsoto: Yeah. For right now I'm just blindly copy-pasting from the example. I don't mind either way on these but they don't add any value, so might just remove.

willsoto:
> There's already an examples folder. can we move these new examples into examples/ts?

I see purpose of the tests as different from what is happening in the `examples` folders. The `examples` are more "end-to-end" what does it look like to work with Objection whereas the tests are just going through the many possible forms of queries, models, etc.

With that said, I took a glance at the `examples` folder and it looks like it could maybe be updated.

> The idea here seems to be to copy the examples from the documentation? They are in no way comprehensive, but I guess that's a good place to start. I'd rather have real unit tests like `where.ts` that tests all the possible ways you can call `where/orWhere/andWhere/etc.`.

That is right. And it is just a starting point. The kind of coverage you are describing is exactly where I want to go with this. I will also try and move what makes sense from the existing "tests" to the appropriate locations since we know those are battle-tested.

willsoto:
I haven't forgotten about this. Been busy with other things but I'll have some time to get back to this soon.

koskimas:
I merged new-typings with v2.0. You can point this to v2.0 branch now.

willsoto:
I've rebased and added tests for:

- [Basic Queries](https://vincit.github.io/objection.js/guide/query-examples.html#basic-queries)
- [Relation Queries](https://vincit.github.io/objection.js/guide/query-examples.html#relation-queries)

I am not sure how you prefer to proceed. It might be good to merge these PRs in little by little to prevent something that is 1000+ lines long. Happy to do either, but ultimately your call.

If you feel like this is good enough to merge, please go ahead and do so and I'll make follow-up PRs.

PS: I am not sure why tests are failing only on Node 12. I can take a look at fixing unless you already have a fix in mind?

willsoto:
@koskimas pinging to get your thoughts on my previous comment

koskimas:
We can merge this as soon as the tests pass

willsoto:
@koskimas looks like tests are passing now 👍

PR #1451, opened 2019-08-28
Update aggregation method types.

mastermatt:
Uses the method interface now in Knex, as of 0.17.3.


koskimas:
All those `TypePreservingAggregation` return a knex query builder and not the objection query builder. Unfortunately we cannot use that type directly. A similar type must be implemented for objection that uses objection query builder.

The typings have also been completely rewritten in v2.0 branch, so I don't know if it's worth it to work more on the v1.0 typings. v1.0 will get critical bug fixes after 2.0 has been releasesd, but I don't know if this qualifies as one.

mastermatt:
Ah and I see 2.0 already has `AggregateMethod`.... although it doesn't seem to allow the form of passing in an object to alias columns.

I'll just close this and play around with the 2.0 branch. 
Thanks @koskimas

PR #1453, opened 2019-08-29
fix(typings): throwIfNotFound() yields one when used with first(), not one or none

sammarks:
I noticed when converting my project to TypeScript that if you call `.throwIfNotFound()` with something like `.first()`, it still thinks the result may not be undefined, even though we'll be throwing an error if it is not defined.

I noticed you already had `QueryBuilderYieldingOneOrNone` associated with `.first()`, so I just updated its interface to include an override for `throwIfNotFound()` to return `QueryBuilderYieldingOne` instead of just `this`

Looks like my IDE also found some extra whitespace! Let me know if you want that removed and I'll remove it :)


sammarks:
Nevermind, looks like this is already the case!

PR #1483, opened 2019-09-24
Additional TypeScript tests

willsoto:
- Add the rest of the finder methods
- Add mutating methods

#1431

### DIFF
--- a/tests/ts/query-builder-api/mutating-methods.ts
+++ b/tests/ts/query-builder-api/mutating-methods.ts
@@ -1,0 +1,205 @@
+import { raw, ref } from '../../../';
+import { Movie } from '../fixtures/movie';
+import { Person } from '../fixtures/person';
+
+(async () => {
+  await Person.query().insert({ firstName: 'Jennifer', lastName: 'Lawrence' });
+
+  await Movie.relatedQuery('actors')
+    .for(1)
+    .insert([
+      { firstName: 'Jennifer', lastName: 'Lawrence' },
+      { firstName: 'Bradley', lastName: 'Cooper' }
+    ]);
+
+  await Person.query().insert({
+    age: Person.query().avg('age'),
+    firstName: raw("'Jenni' || 'fer'")
+  });
+
+  await Person.query()
+    .insert({
+      age: Person.query().avg('age'),
+      firstName: raw("'Jenni' || 'fer'")
+    })
+    .returning('*');
+
+  await Movie.relatedQuery('actors')
+    .for(1)
+    .insert({
+      firstName: 'Jennifer',
+      lastName: 'Lawrence',
+      someExtra: "I'll be written to the join table"
+    });
+
+  await Person.query().insertAndFetch({ firstName: 'Jennifer', lastName: 'Lawrence' });
+
+  await Movie.relatedQuery('actors')
+    .for(1)
+    .insertAndFetch([
+      { firstName: 'Jennifer', lastName: 'Lawrence' },
+      { firstName: 'Bradley', lastName: 'Cooper' }
+    ]);
+
+  await Person.query().insertAndFetch({
+    age: Person.query().avg('age'),
+    firstName: raw("'Jenni' || 'fer'")
+  });
+
+  await Movie.relatedQuery('actors')
+    .for(1)
+    .insertAndFetch({
+      firstName: 'Jennifer',
+      lastName: 'Lawrence',
+      someExtra: "I'll be written to the join table"
+    });
+
+  await Person.query().insertGraphAndFetch({ firstName: 'Jennifer', lastName: 'Lawrence' });
+
+  await Person.query()
+    .patch({ age: 24 })
+    .findById(1);
+
+  await Person.query()
+    .patch({ age: 20 })
+    .where('age', '<', 50);
+
+  await Person.query()
+    .patch({ age: raw('age + 1') })
+    .where('age', '<', 50);
+
+  await Person.query().patch({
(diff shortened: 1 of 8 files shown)


tests/ts/query-builder-api/mutating-methods.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  const numUnrelatedRows = await person
+    .$relatedQuery('movies')
+    .unrelate()
+    .where('id', 50);
+
+  await Person.query().increment('age', 1);
+  await Person.query().decrement('age', 1);
+
+  await Person.query().truncate();
```
willsoto: Currently no typings for `truncate` exist. Should I add? Is it chained off of `query`?

typings/objection/index.d.ts
```diff
@@ -562,10 +563,6 @@ declare namespace Objection {
     <RelatedModel extends Model>(
       ids: MaybeCompositeId | Partial<RelatedModel> | Partial<RelatedModel>[]
     ): NumberQueryBuilder<QB>;
-
-    <RelatedModel extends Model>(
```
willsoto: This was a total duplicate of the above types. Unsure if copy-pasta mistake, or intended to be tweaked.

typings/objection/index.d.ts
```diff
@@ -391,6 +391,7 @@ declare namespace Objection {
   }
 
   interface WhereCompositeMethod<QB extends AnyQueryBuilder> {
+    (column: ColumnRef[], op: Operator, value: Value[]): QB;
```
willsoto: Pretty sure this was just missing, so I added it.

tests/ts/query-builder-api/mutating-methods.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  await Person.query().patch({
+    age: Person.query().avg('age'),
+    // You can use knex.raw instead of `raw()` if
+    // you prefer.
+    firstName: raw("'Jenni' || 'fer'"),
+    oldLastName: ref('lastName'),
+    // This updates a value nested deep inside a
+    // json column `detailsJsonColumn`.
+    'detailsJsonColumn:address.street': 'Elm street'
```
willsoto: The typings currently don't support this syntax, any thoughts? I am not sure if there is a way in TypeScript to type string object access like this....

koskimas: Yeah, I don't think we csn make the typings support this syntax. We could allow any property there, but it would weaken the typings for most people. We probably should add another syntax for updating json columns at some point. For now, you can just leave those out.

typings/objection/index.d.ts
koskimas: Yep, thanks!

tests/ts/query-builder-api/mutating-methods.ts
koskimas: It should be there, so feel free to add. Yes it's also chained from query.

typings/objection/index.d.ts
koskimas: Delicious copy-pasta. Thanks

typings/objection/index.d.ts
```diff
@@ -563,9 +564,7 @@ declare namespace Objection {
       ids: MaybeCompositeId | Partial<RelatedModel> | Partial<RelatedModel>[]
     ): NumberQueryBuilder<QB>;
 
-    <RelatedModel extends Model>(
-      ids: MaybeCompositeId | Partial<RelatedModel> | Partial<RelatedModel>[]
-    ): NumberQueryBuilder<QB>;
+    <RelatedModel extends Model>(ids: { [key: string]: MaybeCompositeId }): NumberQueryBuilder<QB>;
```
willsoto: I'm not sure how I feel about this. Do you have any thoughts? 

I added it to support the following case:

```ts
  await Person.relatedQuery('movies')
    .for(123)
    .relate({ foo: 50, bar: 20, baz: 10 });
```

koskimas: `Partial<RelatedModel>` already covers that case. The problem here is that `Person` doesn't have fields `foo`, `bar` or `baz`

willsoto:
Makes sense. So those are intended to be id columns? I’d like to improve
the naming a bit since I don’t think it’s really clear

typings/objection/index.d.ts
willsoto: Makes sense. So those are intended to be id columns? I’d like to improve
the naming a bit since I don’t think it’s really clear

willsoto: This is currently the error I am seeing:

Screen Shot 2019-09-27 at 8 56 58 AM

It seems like it might not be inferring the model correctly since it is still typed as `Partial<Model>`

koskimas: Oh, that's true! There's no way for that to infer the model type. That should be something like

```ts
  interface RelateMethod<QB extends AnyQueryBuilder> {
    (ids: MaybeCompositeId | Partial<ModelType<QB>> | Partial<ModelType<QB>>[]): NumberQueryBuilder<
      QB
    >;
  }
```

willsoto:
@koskimas This is good for another look. Should we delete the existing `example.ts` tests?

koskimas:
No, let's not delete any tests. The new tests don't cover everything in the example.ts file yet. Even if they did, I don't see a reason to delete the example.ts file.

willsoto:
No problem! I'll have some more PRs soon. Thanks for all the reviews.

Issue #1455, opened 2019-08-30
UpsertGraph deletes rows for relation which is not mentioned in graph


lynxtaa:
I'm using `objection@1.6.9` and have these models:

```js
class Role extends Model {
  static get tableName() {
    return 'roles'
  }

  static get relationMappings() {
    return {
      sets: {
        relation: Model.HasManyRelation,
        modelClass: 'Set',
        join: { from: 'roles.id', to: 'sets.roleId' },
      },
    }
  }
}

class Set extends Model {
  static get tableName() {
    return 'sets'
  }

  static get relationMappings() {
    return {
      setAttributes: {
        relation: Model.HasManyRelation,
        modelClass: 'SetAttribute',
        join: { from: 'sets.id', to: 'sets_attributes.setId' },
      },
    }
  }
}

class SetAttribute extends Model {
  static get tableName() {
    return 'sets_attributes'
  }
}
```

Suppose role with ID 1 already has set with ID 1.

If I'm using upsertGraph like this, It works correctly -- new row is inserted in "sets" table:

```js
transaction(Role.knex(), trx =>
  Role.query(trx).upsertGraph({
    id: 1,
    sets: [{ id: 1 }, { name: 'New Set' }],
  }),
)
```

But if I pass "setAttributes" for newly created set, **Objection deletes all setAttributes for the first set** (with id=1):

```js
transaction(Role.knex(), trx =>
  Role.query(trx).upsertGraph({
    id: 1,
    sets: [
      { id: 1 },
      { name: 'New Set', setAttributes: [{ name: 'New Set Attribute' }] },
    ],
  }),
)
```

Is this behaviour expected?

koskimas:
Could you provide a reproduction for this? I don't believe this is really happening. There are tests for this case. I'll fix it immediately if you can provide a reproduction.

lynxtaa:
Thanks for a quick response ! I've created reproducible example: https://github.com/lynxtaa/objection-issue

koskimas:
Thanks for the repro! I'll see what's going on asap.

lynxtaa:
@koskimas any updates on this?

koskimas:
I'll post here once I find time to work on this.

koskimas:
Fix coming up soon. Sorry it took so long.

koskimas:
Oh man, this was a pretty nasty bug! I can't believe why nobody reported this earlier. objection 1.6.10 will be released in couple of minutes.

[...]

PR #1487, opened 2019-09-26
Fixing bug #1486

### LINKED ISSUE #1486: TypeError: properties.reduce is not a function
I'm having this issue when i try to `omit` a property from a query:
```
ProductModel.query()
  .insert({ ...data, deleted: false })
  .omit('deleted')
  .returning('*');
```

But is working with the array version of `omit`
```
ProductModel.query()
  .insert({ ...data, deleted: false })
  .omit(['deleted'])
  .returning('*');
```

I guess it has something to do with [this line](https://github.com/Vincit/objection.js/blob/master/lib/queryBuilder/QueryBuilder.js#L711)

Version: `1.6.9`

rburdet:
This PR fix #1486

### DIFF
--- a/lib/queryBuilder/QueryBuilder.js
+++ b/lib/queryBuilder/QueryBuilder.js
@@ -707,12 +707,15 @@ class QueryBuilder extends QueryBuilderBase {
       modelClass = null;
     }
 
-    // Turn the properties into a hash for performance.
-    properties = properties.reduce((obj, prop) => {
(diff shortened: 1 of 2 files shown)


rburdet:
Tests are failing only in `node 12` because of 

*        UpsertGraph deletes rows for relation which is not mentioned in graph #1455

Do you have any hints on this ?

koskimas:
Thank you !

PR #1495, opened 2019-10-03
clarify error message on relation name collision

sfabriece:
When the name of a relation mapping collides with a column name, the current error message is not clear enough,


sfabriece:
It seems a test not related to my change is failing @koskimas

koskimas:
I don't think this is better than the old message. This would imply that the relation name and the relation mapping name are somehow different things. The documentation refers to relation's name as "relation name" everywhere. Maybe you could still use "relation name", but somehow make it more clear that it can be changed in the `Model.relationMappings`? (which really shouldn't be unclear if you have already defined the relationships...)

sfabriece:
I'm not sure about how to clarify this. The problem for me was that the error message didn't immediately make it clear what was wrong. Maybe an explanation on how the relation name and the relation mapping are the same would give me some clue? @koskimas  😄

koskimas:
I can take care of this.

PR #1500, opened 2019-10-04
Eager loading TypeScript tests

willsoto:
#1431

### DIFF
--- a/tests/ts/query-builder-api/eager-loading-methods.ts
+++ b/tests/ts/query-builder-api/eager-loading-methods.ts
@@ -1,0 +1,165 @@
+import { Person } from '../fixtures/person';
+
+(async () => {
+  await Person.query()
+    .where('firstName', 'Arnold')
+    .withGraphFetched('pets');
+
+  await Person.query().withGraphFetched('children.[pets, movies]');
+
+  await Person.query().withGraphFetched({
+    children: {
+      pets: true,
+      movies: true
+    }
+  });
+
+  await Person.query()
+    .withGraphFetched('children(selectNameAndId).[pets(onlyDogs, orderByName), movies]')
+    .modifiers({
+      selectNameAndId(builder) {
+        builder.select('name', 'id');
+      },
+
+      orderByName(builder) {
+        builder.orderBy('name');
+      },
+
+      onlyDogs(builder) {
+        builder.where('species', 'dog');
+      }
+    });
+
+  await Person.query().modifiers({
+    // You can bind arguments to Model modifiers like this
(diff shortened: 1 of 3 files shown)


tests/ts/query-builder-api/eager-loading-methods.ts
```diff
@@ -0,0 +1,74 @@
...
+    children: {
+      pets: true,
+      movies: true
+    }
+  });
+
+  await Person.query()
+    .withGraphFetched('children(selectNameAndId).[pets(onlyDogs, orderByName), movies]')
+    .modifiers({
+      selectNameAndId(builder) {
```
willsoto: I wonder if we can do better here with the `builder` typings. It is currently `any`.

koskimas: Really? Yep, we should definitely be able to do better. I can take a look.

typings/objection/index.d.ts
```diff
@@ -1099,6 +1099,7 @@ declare namespace Objection {
     mergeContext: ContextMethod<this>;
 
     modify: ModifyMethod<this>;
+    modifiers: ModifyMethod<this>;
```
koskimas: This type is not correct. Modifiers takes an object full of modifiers. Something like

```js
interface ModifiersMethod<QB extends AnyQueryBuilder> {
  (modifiers: Modifiers): QB
}
```

willsoto: Thank you! I was gonna ask about it. 

On a side note, I really like that API :+1:

typings/objection/index.d.ts
```diff
@@ -774,6 +774,10 @@ declare namespace Objection {
     ): SingleQueryBuilder<QB>;
   }
 
+  interface GraphExpressionObjectMethod<QB extends AnyQueryBuilder> {
+    (): any;
```
willsoto: I really think we can improve the types here. I was playing with `ModelRelations` but couldn't quite get it right. Any suggestions for this? I think we can access the current Model via `QB`?

koskimas: We probably could, but I don't think it's worth the trouble at this point. I'd guess this method is rarely used.

willsoto: Sounds good, I will leave as-is.

willsoto:
@koskimas Are you able to retry the build please?

koskimas:
Sure. I should really iron out those randomly failing tests.

willsoto:
@koskimas Thanks! CI passed so this should be good for review/merge 😄

PR #1504, opened 2019-10-07
Add pre-commit hook to run prettier and eslint

### LINKED ISSUE #1503: [Meta] Add pre-commit hook to run prettier
[See here for more information](https://prettier.io/docs/en/precommit.html).

Would you be open to adding a pre-commit hook to run prettier over files? I find that as I am working I end up causing unnecessary churn because certain files haven't been formatted.

willsoto:
Closes #1503


koskimas:
Thanks!

PR #1505, opened 2019-10-07
Add tests for join methods

willsoto:
#1431

### DIFF
--- a/tests/ts/query-builder-api/join-methods.ts
+++ b/tests/ts/query-builder-api/join-methods.ts
@@ -1,0 +1,63 @@
+import { Person } from '../fixtures/person';
+import { raw } from '../../..';
+
+(async () => {
+  await Person.query()
+    .joinRelation('pets')
+    .where('pets.species', 'dog');
+
+  await Person.query()
+    .joinRelation('pets', { alias: 'p' })
+    .where('p.species', 'dog');
+
+  await Person.query()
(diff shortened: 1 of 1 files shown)


tests/ts/query-builder-api/join-methods.ts
```diff
@@ -0,0 +1,63 @@
...
+    .where('pr:pt.species', 'dog');
+
+  await Person.query().innerJoinRelation('pets');
+  await Person.query().outerJoinRelation('pets');
+  await Person.query().leftJoinRelation('pets');
+  await Person.query().leftOuterJoinRelation('pets');
+  await Person.query().rightJoinRelation('pets');
+  await Person.query().rightOuterJoinRelation('pets');
+  await Person.query().fullOuterJoinRelation('pets');
+  await Person.query().join(raw('pets'));
```
willsoto: Typings don't currently allow `.join('pets')` - not sure if intentional or oversight. Happy to fix if an oversight.

koskimas: They are not supposed to support that. `*joinRelation` methods take relation names. Other join methods simply call the knex equivalents and know nothing about relations.

willsoto:
This should be ok to merge, that was the only thing I had a question about.

koskimas:
Thanks again!

PR #1506, opened 2019-10-08
"Other methods" TypeScript tests

willsoto:
#1431

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,12 +112,12 @@
     }
   },
   "lint-staged": {
-    "*.{js}": [
+    "*.js": [
       "prettier --write",
       "eslint --fix",
       "git add"
     ],
-    "*.{md,yml}": [
+    "*.{md,yml,ts}": [
       "prettier --write",
       "git add"
     ]
--- a/tests/ts/query-builder-api/other-methods.ts
+++ b/tests/ts/query-builder-api/other-methods.ts
@@ -1,0 +1,202 @@
+import { Person } from '../fixtures/person';
+import { UniqueViolationError } from 'db-errors';
+import { Animal } from '../fixtures/animal';
+import { Model, transaction } from '../../../typings/objection';
+
+(async () => {
+  const debugResult = Person.query()
+    .joinRelation('children')
+    .where('age', '>', '21')
+    .debug();
+
+  const personId = 1;
+  const pets = await Person.relatedQuery('pets').for(personId);
+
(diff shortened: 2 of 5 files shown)


koskimas:
This is failing in CI for real an not because of random fails.

willsoto:
Yes sorry! I’m gonna wrap this one up soon, just been really busy.

package.json
```diff
@@ -112,12 +112,12 @@
     }
   },
   "lint-staged": {
-    "*.{js,ts}": [
+    "*.js": [
```
willsoto: When I originally made this PR, I forgot to test eslint on `ts` files, and it doesn't support `ts` right now, so I've moved them into the other hook.

tests/ts/query-builder-api/other-methods.ts
```diff
@@ -0,0 +1,202 @@
...
+  builder.then(
+    () => {
+      console.log('success');
+    },
+    () => {
+      console.log('error');
+    }
+  );
+
+  // builder.map((obj) => obj);
```
willsoto: These are not present in the typings currently, but I wonder if they are being deprecated like the other similar methods (`pluck`, etc).

tests/ts/query-builder-api/other-methods.ts
```diff
@@ -0,0 +1,202 @@
...
+  // builder.reduce()
+
+  builder.catch(error => {
+    console.log(error);
+  });
+
+  // builder.return()
+  // builder.bind()
+  // builder.asCallback();
+  // builder. nodeify()
```
willsoto: No typings for this, unsure if deprecated or removed in v2.

tests/ts/query-builder-api/other-methods.ts
```diff
@@ -0,0 +1,202 @@
...
+    onBuild(builder: any) {}
+  });
+
+  Person.query()
+    .withGraphFetched('[movies, children.movies]')
+    .context({
+      onBuild(builder: any) {
+        builder.withSchema('someSchema');
+      }
+    });
```
willsoto: All the various context callbacks are untyped right now, hence the `any`

willsoto:
@koskimas few questions/comments - but otherwise ready for review.

willsoto:
@koskimas Now that 2.0 is released, do you think you have a chance to take a look at this?

PR #1532, opened 2019-10-26
On insert builder, verify if it is MSSQL

juniorUsca:
Check if the connections is MS SQL because on returning `output` has conflicts with triggers


koskimas:
So this disables implicit `returning` on mssql. Doen't this mean that we don't get the inserted item's id back, effectively breaking all insert related features in objection?

koskimas:
No response, closing.

PR #1557, opened 2019-11-12
- Require loops possible solution.

damian-pastorini:
Hi @koskimas , I had this require loop issue for a long time and been using the solution #1 from the guide, but today doing a refactor on my code I've managed to fixed it by not using default exports on my Models, so I've created this PR in case you consider it proper to have in the docs, is a short addition to comment that not using default exports may help avoid the issue.
BTW: I love Objection! Congratulations on your work! Is really great!


koskimas:
This is not true.

PR #1565, opened 2019-11-22
Make typescript more strongly (optionally)

udanpe:
It would be nice to have type safety for the query builder.
```ts
// without type safety
const user = await User.query()
    .where({ id: 1 })
    .select('id')
    .orderBy('id', 'desc');
// with type safety
const user = await User.query()
    .where<User>({ id: 1 })
    .select<User>('id')
    .orderBy<User>('id', 'desc')
```

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -1242,6 +1242,8 @@ declare namespace Objection {
       column: ColumnRef,
       callback: (this: QueryBuilder<QM, QM[]>, queryBuilder: QueryBuilder<QM, QM[]>) => void
     ): QueryBuilder<QM, RM, RV>;
+    <T>(object: Partial<T>): QueryBuilder<QM, RM, RV>;
+    <T>(column: Partial<keyof T>, op: string, value: string): QueryBuilder<QM, RM, RV>;
   }
 
   interface FindOne<QM extends Model> {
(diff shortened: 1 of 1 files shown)


koskimas:
Thanks for this PR, but we're about to release objection 2.0 where the types have been fully rewritten. What you are suggesting cannot be added though. You can easily have a query that has alot more columns than just the model columns.

udanpe:
> You can easily have a query that has alot more columns

I didn't touch previous declarations. If you use where with generic you have half of functionality but with type safety. But if you don't use you can write everything. It's optionally. Its better then nothing.
```ts
class User {
   firstName: string
}
const user = await User.query().where({ firstName3232323: 1 }) // ok
const user = await User.query().where<User>({ firstName3232323: 1 }) // error
```

udanpe:
I looked v2 and it's horrible. If now i can override d.ts v2 already have generic. Any reason to do that?
```ts
declare module 'objection' {
  interface Where<QM extends Model, RM, RV> extends WhereRaw<QM, RM, RV> {
    <T>(columns: Partial<T>): QueryBuilder<QM, RM, RV>;
    <T>(column: Partial<keyof T>, op: string, value: string): QueryBuilder<QM, RM, RV>;
  }
  interface OrderBy<QM extends Model, RM, RV> {
    <T>(column: keyof T, order?: 'asc' | 'desc'): QueryBuilder<QM, RM, RV>;
    <T>(columns: (Array<{ column: keyof T; order?: 'asc' | 'desc' }>)): QueryBuilder<QM, RM, RV>;
  }
  interface ColumnNamesMethod<QM extends Model, RM, RV> {
    <T>(...columnNames: Array<Partial<keyof T>>): QueryBuilder<QM, RM, RV>;
    <T>(columnNames: Array<Partial<keyof T>>): QueryBuilder<QM, RM, RV>;
  }
}
```

PR #1570, opened 2019-11-24
Update dependencies. Drop obsolete dependencies

kibertoad:


.eslintrc.json
```diff
@@ -1,5 +1,7 @@
 {
-  "parser": "babel-eslint",
```
kibertoad: Babel is no longer needed, Node 8 supports everything out-of-the-box.

package.json
```diff
@@ -68,28 +68,26 @@
     "knex": "0.x"
   },
   "devDependencies": {
-    "@babel/polyfill": "^7.4.4",
```
kibertoad: Newer versions of knex no longer rely on babel/polyfill.

PR #1574, opened 2019-11-28
fix:1572 incorrect type inference on insert

### LINKED ISSUE #1572: typings/objection - Incorrect type inference when using model instance query builder
```
const created = await Person
  .fromJson({ firstName: 'Jennifer' })
  .$query()
  .insert();
```
The type of `created` is incorrectly inferred as `Person[]` whereas it should be `Person`.

Type inference works nicely if we insert using static query builder

```
const created = await Person.query().insert({ firstName: 'Jennifer' });
```
Here, `created` is correctly inferred as `Person`.

royshouvik:
```
const created = await Person
  .fromJson({ firstName: 'Jennifer' })
  .$query()
  .insert();
```
The type of created is incorrectly inferred as Person[] whereas it should be Person.

Fixes #1572

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -683,9 +683,9 @@ declare namespace Objection {
   }
 
   interface Insert<QM extends Model> {
-    (modelsOrObjects?: Partial<QM>[]): QueryBuilder<QM, QM[]>;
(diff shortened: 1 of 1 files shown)


koskimas:
Thank you!

royshouvik:
Thank you  for `objection.js` 🙏

PR #1580, opened 2019-12-02
Remove `asCallback` / `nodeify` methods from docs

fiznool:
`asCallback` and `nodeify` were removed in v2, but these references still remain in the documentation.


koskimas:
Thank you!

PR #1582, opened 2019-12-02
Added statusCode to CreateValidationErrorArgs

fiznool:
To support (e.g.):

``` js
throw new ValidationError({
  statusCode: 409,  // <-- currently results in a TS error
  type: 'InvalidOneTimeCode',
  message: 'Wrong code',
  data: {
    supplied: code,
  },
});
```

### DIFF
--- a/tests/ts/validation.ts
+++ b/tests/ts/validation.ts
@@ -62,4 +62,22 @@ import { Person } from './fixtures/person';
       }
     }
   }
+
+  // gh-1582
+  try {
+    throw new ValidationError({
+      statusCode: 409,
+      type: 'InvalidOneTimeCode',
+      message: 'Wrong code',
+      data: {
+        supplied: '1234',
+      },
+    });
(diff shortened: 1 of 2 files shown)


fiznool:
Hmm. I believe that the `ValidationError.type` field should also be updated to

``` ts
type: ValidationErrorType | string;
```

Since, at the moment, if you create an error as above and later come to check whether the error type is equal to the original string, you get a TS error.

i.e. this is problematic:

``` ts
try {
  throw new ValidationError({
    type: 'InvalidOneTimeCode',
    message: 'Wrong code',
    data: {
      supplied: code,
    },
  });
} catch (e) {
  if(e instanceof ValidationError) {
    if(e.type === 'InvalidOneTimeCode') {  // <-- currently causes a TS error
      // Do something
    }
  }
}
```

koskimas:
Thank you :bowing_man:

PR #1584, opened 2019-12-02
Fix broken link in README.MD

joziahg:
Fixed db-errors Github link.


koskimas:
Thanks! Actually the `issues` suffix shouldn't be there at all.

PR #1587, opened 2019-12-02
adjust typescript definition for fromJson and fromDatabaseJson

jamespedid:
Split the `FromJsonMethod` interface into `FromJsonMethod` and `FromDatabaseJsonMethod`.

The new `FromJsonMethod` parameter now accepts the additional ModelOptions argument.
The new `FromDatabaseJsonMethod` interface is the same as the old `FromJsonMethod`.


koskimas:
Thanks!

PR #1592, opened 2019-12-03
[Objection2] Types: Ensure single query builder return type contains undefined

mattleff:
In Objection v2 the single query builder termination methods are typed as always returning the model:

```ts
const maybePerson = await Person.query().findById(123);
// maybePerson: Person
```

In v1 this was `maybePerson: Person | undefined`. Assuming this change was unintentional I can finish this PR to cover all the methods and pass the tests.


koskimas:
This change was intentional.

mattleff:
@koskimas The type doesn't match the actual result of the method. `findOne()`/`findById()`/`first()`/etc. can return `undefined`. Why shouldn't the type reflect that?

PR #1601, opened 2019-12-10
Update documentation links in README

eloyekunle:


koskimas:
Thanks for bringing this to my attention! the fix is not correct though. I've fixed them now.

PR #1622, opened 2019-12-19
Allow batch operations for sql server

caiquecastro:


koskimas:
So mssql returns an id for each inserted row, and not just the first or the last one?

caiquecastro:
image

caiquecastro:
image

lib/queryBuilder/operations/InsertOperation.js
```diff
@@ -27,8 +27,8 @@ class InsertOperation extends QueryBuilderOperation {
   }
 
   async onBefore2(builder, result) {
-    if (this.models.length > 1 && !isPostgres(builder.knex())) {
-      throw new Error('batch insert only works with Postgresql');
+    if (this.models.length > 1 && !isPostgres(builder.knex()) && !isMsSql(builder.knex())) {
```
caiquecastro: What do you think about creating a function to wrap this logic. Like `supportsBatchInsert`

PR #1643, opened 2020-01-10
make startTransaction typings return a Promise

JReinhold:
Judging from my experience when working with the `startTransaction` function, and from the documentation, I believe `startTransaction` returns a Promise that resolves to a `Transaction`, and not the actual `Transaction` itself.

Docs mentioning `startTransaction`, where it is awaited:

- https://vincit.github.io/objection.js/api/model/static-methods.html#static-starttransaction
- https://vincit.github.io/objection.js/guide/transactions.html#creating-a-transaction


koskimas:
Thank you! I've fixed this in 2.1.2. Your PR had errors so I couldn't merge it.

PR #1651, opened 2020-01-19
[Objection2] Types: Ensure single query builder return type contains undefined

### LINKED ISSUE #1649: New types in v2 have some issues with single/multiple return types, e.g. with returning("*")
We're having some issues with typings in 2.0.  (v2.1.2 - latest)

Breaking out into multiple issues, I can log separately if needed.

1. This example from Objection's [Postgres Returning Tricks](https://vincit.github.io/objection.js/recipes/returning-tricks.html) actually is typed as returning an array - not a single object.  So this code in the official Recipe won't work.  (The recipe code may just be wrong since patch().where() seems like it could always return multiple, unless querying for 'id' is a specific case in the types that indicates singular return.)

```
const jennifer = await Person.query()
  .patch({ firstName: 'Jenn', lastName: 'Lawrence' })
  .where('id', 1234)
  .returning('*');
```

2. findOne() is typed as returning `T`, not `T | undefined`.  This means we always have to manually add `as T | undefined`, as findOne could always fail.  In TS there are things that can fail but nevertheless return only T (array indexing for example), but findOne function seems far more analagous to JS array.find() which returns `T | undefined`, as I believe findOne should.  Or, perhaps a tryFindOne function would make sense, which could return undefined.

3. I'd expect these two queries to both return single User objects:

`await User.query().patch({}).findOne({}).returning("*");`

`await User.query().findOne({}).patch({}).returning("*");`

However, the former is typed as returning a single User and the latter returns an array.  I'd suggest the latter is the more intuitive usage (find object then patch it) and should return a single User.

---

I'm sure I'm not aware of all the interactions in the types and whether the changes above are mutually compatible.  We're huge fans of Objection on our team, thanks for all your work contributing this to the JS/TS community!

mattleff:
This fixes part of #1649 and refs #1592.

### DIFF
--- a/doc/recipes/custom-query-builder.md
+++ b/doc/recipes/custom-query-builder.md
@@ -83,6 +83,7 @@ class MyQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<M, R> {
   // name of the query builder class.
   ArrayQueryBuilderType!: MyQueryBuilder<M, M[]>;
   SingleQueryBuilderType!: MyQueryBuilder<M, M>;
+  MaybeSingleQueryBuilderType!: MyQueryBuilder<M, M | undefined>;
   NumberQueryBuilderType!: MyQueryBuilder<M, number>;
   PageQueryBuilderType!: MyQueryBuilder<M, Page<M>>;
 
--- a/tests/ts/custom-query-builder.ts
+++ b/tests/ts/custom-query-builder.ts
@@ -3,6 +3,7 @@ import { Model, QueryBuilder, Page, TransactionOrKnex } from '../../';
 class CustomQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<M, R> {
   ArrayQueryBuilderType!: CustomQueryBuilder<M, M[]>;
   SingleQueryBuilderType!: CustomQueryBuilder<M, M>;
+  MaybeSingleQueryBuilderType!: CustomQueryBuilder<M, M | undefined>;
   NumberQueryBuilderType!: CustomQueryBuilder<M, number>;
   PageQueryBuilderType!: CustomQueryBuilder<M, Page<M>>;
 
@@ -51,7 +52,7 @@ const people: Promise<Person[]> = Person.query()
       .someCustomMethod()
   );
 
-const pets: Promise<Animal> = new Person()
+const pets: Promise<Animal | undefined> = new Person()
   .$relatedQuery('pets')
   .someCustomMethod()
   .where('id', 1)
--- a/tests/ts/documents.ts
+++ b/tests/ts/documents.ts
@@ -15,5 +15,5 @@ import { Person } from './fixtures/person';
   const jenniferFromDb = await Person.query().findById(jennifer.id);
 
   console.log(jennifer.address.city); // --> Tampere
-  console.log(jenniferFromDb.address.city); // --> Tampere
+  console.log(jenniferFromDb!.address.city); // --> Tampere
 })();
--- a/tests/ts/examples.ts
+++ b/tests/ts/examples.ts
@@ -67,7 +67,7 @@ class Person extends objection.Model {
 
     properties: {
       firstName: {
-        type: 'string',
+        type: 'string'
       }
     }
   };
@@ -173,6 +173,14 @@ function takesModelSubclass<M extends objection.Model>(m: M) {}
 function takesModel(m: objection.Model) {}
 function takesModelClass(m: objection.ModelClass<any>) {}
 
+// Borrowed from https://github.com/TypeStrong/ts-expect/blob/39f04b5/src/index.ts
+type TypeEqual<T, U> = Exclude<T, U> extends never
+  ? Exclude<U, T> extends never
+    ? true
+    : false
+  : false;
+const expectsTrue = <T extends true>() => 1;
+
 const takesPerson = (person: Person) => {
   person.examplePersonMethod('');
 };
@@ -203,7 +211,14 @@ takesPersonQueryBuilder(Person.query());
 async () => {
   takesPeople(await Person.query().where('lastName', lastName));
   takesPeople(await Person.query().where({ lastName }));
-  takesMaybePerson(await Person.query().findById(123));
+  let person = await Person.query().findById(123);
+  takesMaybePerson(person);
+  expectsTrue<TypeEqual<Person | undefined, typeof person>>();
+  person = await Person.query().findOne({
+    id: 123
+  });
+  takesMaybePerson(person);
+  expectsTrue<TypeEqual<Person | undefined, typeof person>>();
   takesMaybePerson(await Person.query().findById('uid'));
 };
 
@@ -494,7 +509,7 @@ async () => {
 // (fewer characters than having each line `const qbNNN: QueryBuilder =`):
 
 const maybePerson: Promise<Person | undefined> = qb.findById(1);
-const maybePerson2: Promise<Person> = qb.findById([1, 2, 3]);
+const maybePerson2: Promise<Person | undefined> = qb.findById([1, 2, 3]);
 
 // query builder knex-wrapping methods:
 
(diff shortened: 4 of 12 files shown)


koskimas:
Unfortunately this is a breaking change and will need to wait for 3.0, which is at least a year away. I repeatedly asked for help when I was writing 2.0. Unfortunately I got none, and we're stuck with the current typings for a while.

mattleff:
@koskimas I'm sorry I didn't see your requests for types help earlier! Objection is an important library for us and I'm glad to help out. Were your requests in issues or do you have a blog/newsletter/etc. that I could follow to catch those in the future?

While I agree that technically this is a breaking change, I would argue that the current types are broken, so this is more like a bugfix. Having the types tell me that something as common as a single find (`findOne()`/`findById()`/`first()`) is guaranteed to return a `Model`, when in fact it could return `undefined` is setting myself up for broken code (hello `Cannot read property 'xyz' of undefined`). I'm only one user but I'll probably wait until this is fixed to upgrade from 1.x.

eugene-kim:
> While I agree that technically this is a breaking change, I would argue that the current types are broken, so this is more like a bugfix. Having the types tell me that something as common as a single find (findOne()/findById()/first()) is guaranteed to return a Model, when in fact it could return undefined is setting myself up for broken code (hello Cannot read property 'xyz' of undefined).

Agreed. If the typings suggest that a model instance exists, these are false positives. Semver with respect to Typescript changes make this a bit of a gray area, though in terms of being a breaking API change. @koskimas what do you think?

koskimas:
@eugene-kim Every single typescript project using objection would break --> a breaking change.

mattleff:
@koskimas 

> Every single typescript project using objection would break --> a breaking change.

In many/most cases the fact that a project's TS compilation would error/"break" indicates that we are exposing a runtime error at compile time instead. I know those cases will need to be addressed but compile time errors are always better than runtime errors.

Here's a few comments that speak to how breaking changes in types are handled on DefinitelyTyped:

https://github.com/DefinitelyTyped/DefinitelyTyped/issues/25677#issuecomment-388215959
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43434#issuecomment-610455341
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44589

Please know I respect your right to wait for 3.0 or whenever to merge this PR. I would just encourage considering this as a bugfix for a false confidence the types shouldn't be giving. Thanks for a great library and all your work maintaining it! 👏

simon-hrabec:
I encountered this issue recently. I had conditions where I was checking if I got a value back or not but eslint was failing on this if statement because it was considered unreachable code/always truthy.

rclmenezes:
@koskimas I appreciate you, but please reconsider.

It's not possible to write our own custom query builder. When extending an object in Typescript, you can't modify the type signature of an existing method. The only option would be to make a fake new method (`findByOne2` or something).

Without some kind of fix, these broken type signatures will cause of sorts of runtime errors.

Thanks for your time ❤️

llamadeus:
I can only join @mattleff @eugene-kim and @rclmenezes and I really appreciate the work you put in such an extensive ORM. While I totally understand the point of a "breaking change", I would like to mention two things which came to my mind.

1. *The current type definition does not reflect the underlying javascript code.* As Objection.js is a javascript library and the typings have been generated with hindsight, I would consider the missing `undefined` type more of not yet implemented thing because the source of truth _is_ the code and not the type definition.

2. *Code relying on the presence of a model after `.first()` is already broken.* Extending the return type with `undefined` will not break non-broken code. IMO checks for the possible `undefined` return value should already be present.

Yet I would like you to know that I will completely respect your decision whether this being a v2 or a v3 change. This is your project and many people rely on a stable and reliable code base, especially people using Objection.js in production.

koskimas:
Ok, I'll implement this. This PR however is nowhere near enough, so this won't get merged. I hope I'll get some help answering the 10 angry issues created once this breaks people's code. Often you know for sure an item exists in the DB (for example in a transaction after inserting or locking a row using a row lock) and there's no need to handle the `undefined` case. All those will break even though the code is correct.

rclmenezes:
Woot! Thanks @koskimas !

> Often you know for sure an item exists in the DB (for example in a transaction after inserting, updating or locking a row using a row lock) and there's no need to handle the undefined case.

That's a perfect use case for Typescript's not-null assertion operator:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

If I know that this query will certainly have a result, I can do:

```typescript
const result = (await Model.query().findById())!
```

If anyone complains about their broken typings, we can tell them to add `!` and who knows, it may actually catch some bugs!

eugene-kim:
> Ok, I'll implement this. This PR however is nowhere near enough, so this won't get merged. I hope I'll get some help answering the 10 angry issues created once this breaks people's code. Often you know for sure an item exists in the DB (for example in a transaction after inserting, updating or locking a row using a row lock) and there's no need to handle the `undefined` case. All those will break even though the code is correct.

@koskimas one thing that might make your life easier would be to give people a heads up in advance as much as possible, so pinning this issue and making a highly visible section in the README may reduce the number of people surprised when this change gets released.

Thank you again for all the great work

capaj:
@koskimas I would love to answer to those 10 angry issues.
I am serious. Just forward them to my inbox and I'll gladly explain to the issue creator why it has to be this way.

Please merge this PR or implement it for the current major. Or break the major. As long as those typings get fixed.

elhigu:
I would still like to see this kind of change to be done as a breaking change even if it is debatable if it is a "bug fix" or not. Main point being the one that @koskimas originally commented about it. 

Though if there would be a way to opt in "fixed" typings earlier that could be released meanwhile. It would also provide easier route to update project to new typings.

wanton7:
Please merge this because currently typing are incorrect. It's been also very annoying because eslint thinks I used `if` unnecessarily to check result from `findOne` because it doesn't return undefined.

theodoregold:
Everyone clearly wants to get this ASAP. There really should be a sensible solution to this. Like releasing 3.0 for this and later 4.0?

eugene-kim:
> Everyone clearly wants to get this ASAP. There really should be a sensible solution to this. Like releasing 3.0 for this and later 4.0?

Big fan of this as it would allow for progress *and* not break existing projects. @koskimas what do you think?

acro5piano:
If your purpose is to prevent unexpected `undefined`, `throwIfNotFound()` is useful.

```ts
await User.query().findById(9999).throwIfNotFound()
```

wanton7:
@acro5piano this is not about that. This is about TypeScript definitions being wrong, meaning they don't match JavaScript code is doing.

acro5piano:
@wanton7 Thank you for clarifying it. I know type definition has no effect on JavaScript code and the problem is only related to type definition.

Until this PR is merged and a new version released, we can override the type definition by merging interface like this. It is not the ideal solution but it works.

```
export type { Model, MaybeCompositeId, SingleQueryBuilder } from 'objection'

declare module 'objection' {
  export interface QueryBuilder<M extends Model, R = M[]> extends Promise<R> {
    findById(id: MaybeCompositeId): SingleQueryBuilder<this> | undefined;
  }
}
```

image

Another solution is to use patch-package but overriding is the cleaner way to do it IMO

koskimas:
Supporting knex >= 0.95 requires us to make breaking changes, so I'll release a 3.0 soon. This will be fixed at the same time.

fiznool:
A humble request: if the query builder contains a `throwIfNotFound`, the typings should reflect this by always returning the model type, without `| undefined` - since, the nature of `throwIfNotFound` is to throw an error if the record is not found in the database. Hence, we can be sure that the return value will always be a model, never `undefined`.

I believe this would also satisfy the scenario mentioned above, without resorting to the non-null operator (which is discouraged by some ESLint configs):

> Often you know for sure an item exists in the DB (for example in a transaction after inserting, updating or locking a row using a row lock) and there's no need to handle the undefined case.

Yes, a breaking change, but by chaining a `throwIfNotFound` call to the query, this is more explicit about the behaviour you are anticipating.

Just my own opinions here, I agree wholeheartedly with the proposal in this PR, and also offer my warmest thanks to all the maintainers of this fabulous library.

acro5piano:
> A humble request: if the query builder contains a throwIfNotFound, the typings should reflect this by always returning the model type, without | undefined - since, the nature of throwIfNotFound is to throw an error if the record is not found in the database. Hence, we can be sure that the return value will always be a model, never undefined.

This PR already includes that feature, which is really awesome!

koskimas:
This one couldn't be merged or rebased easily so I copy pasted the relevant parts. This is now implemented in the latest 3.0 alpha version. To install, use `npm install objection@next`

capaj:
Amazing stuff! Thanks @koskimas! I will take this for a spin this or next weekend.

PR #1657, opened 2020-01-21
docs: move v1.x and changelog inside release-notes

### LINKED ISSUE #1655: Document clarification
Hi, I think there are some confusing regarding documentation.

Here is the header of the docs:
image
There is a tab named `v1.x`. But it is not clear what that means.
Although I know that objection is now on version 2, but some of my friends don't and make an assumption that objection is on version 1.x and the current docs is about version 1.x.

I think this need to be clarified.
Maybe that tab should be grouped with the `Changelog` tab and renamed to `Revisions` or `Versions` for better understanding?

peara:
Fix #1655:
- Create a new nav **Release Notes**
- Move contents of **v1.x** and **Changelog** inside **Release Notes**

Screenshot:
image


koskimas:
Awesome! Thank you!

PR #1658, opened 2020-01-21
docs: fix typo

matheuspiment:
Apparently the correct reference is to the `toString()` method.


koskimas:
Nope, the current is correct.

PR #1659, opened 2020-01-22
Fix typo

SteveDeWald:
I'm still familiarizing myself with Objection, so I may be mistaken, but I believe this was what was intended.


koskimas:
Thanks!

PR #1665, opened 2020-01-28
added deepRemoveRequired

### LINKED ISSUE #1664: patch validator logic does not deeply remove "required" key from jsonSchema
I ran into an issue when a jsonSchema contains nested objects with required keys. Currently, it appears that only the top-level required key is removed when using patch.

```
static get jsonSchema() {
    return {
      type: 'object',
      required: ['id'],
      properties: {
        id: { type: 'integer' },
        meta: {
          type: 'object',
          required: ['id2'],
          properties: {
            id2: { type: 'integer' },
            foo: { type: 'string' }
        }
   }
};
```

Because of this, when I try to patch ["meta:foo"] = "bar", I get a validation error because id2 isn't present. I'm submitting a PR for review in a few minutes.

bompus:
fixes #1664

### DIFF
--- a/lib/model/AjvValidator.js
+++ b/lib/model/AjvValidator.js
@@ -245,6 +245,10 @@ function objectHasDefaults(obj) {
 }
 
 function deepRemoveRequired(jsonSchemaProperties) {
+  if (jsonSchemaProperties === undefined) {
+    return;
+  }
+
   if (jsonSchemaProperties.required !== undefined) {
     delete jsonSchemaProperties.required;
   }
--- a/lib/model/AjvValidator.js
+++ b/lib/model/AjvValidator.js
@@ -131,9 +131,10 @@ class AjvValidator extends Validator {
   }
 
   compilePatchValidator(jsonSchema) {
-    jsonSchema = jsonSchemaWithoutRequired(jsonSchema);
+    let clonedSchema = cloneDeep(jsonSchema);
+    clonedSchema = jsonSchemaWithoutRequired(clonedSchema);
     // We need to use the ajv instance that doesn't set the default values.
-    return this.ajvNoDefaults.compile(jsonSchema);
+    return this.ajvNoDefaults.compile(clonedSchema);
   }
 
   compileNormalValidator(jsonSchema) {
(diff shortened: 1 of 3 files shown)


lib/model/AjvValidator.js
```diff
@@ -244,12 +244,33 @@ function objectHasDefaults(obj) {
...
+
+  for (const key of Object.keys(jsonSchemaProperties.properties)) {
+    deepRemoveRequired(jsonSchemaProperties.properties[key]);
+  }
+}
+
 function jsonSchemaWithoutRequired(jsonSchema) {
   const subSchemaProps = ['anyOf', 'oneOf', 'allOf', 'not', 'then', 'else'];
 
+  const clonedSchema = Object.assign(jsonSchema, {});
```
koskimas: You only clone the outermost level of the object, but modify all levels in the `deepRemoveRequired`. That will end up modifying the original schema.

tests/unit/model/AjvValidator.js
```diff
@@ -0,0 +1,73 @@
...
+                  meta3: { type: 'object' }
+                }
+              }
+            }
+          }
+        }
+      };
+
+      const validator = new AjvValidator({ onCreateAjv: () => {} });
+      const validators = validator.getValidator(modelClass('test', schema), schema, true);
```
koskimas: You should also add a test that `validator.getValidator(modelClass('test', schema), schema, false)` still has *all* the required fields.

lib/model/AjvValidator.js
koskimas: Instead of using deepRemoveRequired, which mutates the object, you should recursively apply `subSchemaWithoutRequired` for all properties of type `object` or `array` with objects in them.

koskimas:
Thank you! This is a good addition, and I don't know why it has been missing all these years :grimacing: Your implementation has some room for improvement though. The rest of the code around it use a "functional" approach, where `subSchemaWithoutRequired` returns a new shallow clone of a schema without `required`. Your code uses a mutating function that, at the moment, incorrectly mutates the original schema. To keep a consistent style, you should use `subSchemaWithoutRequired` for this too, and create a pure function that recursively calls `subSchemaWithoutRequired` for all `object` and `array` (that contain objects) type properties.

lib/model/AjvValidator.js
bompus: Are you opposed to just doing a cloneDeep at the top of compilePatchValidator() ? This way we ensure the original object never gets mutated and it could simplify some stuff downsteam as a result.

bompus:
Alright. Take a look at this and see if it's any better. I wrestled with it for a while before I got to a working result.

I also ran across and fixed another bug:

After merging master and updating unit tests, I noticed subSchemaProps ['anyOf', 'oneOf', 'allOf', 'not', 'then', 'else'] were not being removed from the schema at all. I broke some logic outside of the Object.assign for easier debugging and it also had the side effect of resolving that issue. Unit test has been updated to verify all of the fields that should be removed have actually been removed.

Hopefully the PR is now closer to what you had in mind. I'm running out of time to devote to this issue so I'm going to have to move on. If it needs massaging to meet your coding standards, feel free to complete the work, don't let this go stale!

koskimas:
I can implement this.

PR #1679, opened 2020-02-11
feat: parallelize initialization of table metadata

rockrep:
Loading tableMetadata for lots of tables is slow because it's done in serial.  This PR parallelizes the process for faster initialization

### DIFF
--- a/lib/initialize.js
+++ b/lib/initialize.js
@@ -6,11 +6,7 @@ async function initialize(knex, modelClasses) {
     knex = modelClasses[0].knex();
   }
 
-  const promises = modelClasses.reduce((acc, modelClass) => {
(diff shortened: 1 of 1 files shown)


lib/initialize.js
```diff
@@ -6,11 +6,11 @@ async function initialize(knex, modelClasses) {
     knex = modelClasses[0].knex();
   }
 
-  for (const modelClass of modelClasses) {
-    await modelClass.fetchTableMetadata({
-      knex
-    });
-  }
+  const promises = modelClasses.reduce((acc, modelClass) => {
```
elhigu: Isn't `.map` more appropriate in this case?

rockrep: yes, doh.  Thanks.

PR #1683, opened 2020-02-12
Add underscoreBetweenUppercaseLetters to knexSnakeCaseMappers

### LINKED ISSUE #1676: Feature request: make knexSnakeCaseMappers behavior with multiple consecutive uppercase letters configurable
First off, thanks so much for your work on Objection.  It's really a great piece of work.

The issue I'm running into is that the current `knexSnakeCaseMappers` implementation converts strings like `isALevel` to `is_alevel`, `isBLevel` to `is_blevel`, etc.  This means that if you have a column like `is_a_level` or `is_b_level`, you can't access it when using `knexSnakeCaseMappers` (as far as I can tell, at least). `isBLevel` and `isBlevel` both get mapped to the same thing: `is_blevel`.

It looks like the code adds a special case for when there are two or more consecutive uppercase letters here: https://github.com/Vincit/objection.js/blob/5b76a4c1e0dee70649d32b8a59383021bb5d5927/lib/utils/identifierMapping.js#L55-L61

Would it make sense to add a parameter to disable this special case, similar to how you handled `underscoreBeforeDigits`?
https://github.com/Vincit/objection.js/blob/5b76a4c1e0dee70649d32b8a59383021bb5d5927/lib/utils/identifierMapping.js#L24

brianmcd:
Fixes #1676

### DIFF
--- a/lib/utils/identifierMapping.js
+++ b/lib/utils/identifierMapping.js
@@ -21,7 +21,14 @@ function memoize(func) {
 // camelCase to snake_case converter that also works with non-ascii characters
 // This is needed especially so that aliases containing the `:` character,
 // objection uses internally, work.
-function snakeCase(str, { upperCase = false, underscoreBeforeDigits = false } = {}) {
+function snakeCase(
+  str,
+  {
+    upperCase = false,
+    underscoreBeforeDigits = false,
+    underscoreBetweenUppercaseLetters = false
+  } = {}
+) {
   if (str.length === 0) {
     return str;
   }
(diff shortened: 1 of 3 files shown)


koskimas:
Thank you! Clean implementation and good tests! Wish I got more of these 💯

brianmcd:
Thanks for the quick merge!  The up-to-date dev setup instructions, docker containers, and extensive test quite made it really easy to contribute.

koskimas:
This still needs documentation and typescript typings. I can do that, but I'm currently quite busy with my work so it will be a while. If you have more time, I'd be more than happy about a PR for those too.

brianmcd:
Sure!  I'm happy to make those updates.  I think docs are missing for the `underscoreBeforeDigits` option, so I'll add those, too.

Are the relevant pages [here](https://github.com/Vincit/objection.js/blob/e6466352e8ecec300e4308ec562ff489267c3ec1/doc/api/objection/README.md#knexsnakecasemappers) and [here](https://github.com/Vincit/objection.js/blob/e6466352e8ecec300e4308ec562ff489267c3ec1/doc/recipes/snake-case-to-camel-case-conversion.md)?

For the TypeScript typings, I added the property [here](https://github.com/Vincit/objection.js/pull/1683/files#diff-cd41e62e17d6020205941718b64712a9R1288) in this PR.  Is there another place I need to add it, or are we good on that part?

koskimas:
Oh sorry I missed the typing change in your PR. That's it for those 👍 And yep, those are the relevant pages. I'll release a new version as soon as the docs are updated.

Issue #1688, opened 2020-02-15
QueryBuilder promise typings are incorrect.


sripberger:
Using `extends Promise<R>` as is being done in Objection's QueryBuilder typings is incorrect, according to how Objection is implemented. The QueryBuilder does not actually have the `Promise.prototype` in its prototype chain. Saying that it does causes tsc to believe properties exist on them that actually don't.

Notably, the native Promise prototype in Node (as well as most browsers) implements [finally][1]. Objection's QueryBuilders do *not* implement `finally`, nor do they implement any other functionality of that type besides `then` and `catch`.

I'm not simply suggesting that they should add `finally`. The use of `extends` here is simply incorrect and will produce more problems like this if and when any additional functionality is added to the native Promise type.

TypeScript provides the `PromiseLike<R>` interface for these situations, though that interface only
requires `then` and not `catch`. What QueryBuilders implement as far as promise functionality is concerned is this:

```ts
interface CatchablePromiseLike<R> extends PromiseLike<R> {
    catch<FR = never>(
      onrejected?: ((reason: any) => FR | PromiseLike<FR>) | undefined | null
    ): Promise<R | FR>;
}
```

I will have a PR up shortly to make the necessary corrections, but this will be a breaking change for dependents written in TypeScript. Promise and PromiseLike (even CatchablePromiseLike) are not interchangeable, specifically *because* Promises implement additional stuff.

I understand there might me some unwillingness to make this change, but it *is* causing some problems for some people, notably me, as discussed [here][2].

It's not something that should be rushed into, of course, but it would be nice to have this for the next major release.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally
[2]: https://github.com/typescript-eslint/typescript-eslint/issues/1572#issuecomment-582577518

PR #1689, opened 2020-02-15
Fix query builder promise typings

sripberger:
Fixes for #1688.

Note that this is a breaking change for any dependents written in TypeScript. `CatchablePromiseLike` can be returned from `async` functions no problem, but as can be seen in the changes to `examples.ts`, any non-async function or variable that explicitly defines a `Promise` type will need to be changed.

### DIFF
--- a/tests/ts/custom-query-builder.ts
+++ b/tests/ts/custom-query-builder.ts
@@ -34,7 +34,7 @@ class Person extends BaseModel {
   pets!: Animal[];
 }
 
-const people: Promise<Person[]> = Person.query()
+const people: PromiseLike<Person[]> = Person.query()
   .someCustomMethod()
   .where('firstName', 'lol')
   .someCustomMethod()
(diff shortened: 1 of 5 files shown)


koskimas:
This is an insanely breaking change and we cannot merge it before 3.0, which is far in the future.

sripberger:
> This is an insanely breaking change and we cannot merge it before 3.0, which is far in the future.

Which is totally reasonable. Though it is worth noting that it is kind of broken as-is. I just want to make sure everyone is aware of it so it can be fixed, far in the future if need be.

koskimas:
Yeah, it's broken and this is s good change, but you are the first person to bring this up so it doesn't seem that critical.

sripberger:
Fair enough. I think what that means is that nobody is using `finally` on these, which isn't surprising. I don't think I've ever used it anywhere, to be honest.

Feel free to close and reopen when 3.0 is realistic, or whatever else seems appropriate. :+1:

capaj:
> This is an insanely breaking change and we cannot merge it before 3.0, which is far in the future.

I don't think I agree with that. Sure, the typings change a bit, but no one is having `Model.query().finally()` in their code anyway even if the TS typings let them do that now, because this code will never run.
I can't imagine a situation where this could break a TS compilation of a code that's running correctly now,
If I were the maintainer, I'd merge this into 2.3.0, or 2.4.0 releases, but that just me.

koskimas:
Supporting knex >= 0.95 requires us to make breaking changes, so I'll release a 3.0 soon. This will be fixed at the same time.

koskimas:
Cant merge because of conflicts, but I copy pasted the relevant parts and this will be merged soon. Thank you for your contribution!

PR #1707, opened 2020-03-07
Update dependencies

kibertoad:
Also updates Objection.js tests to be compatible with newer Knex versions.


kibertoad:
Knex removed the last bits of Bluebird from it, so we need to remove all references to Bluebird-specific methods from Objection as well.

kibertoad:
@maximelkin Considering that you are an expert on Bluebird removal - any suggestions on what to replace calls to .reflect() with?

kibertoad:
@koskimas What's the original intent behind assertions like
```
        Promise.try(() => {
          return Model1.query();
        }).reflect(),
```
? Wouldn't just executing same thing without `try` or `reflect` also fail test as rejection would?

kibertoad:
@koskimas All right, I think I've got all of them, tests are green :)

tests/unit/queryBuilder/QueryBuilder.js
```diff
@@ -666,7 +666,7 @@ describe('QueryBuilder', () => {
       })
       .runBefore(() => {
         expect(mockKnexQueryResults[0]).to.equal(1);
-        return Bluebird.delay(1).return(++mockKnexQueryResults[0]);
+        return Bluebird.delay(1).then(() => ++mockKnexQueryResults[0]);
```
maximelkin: here logic changed, because mockQueryResults[0] was incremented before delay, now after

tests/integration/misc/usingUnboundModelsByPassingKnex.js
```diff
@@ -428,59 +429,59 @@ module.exports = session => {
...
+        }).catch(err => createRejectionReflection(err)),
 
         Promise.try(() => {
           return Model1.query(session.knex)
             .findById(1)
             .then(model => {
               return model.$query();
             });
-        }).reflect()
+        }).catch(err => createRejectionReflection(err))
```
maximelkin: This can be moved to map over Array<Promise> and 
`expect(promise).to.eventually.be.rejectedWith({name: '', message: ''})`

tests/unit/queryBuilder/QueryBuilder.js
kibertoad: Isn't it more correct to increment it after delay, if we are simulating query happening during that delay?

maximelkin: Yes, you are right

tests/integration/misc/usingUnboundModelsByPassingKnex.js
maximelkin: Not forcing to this solution, just proposing

kibertoad: Nah, you are right, that is a cleaner solution, current util method is basically a compatibility hack. Just need a bit of time to implement the change. Thanks for reviewing!

koskimas:
Thank you @kibertoad!

PR #1712, opened 2020-03-12
Fix incorrect method usage in patchAndFetch example

georgejdanforth:
The example for `patchAndFetch` is incorrectly using `patchAndFetchById`. That is fixed in this PR :)


koskimas:
Thanks!

PR #1744, opened 2020-05-04
Fix wrong link in docs

agarwalrounak:
No functional changes here, just fixing a wrong link in docs.
`jsonSchema` was pointing to `/api/model/static-properties.html#static-tablename`, modified it to point to `/api/model/static-properties.html#static-jsonschema`


koskimas:
Thank you!

PR #1758, opened 2020-05-23
add `withRecursive` to QueryBuilder typing

lecstor:
Let me know if there's anything else that needs to be added, I didn't see anything for it's counterpart `withWrapped`.


koskimas:
Thank you!

PR #1767, opened 2020-06-04
(feature) optional param added to throwIfNotFound method

### LINKED ISSUE #1722: [Feature Request] Optional message param to throwIfNotFound
**Describe the solution you'd like**
Would love to be able to provide a message string to the throwIfNotFound method.

**Describe alternatives you've considered**
N/A

**Additional context**
Would help solve the pain of writing a few lines of code to throw a custom error every time a necessary object doesn't exist.

vasdhara:
closes #1722 

Is this what was required for the feature request? Please let me know if I have missed anything. A first-time contributor so slightly overwhelmed.

### DIFF
--- a/doc/api/query-builder/other-methods.md
+++ b/doc/api/query-builder/other-methods.md
@@ -999,9 +999,9 @@ You can replace `Model.NotFoundError` with your own error by implementing the st
 
 ##### Arguments
 
-| Argument | Type   | Description                      |
-| -------- | ------ | -------------------------------- |
-| data     | object | optional object with custom data |
+| Argument | Type   | Description                                                                                                                                                                                                                                                                                                                      |
+| -------- | ------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| data     | object | optional object with custom data may contain any property such as message, type. The object is returned in the error thrown under the `data` property. A special case is for the optional message property. This is used to set the title of the error. These extra properties can be leveraged in the error handling middleware |
 
 ##### Return value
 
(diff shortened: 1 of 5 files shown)


doc/api/query-builder/other-methods.md
```diff
@@ -1071,13 +1071,19 @@ console.log(firstPerson.age);
...
 
 Causes a [Model.NotFoundError](/api/types/#class-notfounderror) to be thrown if the query result is empty.
 
 You can replace `Model.NotFoundError` with your own error by implementing the static [Model.createNotFoundError(ctx)](/api/model/static-methods.html#static-createnotfounderror) method.
 
+##### Arguments
+
+| Argument | Type   | Description                      |
+| -------- | ------ | -------------------------------- |
+| data     | object | optional object with custom data |
```
koskimas, line 1085: There's no way for people to use this based on this documentation. Where does this data go? What properties can it have? How are different properties treated (like the special case for `message`).

tests/integration/find.js
```diff
@@ -574,7 +574,7 @@ module.exports = session => {
             });
         });
 
-        it('.throwIfNotFound() with empty result', done => {
+        it('custom .throwIfNotFound() with empty result', done => {
```
koskimas, line 577: Why did you add the word `custom` here?

koskimas:
Hi @vasdhara! Thank you so much for your contribution! I added a couple of comments. You also need to update the typescript typings.

tests/integration/find.js
vasdhara: was a typo, reverted

doc/api/query-builder/other-methods.md
vasdhara: I've updated the description and added a custom type example, is that sufficient?

vasdhara:
@koskimas thanks. added fixes. Also, added a new interface for throwIfNotFound typings, since I couldn't find a previously written interface that could be reused in this case.

doc/api/query-builder/other-methods.md
koskimas: That's excellent!

Issue #1782, opened 2020-06-26
Concatenating map function is not working


roggervalf:
Hi, I'm using the latest version of this package on one of my project and looking on its documentation, appears map function here:
https://vincit.github.io/objection.js/api/query-builder/other-methods.html#map
I tried in this way:
```js
const userNames=await User.query().map((user=>user.name))
```
As I see now, I avoid using pluck since is going to be deleted in the next version (3v), so I would like to know if this function should be deprecated too, so docs should be updated, or just am I using it in a wrong way?`

koskimas:
That method has been removed in 2.0 version. I forgot to remove the documentation.

roggerval:
@koskimas thank you for the quick response.

roggervalf:
Thank you @koskimas for the quick response.

PR #1785, opened 2020-06-27
Deleting map method from documentation

roggervalf:
This PR deletes map method as explained here: #1782


koskimas:
Thank you! Could you also remove the `reduce` method while you're at it :smile: It has also been removed.

roggervalf:
Sure, no problem.

PR #1810, opened 2020-08-04
consider currently specified schema when getting column info

BuckAMayzing:
I brought the issue that this addresses up over in gitter - the short version is that when getting column metadata, the currently specified schema is not considered and passed on to the column info query.

The only way I could find to make this happen was to track the current schema in `internalOptions`. Other recommendations for a way to make this happen are welcome.

### DIFF
--- a/lib/queryBuilder/InternalOptions.js
+++ b/lib/queryBuilder/InternalOptions.js
@@ -6,6 +6,7 @@ class InternalOptions {
     this.keepImplicitJoinProps = false;
     this.isInternalQuery = false;
     this.debug = false;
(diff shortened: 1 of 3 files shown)


koskimas:
I can't find any problem with this. Thanks for the PR!

koskimas:
v2.2.3 will be released in a couple of hours with this fix.

BuckAMayzing:
> I can't find any problem with this. Thanks for the PR!

Thanks @koskimas! In the event there is a regression, tag me, and I'll pick it up.

PR #1822, opened 2020-08-19
fix: toKnexQuery accepts optional knexBuilder argument

alexkvak:
Typings fix

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -1058,7 +1058,9 @@ declare namespace Objection {
     timeout: TimeoutMethod<this>;
     columnInfo: ColumnInfoMethod<this>;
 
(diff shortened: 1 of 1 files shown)


koskimas:
As you can see [here](https://vincit.github.io/objection.js/api/query-builder/other-methods.html#toknexquery) the optional argument is not part of the public API and therefore should not be in the typings.

alexkvak:
@koskimas but my case is not public API. I created CustomQueryBuilder which extends QueryBuilder and I need to override toKnexQuery.

Of course I need to preserve all arguments and pass them to the parents toKnexQuery method.

alexkvak:
Maybe you can help me to choose the best solution for this?

I need to change the query while regular query execution and while resultSize execution.

I've overridden execute and toKnexQuery for these. Maybe there is the better way? Probably I can add onBuild hook in query builder constructor? It is correct way?

PR #1825, opened 2020-08-20
multiple databases recipe using AsyncLocalStorage  API

harish2704:


capaj:
the same technique can be used to run transactions. We used to pass `trx` everywhere, but ever since @cztomsik came up with the idea of using  AsyncLocalStorage we don't need to pass it around in arguments everywhere. 
It's quite a nice simplification of the `transaction`.
Would be great to have it supported in objection 3.0.
Only drawback is that minimal node.js version has to be 12.17.0.
Lower versions don't have AsyncLocalStorage. Would you accept a PR adding something like `transactionAsyncStorage` as an alternative to `transaction` @koskimas ?

harish2704:
@capaj: I am also using this setup in our production env.  I m not sure why authors are not addressing this PR.

I would suggest to either reject this PR by pointing out a valid reason, or accept this . 

since it is a documentation change, it won't be breaking anything.

koskimas:
This is a nice trick but there's absolutely no guarantee anywhere that the `knex` method is always called to get the knex instance. Also monkey-patching the prototype is a no-no.

I can't merge this as a recipe since this is not tested to work in all situations. It might work for you and your use cases, but once this is a recipe, it must work for all possible cases.

PR #1835, opened 2020-09-04
fix typo in query examples

jure:
Fixes code to match the query shown in the docs below:

```sql
select "persons".* from "persons"
where ("age" < 40 or "age" > 60)
and "firstName" = 'Jennifer'
order by "lastName" asc
```


koskimas:
Thanks

PR #1841, opened 2020-09-07
Add InsertAndFetchOperation#models

### LINKED ISSUE #1839: insertAndFetch() does not provide inputItems in the static query hooks beforeInsert() and afterInsert()
When using `insetAndFetch()`, the static query hooks `beforeInsert()` and `afterInsert()` always receive an empty array in `inputItems`, instead of the passed model data.

reproduction-template.js:
```js
let Model;

try {
  Model = require('./').Model;
} catch (err) {
  Model = require('objection').Model;
}

const Knex = require('knex');
const chai = require('chai');

let inputItems = null

async function main() {
  await createSchema();

  ///////////////////////////////////////////////////////////////
  // Your reproduction
  ///////////////////////////////////////////////////////////////

  const data = {
    firstName: 'Jennifer',
    lastName: 'Lawrence'
  };
  await Person.query().insertAndFetch(data);
  chai.expect(inputItems[0]).to.include(data);
}

///////////////////////////////////////////////////////////////
// Database
///////////////////////////////////////////////////////////////

const knex = Knex({
  client: 'sqlite3',
  useNullAsDefault: true,
  debug: false,
  connection: {
    filename: ':memory:'
  }
});

Model.knex(knex);

///////////////////////////////////////////////////////////////
// Models
///////////////////////////////////////////////////////////////

class Person extends Model {
  static get tableName() {
    return 'Person';
  }

  static get jsonSchema() {
    return {
      type: 'object',
      required: ['firstName', 'lastName'],

      properties: {
        id: { type: 'integer' },
        firstName: { type: 'string', minLength: 1, maxLength: 255 },
        lastName: { type: 'string', minLength: 1, maxLength: 255 }
      }
    };
  }

  static beforeInsert(args) {
    inputItems = args.inputItems;
  }
}

///////////////////////////////////////////////////////////////
// Schema
///////////////////////////////////////////////////////////////

async function createSchema() {
  await knex.schema
    .createTable('Person', table => {
      table.increments('id').primary();
      table.string('firstName');
      table.string('lastName');
    });
}

main()
  .then(() => {
    console.log('success');
    return knex.destroy();
  })
  .catch(err => {
    console.error(err);
    return knex.destroy();
  });
```

lehni:
So that the beforeInsert() and afterInsert() static query hooks return correct model-data in inputItems when using insertAndFetch()

* Includes tests

Closes #1839

### DIFF
--- a/lib/queryBuilder/operations/InsertAndFetchOperation.js
+++ b/lib/queryBuilder/operations/InsertAndFetchOperation.js
@@ -14,6 +14,10 @@ class InsertAndFetchOperation extends DelegateOperation {
     }
   }
 
+  get models() {
+    return this.delegate.models;
+  }
+
   async onAfter2(builder, inserted) {
     const modelClass = builder.modelClass();
     const insertedModels = await super.onAfter2(builder, inserted);
--- a/tests/integration/staticHooks.js
+++ b/tests/integration/staticHooks.js
@@ -2512,7 +2512,7 @@ module.exports = session => {
             });
         });
 
-        it('should have access to `inputItems`', () => {
+        it('should have access to `inputItems`', async () => {
           Movie.beforeInsert = createHookSpy(({ inputItems }) => {
             expect(inputItems.length).to.equal(1);
             expect(inputItems[0] instanceof Movie).to.equal(true);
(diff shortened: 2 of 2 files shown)


koskimas:
Thanks!

PR #1844, opened 2020-09-08
Fix: Objection iterated over every byte of Buffer properties

rluba:
I noticed huuuge response times when saving models that contain a binary Buffer.

The profiler showed that Objection was burning seconds iterating over every byte of the binary property, pushing and poping its indices, while trying to find a graph string in those bytes.

This PR fixes that particular problem, but there might be other, similar issues since `isObject` casts a _very_ wide net in JS.


koskimas:
Could you use `Buffer.isBuffer` instead of `instanceof`? Other than that, this is good to go. I'll merge this as soon as you make the change. Nice catch!

rluba:
👍 Done!

koskimas:
Thank you!

PR #1845, opened 2020-09-08
Add missing type declaration for $afterFind instance hook

TehNrd:
A simple fix to add a type for $afterFind instance method. There is one for $afterGet but that is deprecated.

I can create an issue if necessary but wasn't sure such a small change required an issue.


koskimas:
Thanks!

alexkvak:
@koskimas Could you release this fix?

PR #1847, opened 2020-09-09
Add clearWithGraphFetched to index.d.ts

puckey:


koskimas:
This is not a public method and therefore shouldn't be added to typings.

PR #1848, opened 2020-09-09
chore(typings): Typecheck WhereMethod object

jeffreypriebe:
When in TypeScript, querying doesn't check the object of `Where` clauses. Removing the broad type (`(obj: object) => QB)`) ensures that TypeScript does the type checking.

Added to the find-methods with a [`// @ts-expect-error` ](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments) compiler directive to ensure that it is invalid.

**Impact:** For repos with incorrect properties referenced, this will change to now show up as compiler errors (which is the desired effect).

The additional change on l104 is a perfect example of what is actually caught with this type change.
```diff
// Errors since 'name' doesn't exist on 'Person'
- .orWhere({ name: 'Tester' });
+ .orWhere({ lastName: 'Tester' });
```

### Notes / Additional

- Currently Travis doesn't appear to run typescript, could be added (happy to submit a PR to do this) to type check.
<br />I'm not a Travis expert, but something like:
  ```diff
  // package.json
  "scripts": {
  + "build-travis": "npm run tsc --noEmit"
  ```
  ```diff .travis.yml
  # .travis.yml
  script:
  + - npm run-script build-travis
  - npm run-script test-travis
  ```

- This doesn't affect tests since it is a typing change only.
- This only affects the single function whose signature was changed (when passing an object). Subsequent draft PR coming with a proposal for other methods.

### DIFF
--- a/tests/ts/query-builder-api/find-methods.ts
+++ b/tests/ts/query-builder-api/find-methods.ts
@@ -19,6 +19,12 @@ import { Person } from '../fixtures/person';
     firstName: 'Jennifer',
     lastName: 'Lawrence'
   });
+  await Person.query().findOne({
+    firstName: 'Jennifer',
+    lastName: 'Lawrence',
+    // @ts-expect-error
+    DOESNT_EXIST: 'none'
+  });
   await Person.query().findOne('age', '>', 20);
   await Person.query().findOne(raw('random() < 0.5'));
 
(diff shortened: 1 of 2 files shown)


koskimas:
See the comment you removed. We cannot remove the object version. Objection is a query builder and the model properties are not the only things you can filter on.

jeffreypriebe:
@koskimas Thanks. I read the comment and didn't understand it. I checked the git history for more info and couldn't surmise it's purpose.

For my understanding only (no worries about not answering): Is the idea that you don't want your Objection model to necessarily be 100% fidelity with the database?
(Such that one could query on things that weren't modeled in the model - but knew to be in the database.)

PR #1859, opened 2020-09-22
doc: minor semantic fix

kamikazechaser:
- update to correct article for meta description


koskimas:
An es-q-el friendly
A sequel friendly

Depends on how you say it right? I never understood the sequel pronunciation.

kamikazechaser:
Fair dinkum

PR #1864, opened 2020-09-27
Make knexSnakeCaseMappers work with raw query responses

kjs3:
The knexSnakeCaseMappers' `postProcessResponse` function is called with raw queries but the data is nested in the `rows` attribute. 

This change maps data in the `rows` array but leaves the rest of the raw result intact.

Test case is a raw Postgres upsert query which I don't think is possible another way.

Refactored `postProcessResponse` generally to be more concise using `map`.


kibertoad:
I've merged the other PR, could you please rebase to reduce noise?

kjs3:
@kibertoad: done

kjs3:
I'm actually rethinking this now. 

It's probably better to map the keys on `rows` _in place_ rather than lift it out. That would otherwise leave the raw query the same as it was.

kjs3:
@kibertoad: I left `rows` where it was in the raw result and only applied the mapping. I think this should be less surprising to people already opting in to raw queries. 

This change definitely shouldn't be a patch release.

lib/utils/identifierMapping.js
```diff
@@ -180,13 +180,12 @@ function knexIdentifierMappers({ parse, format, idSeparator = ':' } = {}) {
...
       if (Array.isArray(result)) {
-        const output = new Array(result.length);
-
-        for (let i = 0, l = result.length; i < l; ++i) {
-          output[i] = parseKeys(result[i]);
-        }
-
-        return output;
+        return result.map(parseKeys);
+      } else if (result.rows) {
```
koskimas, line 184: What if the `result` is a single record (row) and the row has a column called `rows`? Or can `result` ever be an object here?

kjs3: @koskimas: I've put in some test cases for that and it seems fine. 
I had a test for a postgres jsonb column named "rows" with snake_cased keys just to ensure that `camelCaseKeys` didn't parse into that when returning. It didn't, but the test took a lot of setup nested within the `isPostgres()`. It just seemed like too much. I added a unit test for `camelCaseKeys()` directly which I think gets the job done.

elhigu: Also `.raw` query responses are different depending on db driver. Is there something that should be fixed for other drivers too?

kjs3: I started looking into the raw response for MySQL and Sqlite and got it working but the complexity of handling every raw response for the different DBs is adding up. I don't have access to MSSQL or Oracle so I can't do anything there.

If other people think having this in Objection for raw queries is worth it, I think the way to do it would be separate functions for each DB's raw response shape. People can then opt-in to just the function they need for their DB.

Let me know if that's something you would want.

For now, I'm just going to use my own knex mapper that works for what I need.

PR #1870, opened 2020-09-30
Add missing argument types in QueryBuilder's for

fer22f:
This PR addresses the missing QueryBuilder, model instances and array of model instances types in the argument of QueryBuilder's `for` method. Because as per documentation, it should be as such:

---

This method takes one argument (the owner(s) of the relation) and it can have any of the following types:

* A single identifier (can be composite)
* An array of identifiers (can be composite)
* A QueryBuilder
* A model instance.
* An array of model instances.

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -1006,7 +1006,7 @@ declare namespace Objection {
     ): NumberQueryBuilder<this>;
 
     unrelate(): NumberQueryBuilder<this>;
-    for(ids: ForIdValue | ForIdValue[]): this;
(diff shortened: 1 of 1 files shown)


lehni (changes requested):
The code that handles the various possible values of `for()` / `RelationOwner.owner` is:

```js
function detectType(owner) {
  if (isModel(owner) || isModelArray(owner)) {
    return Type.Models;
  } else if (isReferenceArray(owner)) {
    return Type.Reference;
  } else if (isQueryBuilder(owner)) {
    return Type.QueryBuilder;
  } else {
    return Type.Identifiers;
  }
}
```

It looks to me that the suggested typings do not cover the case of `isReferenceArray()` yet, correct?

Also, could you rename `MF` to just `M` like in the rest of the typings?

Issue #1874, opened 2020-10-02
issue with non-strict equality in upsertGraph 'findChanges' 


jjsquillante:
Hi @koskimas 

Thanks for objection.js :)

This stems from issue [732](https://github.com/Vincit/objection.js/issues/732)

While this [logic works](https://github.com/Vincit/objection.js/blob/6ab1ff95ac367b2535a02e3281d3bad9fa73f06c/lib/queryBuilder/graph/patch/GraphPatchAction.js#L201) in most cases, we have a case where `upsertGraph` is failing to identify a value that's changed due to `non-strict equals`. For example, if we have a `false` value stored as our `currentGraph` and an empty array `[]` as our new value, this comparison will not be picked up as a valid diff. ( since `false == [] // true --> gets pushed into the unchangedProps array` ) 

`current: false`
 `new value: []`
`false == [] // true `

As a way to be minimally invasive (since this is an edge case), I'm proposing we add an optional `applyStrictEquality` property to `upsertGraph` that will allow the user to override the default (non-strict equals). 

[link to PR 
](https://github.com/Vincit/objection.js/pull/1875)

let me know if there's anything that needs clarification - thanks!

jjsquillante:
ping

PR #1875, opened 2020-10-02
add option to force apply strict equality in upsertGraph options

jjsquillante:
hey @koskimas, thanks for objection.js 💯 

This PR fixes [issue#1874](https://github.com/Vincit/objection.js/issues/1874)

The test suite passes 100% with the new changes, let me know if you would like that I add a few tests.

Also, let me know if you have any opinions on going in a different direction to fix this edge case.

Thanks,
James


jjsquillante:
ping @koskimas @kibertoad - lmk if you need any more information. thanks

jjsquillante:
friendly ping as a reminder of this PR @koskimas @kibertoad - we're experiencing this bug in production. thanks

PR #1880, opened 2020-10-07
Fix RelationOwner._getPropsFromModels

coderNeos:
It would drop some unique props in case if they had buffer values.
As the result some related models won't be fetched.
This commit fixes the join function, which converts buffers to generate correct key for the prop.


koskimas:
Thank you for the fix!

PR #1894, opened 2020-10-23
Typings: Match $knex() arguments to static knex()

mmartinsky:
Adds the optional knex argument to the instance method to match the static equivalent, as implied by https://vincit.github.io/objection.js/api/model/instance-methods.html#knex


koskimas:
This shouldn't be done. Providing the knex instance for the method would imply that you are setting the knex instance of the model instance, but in reality, this would set the knex instance of the whole model class globally. There's no way to set the knex instance of a model instance unfortunately.

Issue #1466, opened 2019-09-12
 i use knex with objectionjs but i can't retrieved data  error this "TypeError: Cannot read property 'isPartial' of null" 


maprangsoft:
hello i use knex with objectionjs but i can't retrieved data 
error this "TypeError: Cannot read property 'isPartial' of null"  i make relation in database  completed 
i use mysql database
----------------------------
Model:
Categorie: https://1drv.ms/u/s!An-gKc2bfnTOgYgu0MpqJNbT-nCbgg?e=BoI2S9
Blog: https://1drv.ms/u/s!An-gKc2bfnTOgYgtCw_FvlmwCOSkGg?e=QrmVIz
Controller:
BlogController: https://1drv.ms/u/s!An-gKc2bfnTOgYgvgtEx6uemT4SUbw?e=0IiVoP
thank you advance.

koskimas:
You have this

```js
  await Blog
    //.query()
    .relatedQuery("categorie")
    .then(data => {
      res.send(helper.sendResponse(data, "Blog retrieved successfully."));
    })
    .catch(function(err) {
      res.send(helper.sendError("Blog can't retrieved. " + err));
    });
```

what are you trying to do here? You don't specify a blog id anywhere, so how would objection know for which blog to fetch the `categorie`? Or are you trying to fetch all `Blog` instances and load the `categorie` with them? In that case, you need to use `eager`:

```js
  await Blog
    .query()
    .eager("categorie")
    .then(data => {
      res.send(helper.sendResponse(data, "Blog retrieved successfully."));
    })
    .catch(function(err) {
      res.send(helper.sendError("Blog can't retrieved. " + err));
    });
```

The static `relatedQuery` can only be used for subqueries. The instance method `$relatedQuery` can be used to build queries to a single relation of a **model instance**. You probably need to take a closer look to the documentation. The `guide` section is a good place to start.

maprangsoft:
oh thank you very much this is thing i want. ^___^

PR #1895, opened 2020-10-24
added error for relatedQueries ommiting 'for' outside subqueries

Mahamed-Belkheir:
Hello, recently I came across a weird issue where an insert statement was failing silently, that insert was using `relatedQuery` without `for`, so I went through some digging, I found that the foreign key is overridden even if it was included in the insert. through some other digging, I found that lacking .for outside other non-subqueries could result in a not so obvious error, like the one in #1466,

Originally, I thought these were two problems, so I had placed two if statements to check if a target exists or not, one in RelationInsertOperation.js, and the other in RelationOwner.js, and I wrote a quick test, and as I found out, the check in the RelationOwner constructor was able to catch both error conditions, and what I deduced was:

`findFirstNonPartialAncestorQuery` keeps looking for the first non-partial parent query, my condition checks if that parent even exists, let alone isPartial or not. partial queries that lack a parent at all would never pass this function. this is what causes #1466 in `where` queries, but the curious thing is that `insert` statements do not error out here, yet my if statement catches it, I'm not really sure why. If you have a hint as to why I can dig deeper but as this is my first time peeking into Objection's internal I kept my journey short today.

As is though, this doesn't seem to break any other previous tests, and catches my two error conditions.

### DIFF
--- a/lib/relations/RelationOwner.js
+++ b/lib/relations/RelationOwner.js
@@ -190,7 +190,10 @@ function isQueryBuilder(item) {
 
 function findFirstNonPartialAncestorQuery(builder) {
   builder = builder.parentQuery();
-
+  if (!builder)
+    throw Error(
+      'query method `for` ommitted outside a subquery, can not figure out relation target'
+    );
   while (builder.isPartial()) {
     if (!builder.parentQuery()) {
       break;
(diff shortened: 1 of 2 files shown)


kibertoad:
@koskimas Maybe should be either merged, or rejected with an explanation why this change is undesired?

lehni:
@kibertoad I think we should merge this. What's your take on this?

PR #1917, opened 2020-11-18
Progressing refactor to ts

roggervalf:
This PR is trying to handle a ts config for refactoring files to ts, I see this is something planned to do. Hope this could help at least this config. Please let me know what do you think about it.


koskimas:
There are no plans to refactor to TS.

PR #1949, opened 2020-12-21
Add express.js example

Satyam1203:


koskimas:
Thanks for the PR but I don't feel like this adds anything to the table. The only part that's different from the koa examples is the express specific code that has nothing to do with objection.

PR #1953, opened 2020-12-26
Update knex devDependency to 0.21.15

kibertoad:
This is likely to be the last version in 0.21 branch (unless we'll be forced to do hotfixes), so it makes sense to bump.


kibertoad:
ESLint is failing since last dependency update, new ESLint is not compatible with Node 8/9

PR #1965, opened 2021-01-09
Add plugin

JaneJeon:
So hey, thanks for fixing that issue. This was the plugin I was referring to, and I’ve been using it since v1 for my own projects.

It’s quite a bit more complicated than the others listed here, but I think many Objection users will find it useful, especially with plans to integrate other ACLs/authorization frameworks in the future as well.


koskimas:
Awesome! Thanks!

PR #1968, opened 2021-01-13
Fix a misleading code example for returning tricks

jan-osch:
# Summary 
* The example in docs is a incorrect: `.patch(...)` with `.returning('*')` will update multiple rows, result will be an array
* Added `.first()` to showcase a nice chainable API (an alternative would be array restructuring) 

P.S
Thanks, guys for the awesome project 😸


koskimas:
Thank you!

PR #1976, opened 2021-01-19
Update relations.md

jillesme:
Hello,

I've been using Objection in my Node 14 project for a while now. Because I am using ES Modules, there is no need for circular import solutions. This is not reflected in the docs however so I updated them.

If you'd like me to change the wording / examples please do let me know!

### DIFF
--- a/doc/guide/relations.md
+++ b/doc/guide/relations.md
@@ -134,11 +134,36 @@ class Person extends Model {
 }
 ```
 
-## Require loops
+## Require loops (non ECMAScript modules only)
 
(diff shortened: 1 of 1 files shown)


jillesme:
@koskimas wdyt 😊?

capaj:
@jillesme interesting. We have suffered from require looops horribly in our huge backend project at @leaplabs.
We write it in TS in ES modules, but we transpile to commonJS as we run on node 12 in production.
How come ES modules don't suffer from cyclic deps? Is there any mention of this property in ecmascript spec or anywhere else?

mweberxyz:
I realized this was an issue as well and created my PR before I saw this-- whoops.

I would recommend adding in the error that users will encounter if not using the static getter for relationMappings to help people find this documentation when they search for the error that occurs if they try to define a `static relationsMapping = {...}` with a circular import (as here: https://github.com/Vincit/objection.js/pull/2071/commits/dd50b6e7a1dcf14b40cf1a334ea1d5b5fdf7db5f#diff-293ee03e47e7a07dd14dda512ad27e1df4cf4eec7190e9d81e5fdf86e966290fR225)

mifi:
I just had this exact question while looking at the docs, only to have to search for a workaround and I found this. Would be nice if it could be merged

jillesme:
@koskimas should I close this if there is no interest in merging it? I think it could be helpful

lehni:
Thanks @jillesme!

PR #1988, opened 2021-01-30
Return model from setRelated and appendRelated

### LINKED ISSUE #1987: setRelated and appendRelated return undefined
The documentation claims `Model.setRelated` and `Model.appendRelated` return the input `model` for chaining purposes:

 - [Model.setRelated](https://github.com/Vincit/objection.js/blob/master/doc/api/model/instance-methods.md#setrelated)
 - [Model.appendRelated](https://github.com/Vincit/objection.js/blob/master/doc/api/model/instance-methods.md#appendrelated)

I found that these functions return `undefined` instead of `model` as the documentation says they should. Opening up the source code it looks like these functions are incorrectly returning `this` (which is undefined in the scope of the function) instead of `model`.

The offending lines:

 - https://github.com/Vincit/objection.js/blame/master/lib/model/modelSet.js#L98
 - https://github.com/Vincit/objection.js/blame/master/lib/model/modelSet.js#L114

It looks like changing these lines to `return model` should fix the issue.

leontastic:
Fixes #1987

### DIFF
--- a/lib/model/modelSet.js
+++ b/lib/model/modelSet.js
@@ -95,7 +95,7 @@ function setRelated(model, relation, models) {
     }
   }
 
-  return this;
+  return model;
 }
 
 function appendRelated(model, relation, models) {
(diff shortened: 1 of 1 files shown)


koskimas:
Thank you!

PR #1989, opened 2021-02-02
Fix #1954 fix graphOptions typings

### LINKED ISSUE #1954: GraphOptions.joinOperation is incorrectly typed as required
GraphOptions.joinOperation is typed as required here:

https://github.com/Vincit/objection.js/blob/412bb77b9cf69496e477c38fc10bca679ceca461/typings/objection/index.d.ts#L752-L758

which is not correct because it defaults to `leftJoin`.

In particular, this leads to Typescript error in innocent uses like:

```ts
MyModel.query().withGraphFetched("foo", { maxBatchSize: 1 })
```

result:

```
Argument of type '{ maxBatchSize: number; }' is not assignable to parameter of type 'GraphOptions'.
  Property 'joinOperation' is missing in type '{ maxBatchSize: number; }' but required in type 'GraphOptions'.ts(2345)
```

Freakazo:
Update typescript definitions for GraphOptions, the previous type definitions are now incorrect.

Aliases should be an object and joinOperation should be optional.

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -752,8 +752,8 @@ declare namespace Objection {
   export interface GraphOptions {
     minimize?: boolean;
     separator?: string;
-    aliases?: string[];
-    joinOperation: string;
+    aliases?: {[key: string]: string};
+    joinOperation?: string;
     maxBatchSize?: number;
   }
 


koskimas:
Thanks!

PR #1992, opened 2021-02-08
Update to knex 0.95.0

kibertoad:
This is a breaking change and will make Objection incompatible with older versions.


kibertoad:
@koskimas Looking not only on TS changes, but also on tests failures due to direct references to knex internal files, it looks like next knex major will require an objection major to support. How would you prefer to handle that? Do you have timeline for next Objection major already?

koskimas:
@kibertoad I think the internals are only touched in the tests.

Zeswen:
How is this PR doing? @koskimas are you going to ignore/update the tests or going with the major release?

Thank you both for your amazing work with these two packages :)

koskimas:
After giving this some more thought, I think I'll release 3.0 even though I meant to save major updates for significant changes, but that's the right thing to do here. I'll also get the chance to fix #1651  and #1689 at the same time, both of which require breaking changes.

hsynlms:
@kibertoad Knex [upgrading docs](https://github.com/knex/knex/blob/master/UPGRADING.md) says:

> Global static Knex.raw support dropped, use instance knex.raw instead. (require('knex').raw() won't work anymore)

https://github.com/Vincit/objection.js/blob/412bb77b9cf69496e477c38fc10bca679ceca461/lib/queryBuilder/RawBuilder.js#L42

So is not it required to change RawBuilder accordingly? Do not mind if it's nonsense, please.

kibertoad:
@hsynlms From my understanding that is not a global `raw`, but the one invoked on knex instance already connected to DB, which is still perfectly valid.

hsynlms:
@kibertoad I have not investigated a lot at that time but I think you are right. `.knex()` call returns the already created instance itself, not the global one.

charlescapps:
Is there an approximate timeline for this work to support knex 0.95.0+? We really enjoy using Objection.js, and we have a security issue in our dependencies that cascades to requiring `knex` being updated. We appreciate all the work on this!

kasvith:
Any update on this?

grahamb:
Now that this has been merged and 3.0.0 is in alpha, is there a timeline for release?

koskimas:
Still no timeline. 

The breaking changes should now be done. You should be able to use the 3.0 alpha versions pretty safely.

koskimas:
@grahamb Go fuck yourself with that thumb

PR #1999, opened 2021-02-12
Add from method to ReferenceBuilder interface

garusis:
This PR only adds the `from` method to the `ReferenceBuilder` type since right now is missing.

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -58,7 +58,9 @@ declare namespace Objection {
     ): ValueBuilder;
   }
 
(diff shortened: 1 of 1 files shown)


typings/objection/index.d.ts
```diff
@@ -58,7 +58,9 @@ declare namespace Objection {
     ): ValueBuilder;
   }
 
-  export interface ReferenceBuilder extends Castable {}
+  export interface ReferenceBuilder extends Castable {
+    from(tableReference: string): this;
+  }
```
koskimas, line 63: There should be an empty line here like between every other item in this file.

PR #2071, opened 2021-06-25
Add documentation: avoiding require loop with ESM

mweberxyz:
This PR adds documentation for simple method of avoiding require loop when using ES Modules without the need for dynamic import, absolute file path reference, or `modelPaths` usage.


mweberxyz:
Whoops, dupe #1976

Issue #1684, opened 2020-02-12
Nested relation on a recursively loaded graph?


Ryanauger95:
Lets say that an organization has a "unit" field, and this "unit" has an unspecified number of sub-units. To get the organization and all of the units and subunits, one would use:

```
const user = await Organization.query()
      .findOne(...)
      .withGraphFetched(
        "unit.[subunits.^]"
      );

```

And this works awesome! But what if for every unit, there is a "location" relation? Without recursion, we would do the following:
```
   const user = await Organization.query()
      .findOne(...)
      .withGraphFetched(
        "unit.[subunits.[location]]"
      );
```

How do we do this with recursion, so that every sub unit returns a location as well?

Ryanauger95:
https://github.com/Vincit/objection.js/issues/198
found this solution. is a workaround, works!

Ryanauger95:
would be cool built in, but it's not hard to implement.

Trunksome:
@Ryanauger95 did you find a way to fetch unlimited levels of subunits? 

What @koskimas suggested in https://github.com/Vincit/objection.js/issues/198#issuecomment-250723588 only does a set amount of levels, as far as I can tell.

marceliwac:
@Trunksome - this would be a nice feature to have indeed.

marceliwac:
For what it's worth, using the wildcard quantifier (`*`) in the `withGraphFetched`'s template appears to cause infinite loop if there is a recursive relation on that model.

```js
// This causes infinite loop.
SomeModel.query().withGraphFetched('root.*');

// The relationship graph looks roughly like so:
/*

   MyModel has a relationship "pets" which have relationship "children" that relate to other MyModel(s), i.e.:  
 
   <SomeModel>
         |
         | - root - <MyModel> 
                |
                | - children <MyModel>
                      |
                      | - <MyModel>
                      |
                      | - <MyModel>
                            |
                            | - children <MyModel>
                            |      |
                            |      | - <MyModel>
                            |
                            | - <MyModel>

 */  

```

PR #2118, opened 2021-10-06
Remove redundant TODO.

marceliwac:
While attempting to develop a fix for #1684 I've noticed a comment that was meant to be implemented once v2 was released. It appears to be redundant now that the functionality of having multiple relations loaded as a graph has been implemented.


koskimas:
That test is for having multiple sub paths that share a root, so it's actually still relevant and should be enabled instead.

PR #2159, opened 2021-12-01
Fix issues #2158 and #1649

rclmenezes:
I gave issues #2158 and #1649 a shot to see if I could fix them.

My approach was:
- If patch/update are running on a SingleQueryBuilder, return a new kind of query builder called OneQueryBuilder instead of NumberQueryBuilder. OneQueryBuilder is simply QueryBuilder<M, 1>
- In `ReturningMethod`, if QB extends OneQueryBuilder, return SingleQueryBuilder<QB>.

It worked! I added a small test as well to confirm. Not sure if there are any corner cases I forgot about...

PR #2160, opened 2021-12-01
Fix insert graph method typing

rclmenezes:
Previously, the SingleQueryBuilder definition went first, so even if an array was passed as `graph`, it would be typed as SingleQueryBuilder. This created all sorts of weird effects, like making it impossible to write a type predicate on a model subclass.

This commit reorders it so it tries to type it as ArrayQueryBuilder first. It also adds a test to make sure that model subclasses work correctly.

### DIFF
--- a/tests/ts/model/model-inheritance.ts
+++ b/tests/ts/model/model-inheritance.ts
@@ -1,0 +1,9 @@
+import { Animal } from '../fixtures/animal';
+
+export class Dog extends Animal {
+  species!: 'Canis familiaris';
+}
+
+export function isDog(animal: Animal): animal is Dog {
+  return animal.species === 'Canis familiaris';
+}
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -687,15 +687,15 @@ declare namespace Objection {
   interface InsertGraphMethod<M extends Model> {
     <QB extends AnyQueryBuilder>(
       this: QB,
-      graph: PartialModelGraph<M>,
-      options?: InsertGraphOptions
-    ): SingleQueryBuilder<QB>;
-
-    <QB extends AnyQueryBuilder>(
-      this: QB,
       graph: PartialModelGraph<M>[],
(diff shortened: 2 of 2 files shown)


rclmenezes:
Wow! I got a `Maximum call stack exceeded` when running `yarn test:typings` on this branch.

```
RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    at /Users/rmenezes/code/objection.js/node_modules/typescript/lib/tsc.js:53002:90
    at Array.map (<anonymous>)
    at recursiveTypeRelatedTo (/Users/rmenezes/code/objection.js/node_modules/typescript/lib/tsc.js:53002:62)
    at isRelatedTo (/Users/rmenezes/code/objection.js/node_modules/typescript/lib/tsc.js:52642:34)
    at typeArgumentsRelatedTo (/Users/rmenezes/code/objection.js/node_modules/typescript/lib/tsc.js:52949:39)
    at relateVariances (/Users/rmenezes/code/objection.js/node_modules/typescript/lib/tsc.js:53490:34)
    at structuredTypeRelatedToWorker (/Users/rmenezes/code/objection.js/node_modules/typescript/lib/tsc.js:53438:46)
    at structuredTypeRelatedTo (/Users/rmenezes/code/objection.js/node_modules/typescript/lib/tsc.js:53080:30)
    at recursiveTypeRelatedTo (/Users/rmenezes/code/objection.js/node_modules/typescript/lib/tsc.js:53051:53)
error Command failed with exit code 1.
```

I noticed your previous struggles with this:
https://github.com/microsoft/TypeScript/issues/33460#issuecomment-534856200

Wow, we're really stretching what Typescript can do, huh? I'm going to give this a think...

rclmenezes:
Closing because Typescript won't let us fix this.

I'm going to try remove custom query builders in a fork and use that instead. See for more info: https://github.com/Vincit/objection.js/pull/2159#issuecomment-985586202

PR #2183, opened 2022-01-03
fix: jsonSchemaWithoutRequired doesn't handle empty definitions

thejuan:
A schema with an empty definitions object causes `jsonSchemaWithoutRequired` to throw

```
 TypeError: Cannot convert undefined or null to object
      at Function.assign (<anonymous>)
      at jsonSchemaWithoutRequired (lib/model/AjvValidator.js:262:20)
      at AjvValidator.compilePatchValidator (lib/model/AjvValidator.js:136:18)
      at AjvValidator.getValidator (lib/model/AjvValidator.js:120:26)
```


lehni:
Released in v3.0.5

PR #2193, opened 2022-01-16
Update to knex 1.0.0

kibertoad:


package.json
```diff
@@ -67,12 +67,12 @@
...
     "eslint-plugin-prettier": "^4.0.0",
     "expect.js": "^0.3.1",
-    "knex": "0.95.13",
+    "knex": "1.0.1",
     "mocha": "^9.1.3",
     "mysql": "^2.18.1",
     "pg": "^8.7.1",
     "prettier": "2.4.1",
-    "sqlite3": "^5.0.2",
+    "@vscode/sqlite3": "^5.0.7",
```
kibertoad, line 75: sqlite3 seems to be more or less abandoned, and there is a known security vulnerability in it, so we were forced to migrate to microsoft fork.

kibertoad:
Will review failing tests and check where tests need to be updated (e. g. to accomodate for different behaviour in RETURNING, which shouldn't require any objection changes, just test update), and where we might have broken something in knex.

ErisDS:
Would it be worth trying `better-sqlite3` here instead? As far as I understand the vscode fork was pure necessity, whereas better-sqlite3 is actively and deliberately maintained.

Either way, would be amazing to see some progress here. We're waiting for knex 1.0 to land to look at making a major migration away from Bookshelf to Objection 😬

mweberxyz:
@kibertoad the RETURNING data structure change seems to be an actual breaking regression (at least in what I am seeing using postgres) beyond just test changes. 

Upon updating to knex latest with objection at 3.0.0, and running my projects tests, I am seeing things like this:

`'select * from "table" where "id" = $1 limit $2 - invalid input syntax for type bigint: "{"id":"15"}"'` -- which seems suspiciously similar to to the data model change.

Are you aware of this yet or would you like a hand trying to track it down?

stefnotch:
I'm guessing this pull request would need to be updated again, since knex version 2 has been released in the meantime.

stefnotch:
I'm guessing this pull request would have to be updated, since knex version 2 has been released in the meantime https://github.com/knex/knex/releases

falkenhawk:
this can be closed, superseded by #2372

Issue #1957, opened 2020-12-30
aliasFor not working when used with query()


patrikwalker:
Hi,

I stumbled upon an issue with `aliasFor`, it doesn't seem to work when used with a query directly, like described in the [docs](https://vincit.github.io/objection.js/api/query-builder/find-methods.html#aliasfor). But everything seems to work fine if used with `$relatedQuery`.

I used the reproduction template to test it (I added an extra property named `someProp`):

```
const jennifer = await Person.query().findOne({ firstName: 'Jennifer' });

// this works fine
const jenniferMovies = await jennifer
  .$relatedQuery('movies')
  .aliasFor('Person_Movie', 'pm')
  .where('pm.someProp', 100);

// this doesn't alias Person_Movie correctly and throws an error because pm is not known
const jenniferWithMovies = await Person.query()
  .findOne({ firstName: 'Jennifer' })
  .aliasFor('Person_Movie', 'pm')
  .joinRelated('movies')
  .where('pm.someProp', 100);
```

This also explains why the test pass: They only test for the case with the `$relatedQuery`.

This should be working, or am I missing something?

ramos-ph:
Hey, there is any update on this one?
I were able to reproduce the problem and I'm investigating what the cause might be.
I'll open a PR to fix this when I find what is causing the issue.

lehni:
Fixed and merged, finally! 🎉

PR #2216, opened 2022-02-05
fix aliasFor not working with join table in many-to-many relation

RedLobster999:
All tests pass.
This fixs https://github.com/Vincit/objection.js/issues/1957.

```
// This is an example from the docs. But if you call this, it would throw an error
// because the alias "pm" is not applied to the query.
await Person.query()
  .aliasFor('persons_movies', 'pm')
  .joinRelated('movies')
  .where('pm.someProp', 100);

```

In case of Many-to-Many relation, builder.joinRelated() uses tablenode.joinTableAlias to get the alias of the join table.
However tablenode.joinTableAlias does not check if an alias for jointable is set. tablenode.joinTableAlias always uses `${joinTableName}_join` as an alias.
so I make a new function "getJoinTableAlias()" which fix the alias problem and replace every "tablenode.joinTableAlias" with "tablenode.getJoinTableAlias()".

Also. I use and love objection.js.

### DIFF
--- a/lib/queryBuilder/join/RelationJoiner.js
+++ b/lib/queryBuilder/join/RelationJoiner.js
@@ -126,7 +126,7 @@ class RelationJoiner {
 
       ownerTable: tableNode.parentNode.alias,
       relatedTableAlias: tableNode.alias,
-      joinTableAlias: tableNode.joinTableAlias,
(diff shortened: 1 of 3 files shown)


lehni:
Great work, thank you @RedLobster999!

Issue #2170, opened 2021-12-13
Error in insert data type date or timestamp


wafaboussada:
Hello
when updating objection to 3.0.0 and knex to 0.95.14, i got this error : type must be JSONType or JSONType[]: date.
and there are warnings in node modules objection

koskimas:
https://github.com/Vincit/objection.js/issues/2147
https://github.com/Vincit/objection.js/issues/2146
https://github.com/Vincit/objection.js/issues/2142

LordA98:
For anyone wondering, if you go down the route of adding the ajv-formats package and creating a property with `type: string` and `format: date-time` - the easiest way to get a value that matches this format is `new Date().toISOString()`.

Wasn't obvious to me so thought I'd add a note.

Issue #2147, opened 2021-11-17
[Objection 3.0] unknown format "iso-date-time" ignored in schema at path


rhazegh:
```
"objection": "^3.0.0"
"knex": "^0.95.14"
"ajv-formats": "^2.1.1"
```

I updated to the latest objection and realized I had to change some models to avoid type errors (for example `date` is now an invalid type in the json schema). I saw https://github.com/Vincit/objection.js/issues/2142 and updated my code to use `ajv-formats` package. But I am still getting this error:

```
unknown format "iso-date-time" ignored in schema at path "#/properties/main/properties/registeredAt"
```

`iso-date-time` is in the list of supported formats according to https://ajv.js.org/packages/ajv-formats.html#formats

Any idea what I am doing wrong?

Here is what my model looks like:

## base-model.js
```javascript
import { Model, QueryBuilder, AjvValidator } from "objection";

const addFormats = require('ajv-formats').default;

class DefaultSchemaQueryBuilder extends QueryBuilder {
  constructor(modelClass) {
    super(modelClass);
    if (modelClass.defaultSchema) {
      this.withSchema(modelClass.defaultSchema);
    }
  }
}

export default class BaseModel extends Model {
  // Add all shared configurations here
  static get QueryBuilder() {
    return DefaultSchemaQueryBuilder;
  }

  static createValidator() {
    return new AjvValidator({
      onCreateAjv: (ajv) => {
        addFormats(ajv);
      },
      options: {
        allErrors: true,
        validateSchema: false,
        ownProperties: true,
        v5: true,
      },
    });
  }
}
```

## user-setting.js
```javascript
import { set } from "lodash";

import BaseModel from "./base-model";
import { languages } from "../constants";

class UserSettingAbstract extends BaseModel {
  static tableName = "user_settings";

  static get jsonSchema() {
    return {
      type: "object",
      properties: {
        id: { type: "string", format: "uuid" },
        userId: { type: "string", format: "uuid" },
        main: {
          type: "object",
          properties: {
            registeredAt: {type: "string", format: "iso-date-time" },
            dailyGoal: { type: "integer" },
            welcomedAt: { type: "string", format: "iso-date-time" },
          },
        },
      },
    };
  }
}

let dynClassMemoizer = {};

function UserSetting(lang) {
  if (!lang) {
    throw new Error(`lang: argument is required`);
  } else if (!languages.map((l) => l.code).includes(lang)) {
    throw new Error(`Invalid language: ${lang}`);
[...]

Issue #2146, opened 2021-11-17
with v3: error unknown format "date-time"


alqu:
I updated objection in my project from v2.2.18 to v3.0.0 and suddenly my models are kind of invalid:
![image](https://user-images.githubusercontent.com/12250845/142207485-37ea83b2-f3ea-4b4e-849e-f5eee12876cb.png)
and there is an error at `passwordUpdatedAt`-field, which has the "date-time"-Format:

```
error: Error in request unknown format "date-time" ignored in schema at path "#/properties/passwordUpdatedAt"
    at unknownFormat (.../node_modules/ajv/lib/vocabularies/format/format.ts:87:15)
    at validateFormat (.../node_modules/ajv/lib/vocabularies/format/format.ts:75:9)
    at Object.code (.../node_modules/ajv/lib/vocabularies/format/format.ts:40:10)
    at keywordCode (.../node_modules/ajv/lib/compile/validate/index.ts:523:9)
    at .../node_modules/ajv/lib/compile/validate/index.ts:265:9
    at CodeGen.code (.../node_modules/ajv/lib/compile/codegen/index.ts:525:33)
    at CodeGen.block (.../node_modules/ajv/lib/compile/codegen/index.ts:680:20)
    at iterateKeywords (.../node_modules/ajv/lib/compile/validate/index.ts:262:7)
    at groupKeywords (.../node_modules/ajv/lib/compile/validate/index.ts:241:7)
    at .../node_modules/ajv/lib/compile/validate/index.ts:233:38
```

I am running a koaJS-app on node v14.18.1. with knex v0.95.14.

Regarding the upgrade instructions at https://vincit.github.io/objection.js/release-notes/migration.html everything should work out of the box - no problems there.

The model was working perfectly in v2.x.
Do you have any clue, what am I doing wrong?

vdechef:
Same probleme here (and here #2147 ). 

Based on #2042 I tried adding **ajv-formats** to enforce the validation: the error disapear, but the validation is not done. I can provide any data and it will still pass the validator.

My object contains a field `theDate` that should be a date (YYYY-MM-DD). But whatever the value of this field, the data is sent to the database: I can do `patch({ theDate: "this_is_not_a_date" })` without validation error.

Here is what I did to enable ajv-formats:
```javascript
import { Model, AjvValidator } from "objection"
import addFormats from('ajv-formats')

export default class MyModel extends Model {
    ...
    // enable ajv formats to  validate the date field when inserting in database
[...]

Issue #2142, opened 2021-11-11
type must be JSONType or JSONType[]: date-time


thefoxie:
After update to the 3.0.0 I got this validation error.

```js
static get jsonSchema() {
		return {
			type: "object",
			required: [
				"first_name",
				"last_name",
				"email",
				"phone_number",
			],
			properties: {
				first_name: { type: "string" },
				last_name: { type: "string" },
				last_change: { type: "date-time" },
				email: { type: "email" },
				phone_number: { type: "string" },
				last_try: { type: "date-time" },
				last_logged: { type: "date-time" },
			},
		};
	}
```
```js
const now = new Date(Date.now());
const lastLogged = response.isValid ? { last_logged: now } : {};

const toPatch = {
	last_try: now,
	...lastLogged,
};

await user.$query().patch(toPatch);
```

koskimas:
`date-time` is not a valid type. It's been working before by accident. I think `Ajv` has added more validation in the recent versions.

koskimas:
Here's the JSON schema spec https://json-schema.org/specification.html

thefoxie:
Thank you, I see now :smile: 

https://ajv.js.org/guide/formats.html

thefoxie:
Is there any way how to use/import `ajv-formats` globally in objection?

thefoxie:
Ok my fault RTFM :smile: 
https://vincit.github.io/objection.js/api/types/#class-ajvvalidator

rhazegh:
@thefoxie Can you please share how you made `ajv-formats` work?

wafaboussada:
> Ok my fault RTFM smile https://vincit.github.io/objection.js/api/types/#class-ajvvalidator

how can i use it in model like this please @thefoxie 
class Cart extends Model {
  // Table name is the only required property.
  static get tableName() {
    return 'cart';
  }

  static get jsonSchema() {
    return {
      type: 'object',
      properties: {
        id: { type: 'integer' },
        history_id: { type: 'integer' },
        currency_id: { type: ['integer', 'null'] },
        modification_date: { type: 'datetime' },
      },
    };
  }
}

thefoxie:
@wafaboussada you have to change it somewhere where you init objectionjs. In my case:

```js
import knex from "knex";
import Objection from "objection";
import addFormats from "ajv-formats";
import config from "./config.js";

const { Model: ObjectionModel, AjvValidator } = Objection;

class Model extends ObjectionModel {
[...]

PR #2218, opened 2022-02-06
Adding ajv-formats package

TigerC10:
Adding the ajv-formats package so people stop complaining about missing date formats.
Solves #2142, #2146, #2147, #2170

### DIFF
--- a/lib/model/AjvValidator.js
+++ b/lib/model/AjvValidator.js
@@ -12,6 +12,14 @@ const getAjv = once(() => {
   }
 });
 
+const getAjvFormats = once(() => {
+  try {
+    return require('ajv-formats');
+  } catch (err) {
+    throw new Error('Optional ajv-formats dependency not installed. Please run `npm install ajv-formats --save`');
+  }
+});
+
 class AjvValidator extends Validator {
   static init(self, conf) {
     super.init(self, conf);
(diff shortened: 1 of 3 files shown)


LordA98:
Hi, just wondered if there was any chance of getting this reviewed and merged?

vichle:
@kibertoad Sorry to bother you, I get that you have a lot on your plate, but I saw you have approved and merged other changes. 

This is a fix for an undocumented breaking change that has been in limbo for almost a year, it would be very appreciated if it could be reviewed and merged :pray: 

Sorry for pinging you :heart:

lehni:
This generally looks good, with one caveat:

We either make `ajv-formats` an optional dependency and keep the current `require()` check with the logged message, or we make it a normal dependency and don't need to check if it can be imported. Right now the code does a mixture of both, which isn't so great.

I'd say we just make it a normal dependency and require without checking or complaining.

lehni:
the code is a bit convoluted. Since the PR adds `ajv-formats` as a dependency, the require checks in `getAjvFormats()` aren't needed at all. It can just be imported normally.

falkenhawk:
I agree, it should be a normal dependency. Ajv used to include formats, but they extracted them to a separate package, starting from v7. re: https://github.com/Vincit/objection.js/issues/2146#issuecomment-974137873

PR #2232, opened 2022-02-22
Add constructor definition for QueryBuilder class for typescript inheritance

ralcorta:
To allow inheritance from the QueryBuilder class and use the constructor method, we must add the constructor definition in index.d.ts

### DIFF
--- a/lib/objection.js
+++ b/lib/objection.js
@@ -40,8 +40,14 @@ function Model() {
   // Nothing to do here.
 }
 
-function QueryBuilder(...args) {
-  NativeQueryBuilder.init(this, ...args);
+// function QueryBuilder(...args) {
+//   NativeQueryBuilder.init(this, ...args);
+// }
+
+class QueryBuilder {
+  constructor(...args) {
+    NativeQueryBuilder.init(this, ...args);
+  }
 }
 
 function Validator(...args) {
--- a/lib/objection.js
+++ b/lib/objection.js
@@ -40,14 +40,8 @@ function Model() {
   // Nothing to do here.
 }
 
-// function QueryBuilder(...args) {
-//   NativeQueryBuilder.init(this, ...args);
-// }
-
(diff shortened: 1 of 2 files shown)


ankitr:
Would love to know what the type signature of the constructor is supposed to be. Looks like it's a `ModelClass`?

ralcorta:
> Would love to know what the type signature of the constructor is supposed to be. Looks like it's a `ModelClass`?

Hi @ankitr ! The idea of this is like this:
I was trying to do a BaseQueryBuilder in Typescript to manage the soft delete for all of my models, and I found that I cannot extend de constructor of QueryBuilder class like this:

```ts
export class BaseQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<M, R> {

  // ...

  constructor(...args: any[]) {
    super(...args);
    this._handleSoftDelete();
  }

  // ... more code

}
```

This works on Javascript, but don't work in Typescript.
The solution was add the `constructor(...args: any[]): this;` on the `QueryBuilder` class ([This change](https://github.com/ralcorta/objection.js/blob/4ef206742caeabddf7981d0fc41d1ea3f22443ad/typings/objection/index.d.ts#L783)), and that was it. With this constructor typed, I guess people will be able to extend more functionalities to the queryBuilder strategy.

ralcorta:
@ankitr  Hello! It was a long time ago, but it's still a good approach. What do you think? I have a package from a fork with this change, good to check.

ralcorta:
@ankitr is this a forgotten repo or issue?

ptoussai:
@ralcorta `QueryBuilder` only has a default constructor, and you should call `super()` without arguments in the derived class. This is supported by the current type definition.

lehni:
There are a lot of unnecessary formatting changes in this PR, making a review difficult. Could you change it so that only the lines that actually need to change are included?

Issue #2023, opened 2021-03-24
Add esm module support


vladshcherbin:
Hey 👋 

Node 12+ supports esm modules, Node 10 end of life is in April which means we can stop using babel, bundlers, etc in node 🥳
Package maintainers are starting to publish new package versions with only esm modules support, e.g. `sindresorhus` with 1k+ packages is going this road ([article](https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77)).

It would be great to see esm module support in objection. Currently some imports/exports don't work:

```js
// works
import { Model } from 'objection'`
console.log(Model)

// doesn't work - throws error
import { snakeCaseMappers } from 'objection'

// works - current workaround
import objection from 'objection'
console.log(objection.snakeCaseMappers)
```

Produced error (node v15.12.0):

```console
SyntaxError: Named export 'snakeCaseMappers' not found. The requested module 'objection' is a CommonJS module,
which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'objection';
const { snakeCaseMappers } = pkg;
```

Since 3.0 is coming with knex 0.95 support, maybe this can be also added in major release.

Thank you 🤗

koskimas:
`Model` and `snakeCaseMappers` are exported identically. Here's the code

```js
module.exports = {
  Model,
  QueryBuilder,
  QueryBuilderBase,
  QueryBuilderOperation,
  RelationExpression,
  ValidationError,
  NotFoundError,
  AjvValidator,
  Validator,
  Relation,
  HasOneRelation,
  HasManyRelation,
  BelongsToOneRelation,
  HasOneThroughRelation,
  ManyToManyRelation,

  transaction,
  initialize,
  compose,
  mixin,
  ref,
  val,
  raw,
  fn,

  get lit() {
    deprecate('`lit` is deprecated. Use `val` instead. `lit` will be removed in 3.0');
    return val;
  },

  snakeCaseMappers,
  knexSnakeCaseMappers,
  knexIdentifierMapping,

  ...DbErrorClasses,
};
```

Do you have any idea why one would work, but not the other?

vladshcherbin:
@koskimas yep, I've also checked current exports and around 85% are present as esm. The error message has `may not support all module.exports as named exports` part but I have no idea how it re-exports only some of them.

koskimas:
Do all after the `lit` getter fail?

vladshcherbin:
[...]

PR #2247, opened 2022-03-18
Make all named exports work for ESM

lehni:
Even though #2023 was already closed, I noticed that the db-errors classes can't be import named in ESM projects:
https://github.com/Vincit/objection.js/issues/2023#issuecomment-1064511975

The problem is the used spread operator. The classes need to be exported explicitly for this to work. this PR fixes the remaining broken exports.


kibertoad (approved):

kibertoad:
Thank you!

PR #2257, opened 2022-04-13
Patch validator must keep required with discriminator

kapouer:
JSON Schema's OpenAPI specification allows some sort of schema inheritance:
https://ajv.js.org/json-schema.html#discriminator

To support this, objection's ajv "patch" validator should keep `required` field when it sits next a `discriminator` keyword.

To keep the change minimal, `ajvNoDefaults.opts.discriminator` is not checked to know if there is actual support for that keyword.


kapouer:
Force-pushed to fix it.

lehni:
LGT, thanks!

kapouer:
Much appreciated

lehni:
Released in v3.0.5

PR #2274, opened 2022-05-12
Update WhereIN TS-defs to use readonly primitive arrays

### LINKED ISSUE #2271: Using typescript readonly variable as a PrimitiveValue
When I try to pass readonly args into `where` syntax, it throws typescript syntax error

`Argument of type 'readonly string[]' is not assignable to parameter of type 'AnyQueryBuilder'.` 

ex)

```ts
function test(ids: readonly string[]) {
  return User.query().whereIn('id', ids)
}
```

Any way to solve this problem?

comp615:
Because doing something like:
```
await knex('workflow').whereIn(myArray)
```
will not modify `myArray`, if we mark the TS-type as ReadOnly, it makes the type more accessible for a variety of use-cases (i.e. readonly arrays can be passed in). Note that mutable arrays can still be cast into read only so this is "backwards compatible" type-wise.

**Fixes #2271**

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -405,7 +405,7 @@ declare namespace Objection {
     <QBP extends QB>(col: ModelProps<ModelType<QBP>>, cb: CallbackVoid<QB>): QB;
     <QBP extends QB>(col: ModelProps<ModelType<QBP>>, qb: AnyQueryBuilder): QB;
 
-    (col: ColumnRef | ColumnRef[], expr: Expression<PrimitiveValue>[]): QB;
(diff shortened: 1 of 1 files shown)


comp615:
@koskimas lmk if you have any questions or qualms here!

comp615:
@koskimas Any way we can get this change in and released? Would be really helpful for typesafe code and should not be functionality impacting so low-risk

comp615:
@kibertoad wanted to flag this as a (hopefully) simple TS def fix for a potential minor version release!

comp615:
@falkenhawk are you the new maintainer? Trying to get this picked up for a future build at some point!

falkenhawk:
I'm not, currently @kibertoad is

lehni:
@comp615 this look good, thanks!

PR #2276, opened 2022-05-14
fix: improved typing for ModelObject<T>

flodlc:
The current ModelObject<T> does not preserve `?` optional properties.  

Example:
```typescript
export class User extends Model {
  id: string;
  name: string;
  parentId?: User
}

function insertUser(user: Omit<ModelObject<User>, "id">) {...}

// Typescript will throw an error here. Missing property parentId
insertUser({
    name: "john"
});

// Unfortunately, I have to manually set parentId: undefined
insertUser({
    name: "john",
    parentId: undefined
});
```

It's not surprising since with this declaration we lose the `?` optional syntax
```typescript
type ModelObject<T extends Model> = {
    [K in DataPropertyNames<T>]: T[K];
};
```
Typescript has a native `Pick<Type, Keys>` utility Type for this which preserves optional properties and allows a shorter syntax.

Not sure if it can be used also for `PartialModelObject`
What do you think ?

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -192,9 +192,7 @@ declare namespace Objection {
   /**
    * A Pojo version of model.
    */
-  type ModelObject<T extends Model> = {
-    [K in DataPropertyNames<T>]: T[K];
-  };
+  type ModelObject<T extends Model> = Pick<T, DataPropertyNames<T>>;
 
   /**
    * Any object that has some of the properties of model class T match this type.


koskimas:
This is the implementation of `Pick`

```ts
type Pick<T, K extends keyof T> = {
    [P in K]: T[P];
};
```

which is 100% the same as what we have.

posgarou:
@koskimas Can this be reopened and considered? @flodlc's solution fixes some type issues I've run into.

The snippet below fails to compile with the current definition of `ModelObject`, but passes with the proposal from @flodlc.

```ts
class Cat extends Model {
  age?: number;
}

type CatObject = ModelObject<Cat>;

const cat = Cat.fromJson({})
const catObject: CatObject = cat;
// Error: Type 'Cat' is not assignable to type 'CatObject'.
// Property 'age' is optional in type 'Cat' but required in type 'CatObject'
```

lehni:
I'm going with @falkenhawk's suggestion here and go ahead and merge this: https://github.com/Vincit/objection.js/pull/2299#issuecomment-1511569944

PR #2307, opened 2022-07-25
Updates types for TS 4.8

sandersn:
Typescript 4.8 assumes that unconstrained types extend `unknown`, not `{}` (which is `unknown` minus `null` and `undefined`). However, many uses still need to extend `{}`. This PR does that explicitly for objection's types.

See https://devblogs.microsoft.com/typescript/announcing-typescript-4-8-beta/#unconstrained-generics-no-longer-assignable-to

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -1008,7 +1008,7 @@ declare namespace Objection {
     timeout: TimeoutMethod<this>;
     columnInfo: ColumnInfoMethod<this>;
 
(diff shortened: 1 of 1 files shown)


QholmesBRD:
This would be very helpful

kibertoad (approved):

jorgebv:
@kibertoad thank you for merging the PR. Do you have the ability to make a release containing this fix, or do you know if a release is planned in the near future?

Thank you and other maintainers for all the work you do on this and other projects.

PR #2323, opened 2022-10-03
Make relation join error due to lack of initialization a bit more explicit

max-kahnt-keylight:
Make relation joiner error a bit more explicit.


lehni:
Makes sense, thanks!

Issue #2192, opened 2022-01-14
knex.migrate.latest() fails when  ...knexSnakeCaseMappers is part of the knex config


cauboy:
I'm having the following issue:

When trying to do migration with an "knexSnakeCaseMappers"-ized knex instance using `"objection": "3.0.1"` and `knex": "^0.95.15"`, the migration fails with the following error:

```
migration file "1.ts" failed
migration failed with error: Cannot read properties of undefined (reading 'lastIndexOf')

 TypeError {
    message: 'Cannot read properties of undefined (reading \'lastIndexOf\')',
  }

  › node_modules/objection/lib/utils/identifierMapping.js:139:21
  › node_modules/objection/lib/utils/identifierMapping.js:13:16
  › Object.wrapIdentifier (node_modules/objection/lib/utils/identifierMapping.js:180:23)
  › Client_SQLite3.customWrapIdentifier (node_modules/knex/lib/client.js:177:26)
  › TableCompiler_SQLite3.primary (node_modules/knex/lib/dialects/sqlite3/schema/sqlite-tablecompiler.js:192:34)
  › TableCompiler_SQLite3.alterTable (node_modules/knex/lib/schema/tablecompiler.js:245:32)
  › TableCompiler_SQLite3.create (node_modules/knex/lib/schema/tablecompiler.js:59:10)
  › TableCompiler_SQLite3.toSQL (node_modules/knex/lib/schema/tablecompiler.js:39:22)
  › TableBuilder.toSQL (node_modules/knex/lib/schema/tablebuilder.js:49:44)
  › SchemaCompiler_SQLite3.build (node_modules/knex/lib/schema/compiler.js:143:23)
```

## How to reproduce:

The error is thrown when running `ts-node migrate.ts`

```ts
// migrate.ts
import { db } from '../src/db
db.migrate.latest() // <--- This command fails
```

```ts
// db.ts
import { knexSnakeCaseMappers } from 'objection'
export const db = knex({
  client: 'sqlite3',
  migrations: { tableName: 'knex_migrations', },
  connection: { filename: './sqlite.db' },
  useNullAsDefault: true,
  ...knexSnakeCaseMappers(),
})
```

```ts
// migrations/1.ts
import { Knex } from 'knex'
export function up(knex: Knex) {
  return knex.schema.createTable('example', (table) => {
    table.uuid('id').primary()
  })
}

export function down(knex: Knex) {
  return knex.schema.dropTable('example')
}
```

## Possible solution
[...]

PR #2333, opened 2022-10-27
fix issue 2192

tinchoz49:
Hey Everyone, this PR fix the issue: https://github.com/Vincit/objection.js/issues/2192 based on the solution that was proposed there.

The only difference is to avoid to check the `typeof` for performance reasons. Empty strings | null | undefined will return the same.

### DIFF
--- a/tests/integration/knexSnakeCase.js
+++ b/tests/integration/knexSnakeCase.js
@@ -4,7 +4,7 @@ const { Model, knexSnakeCaseMappers } = require('../../');
 const { expect } = require('chai');
 
 module.exports = (session) => {
-  describe.only('knexSnakeCaseMappers', () => {
+  describe('knexSnakeCaseMappers', () => {
     let knex;
 
     class Person extends Model {
@@ -185,9 +185,9 @@ module.exports = (session) => {
         return knex.schema
           .dropTableIfExists('emptyConstraintName')
           .createTable('emptyConstraintName', (table) => {
-            table.integer('id').primary()
-          })
-      })
+            table.integer('id').primary();
+          });
+      });
     });
 
     describe('queries', () => {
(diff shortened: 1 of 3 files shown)


tests/integration/knexSnakeCase.js
```diff
@@ -4,7 +4,7 @@ const { Model, knexSnakeCaseMappers } = require('../../');
 const { expect } = require('chai');
 
 module.exports = (session) => {
-  describe('knexSnakeCaseMappers', () => {
+  describe.only('knexSnakeCaseMappers', () => {
```
aromeronavia, line 7: I think you need to remove the only here?

aromeronavia (commented):
I'd love this PR to get fast forwarded, it'd be really useful for my project to setup a sqlite database for test environment.

tests/integration/knexSnakeCase.js
tinchoz49: oh this embarrassing :smile: I totally forget about that `only`.

aromeronavia: No worries at all :P

deshiknaves:
Bump. This seems to have gone stale. There seems to be no work around for this. Could we get this fix in?

kibertoad (approved):

kibertoad:
Thanks!

deshiknaves:
@kibertoad thanks for merging this. Looking forward to the release, any idea when that might be?

kibertoad:
asked @koskimas for publish permissions, can't release myself yet

aromeronavia:
Thanks for merging this!

openam:
Doesn't look like this was ever released. Should this have resulted in a 3.0.2 release?

PR #2338, opened 2022-11-04
Allow type Raw inside returning()

### LINKED ISSUE #2337: Typescript problem with raw inside returning() function
Here is the issue:

```js
await Address.query()
      .insert(data)
      .returning([
        'id',
        'userId',
        'postalCode',
        'street',
        'city',
        'countryCode',
        'createdAt',
        'updatedAt',
        raw('ST_AsGeoJSON(??) AS ??', ['geoPoint', 'geoPoint']) // Typing error !
      ]);
```

Same issue seemed to be sloved in knex:

https://github.com/knex/knex/issues/4879

I guess something similar could be done.

OlivierCuyp:
Fix #2337

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -627,7 +627,7 @@ declare namespace Objection {
   interface ReturningMethod {
     <QB extends AnyQueryBuilder>(
       this: QB,
-      column: string | string[]
+      column: string | Raw | (string | Raw)[]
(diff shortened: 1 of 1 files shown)


OlivierCuyp:
@koskimas could you please validate this PR, it is only typing and only some characters long ?
This spare me and others comments in code to disable eslint and typescript check.

lehni:
This makes sense and reflects Knex tyings, as you pointed out. Merging!

PR #2372, opened 2023-03-28
adjustments for knex v1 and v2

falkenhawk:
_ported from https://github.com/ovos/objection.js/pull/15_

### adjustments for knex v1 and v2

knex v1+ changed the return values when `.returning()` clause is used (affecting postgres), before it returned array of primitive values if only 1 column was selected to be returned (e.g. `[1]` for `.returning('id')`) but now it always returns objects (e.g. `[{ id: 1 }]` for `.returning('id')`).

Slightly reorganized `InsertOperation`, instead of calling `createModels` in `doExecute()` (right after `onRawResult` hook) and having "empty" model instances created there, possibly with undefined / declared default properties which would later overwrite input models in `InsertOperation.onAfter1`, offload the handling to `InsertOperation.onAfter1`, which now uses also `setDatabaseJson`, same as `createModels` function, to be consistent.

Additionally, adjusted `$parseDatabaseJson` in `GraphInsert` integration tests, to only parse boolean values when they are present in `json`.

related change in knex v1.0: https://github.com/knex/knex/pull/4471

### introduced new (or missing) knex querybuilder's methods

to make shapes of both querybuilders as close to each other as possible.
Most of the methods were introduced in knex v1.
- `fromRaw`
- `whereLike`
- `andWhereLike`
- `orWhereLike`
- `whereILike`
- `andWhereILike`
- `orWhereILike`
- `withMaterialized`
- `withNotMaterialized`
- `jsonExtract`
- `jsonSet`
- `jsonInsert`
- `jsonRemove`
- `whereJsonObject`
- `orWhereJsonObject`
- `andWhereJsonObject`
- `whereNotJsonObject`
- `orWhereNotJsonObject`
- `andWhereNotJsonObject`
- `whereJsonPath`
- `orWhereJsonPath`
- `andWhereJsonPath`

and one new `JoinBuilder` method - `andOnJsonPathEquals` (referencing knex's `JoinClause` method of the same name)

related: https://github.com/knex/knex/pull/4859 https://github.com/knex/knex/pull/5044

`whereJson(Not)SupersetOf` / `whereJson(Not)SubsetOf` are now supported by knex >= 1.0, but for now
objection handles them differently and only for postgres.
Changing them to utilize knex methods directly may require a major version bump and upgrade guide.

Added a test for checking JoinBuilder methods to be inline with knex's JoinClause.

Fixed incorrect checking QueryBuilder methods if they are inline with knex's QueryBuilder.

Added typings for new methods.

Bumped knex `peerDependency` from `>=0.95.13` to `>=1.0.1` because of new methods.

### DIFF
--- a/lib/queryBuilder/JoinBuilder.js
+++ b/lib/queryBuilder/JoinBuilder.js
@@ -120,6 +120,10 @@ class JoinBuilder extends QueryBuilderOperationSupport {
     return this.addOperation(new KnexOperation('andOnNotBetween'), args);
   }
 
+  andOnJsonPathEquals(...args) {
+    return this.addOperation(new KnexOperation('andOnJsonPathEquals'), args);
+  }
+
   onVal(...args) {
     return this.addOperation(new KnexOperation('onVal'), args);
   }
--- a/lib/queryBuilder/QueryBuilder.js
+++ b/lib/queryBuilder/QueryBuilder.js
@@ -1274,11 +1274,16 @@ async function chainHooks(promise, builder, func) {
 }
 
 function createModels(result, builder) {
-  const modelClass = builder.resultModelClass();
-
   if (result === null || result === undefined) {
     return null;
   }
+
+  if (builder.isInsert()) {
+    // results are applied to input models in `InsertOperation.onAfter1` instead.
+    return result;
+  }
+
+  const modelClass = builder.resultModelClass();
 
   if (Array.isArray(result)) {
     if (result.length && shouldBeConvertedToModel(result[0], modelClass)) {
--- a/lib/queryBuilder/QueryBuilderBase.js
+++ b/lib/queryBuilder/QueryBuilderBase.js
@@ -105,6 +105,10 @@ class QueryBuilderBase extends QueryBuilderOperationSupport {
     return this.addOperation(new KnexOperation('fromJS'), args);
   }
 
+  fromRaw(...args) {
+    return this.addOperation(new KnexOperation('fromRaw'), args);
+  }
+
   into(...args) {
     return this.addOperation(new KnexOperation('into'), args);
   }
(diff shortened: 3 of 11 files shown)


kibertoad:
I'll publish this as a new semver major after releasing minor tomorrow.
Thank you!

falkenhawk:
@kibertoad perhaps for a major version bump it would be worth it to rework `whereJson(Not)SupersetOf / whereJson(Not)SubsetOf` methods to utilize knex counterparts directly? 🤔 That would require changes to function signatures and updates in docs / migration guide.

In this PR the only kind-of-breaking change is minimal knex dependency bump from 0.95 to 1.0.1. While it might still work fine with knex 0.95 unless you start using the new json methods.

kibertoad:
@falkenhawk I'm open to using this semver major window for addressing that. Could you please send the PR?

falkenhawk:
@kibertoad I am not sure if I can make it. Meanwhile I updated knex-related docs in https://github.com/Vincit/objection.js/pull/2375

falkenhawk:
> rework `whereJson(Not)SupersetOf / whereJson(Not)SubsetOf` methods to utilize knex counterparts directly? 🤔 Though that would require changes to function signatures and updates to docs / migration guide.

@kibertoad I made an attempt, but as I read through https://vincit.github.io/objection.js/api/query-builder/find-methods.html#wherejsonsupersetof there is an example to reference other json field in a where clause as a value.

```typescript
const people = await Person.query().whereJsonSupersetOf(
  'additionalData:myDogs',
  'additionalData:dogsAtHome'
);
```

I am not sure if the same is supported in knex, docs does not say anything about it:
https://knexjs.org/guide/query-builder.html#wherejsonsupersetof

It would be nice to have those json functions available also for other clients, other than postgres, which knex currently supports, but I am afraid of introducing an unnecessary regression, making it a blocker to update to new version of objection.js for postgres users who currently use those functions that way (referencing json fields as values)

Maybe you know more about the internals, or may @OlivierCavadenti (who implemented https://github.com/knex/knex/pull/4859) shed some light here, please?

falkenhawk:
@kibertoad from my investigations it won't be trivial to switch `whereJson(Not)SupersetOf` / `whereJson(Not)SubsetOf` directly without limiting current functionality.

Even if "field expressions" - https://github.com/Vincit/objection.js/blob/master/lib/queryBuilder/operations/jsonApi/postgresJsonApi.js#L45-L51 - were translated to combinations of column names + `jsonExtract()` to reference a field inside json value, knex currently does not support passing querybuilder instances as arguments to `whereJson(Not)SupersetOf` / `whereJson(Not)SubsetOf` querybuilder methods, such as `.whereJsonSupersetOf('column', q => q.jsonExtract('othercolumn', '$.field.0'))`

In my case it is getting stuck in infinite loops, but maybe I tested it wrong. Otherwise changes to knex codebase would be needed to make it work, i.e. allowing to migrate such json queries while keeping all current features on objection's side.

That said, I am not able to dig deeper, as I don't have too much time on my hands currently. I don't want to hold you back there, Feel free to proceed with the release, and in the future, when someone approaches this again, we could have yet another major version bump ;)

PR #2375, opened 2023-03-31
update knex-related parts in docs

falkenhawk:
- update existing links to knex docs
- update docs to include knex v1+ methods added in #2372

docs might still be not in sync with knex, probably not all existing methods are documented, but I could not spend more time on it. A tool to automatically check them and synchronize would be handy 😉


kibertoad (approved):

kibertoad:
Thanks!

PR #2379, opened 2023-04-07
Typings: Add knex() method type to QueryBuilder

pigpudle:
Objection.js allows us to get Knex instance from QueryBuilder by using `.knex()` method like this:

```javascript
await ModelName.query(trx).knex().raw(query);
```

Source code: https://github.com/Vincit/objection.js/blob/master/lib/queryBuilder/QueryBuilderOperationSupport.js#L194

But this method is not defined in types definitions. Thus TypeScript shows an error like `Property 'knex' does not exist on type 'QueryBuilder<ModelName, ModelName[]>'` when using the above code.

This PR adds `.knex()` method definition to QueryBuilder types.


lehni:
Looks good, thank you @pigpudle!

PR #2383, opened 2023-04-14
Add missing whereNotInComposite() typing

lehni:


lehni:
@kibertoad since I am an admin now on the repo I could merge these directly, but I'm still a bit reluctant to do so. How should we go about it?

PR #2397, opened 2023-04-15
Remove unneeded instanceof check in withGraphFetched()

### LINKED ISSUE #2251: Overwritten fromDatabaseJson does not fetch relations
Objectionjs version: 3.0.1

I am referencing the same table in several models and followed the resolution of the issue #1674.

This [comment](https://github.com/Vincit/objection.js/issues/1674#issuecomment-584619332) from @koskimas looked exactly like what I needed to have a generic model and fetch specific models from it.

> You should be able to override the static query method like this
> 
> ```js
> class Message extends Model {
>   static get tableName() {
>     return 'posts'
>   }
> 
>   static query(...args) {
>     return super.query(...args).where('post_type', 'message')
>   }
> }
> ```
> 
> Model instances are always created using either fromJson or fromDatabaseJson methods. You should be able to override those. Something like this
> 
> ```js
> class Post extends Model {
>   static get tableName() {
>     return 'posts';
>   }
> 
>   static fromDatabaseJson(json) {
>     if (json.post_type === 'message') {
>       return Model.fromDatabaseJson.call(Message, json)
>     } else if (json.post_type === 'blog') {
>       return Model.fromDatabaseJson.call(Blog, json)
>     } else {
>       return Model.fromDatabaseJson.call(Post, json)
>     }
>   }
> }
> ```
> The Model.fromDatabaseJson.call trick is needed to not invoke the inherited fromDatabaseJson method again causing an infinite recursion. You could also pass a second argument indicating the case or whatever you can think of.
> 
> Let me know if this works.

It works perfectly except when I add a relation to the Message model.

Then the query `Post.query().select().withGraphFetched('user')` returns a Message with the field `user: undefined`.
Though, the query `Message.query().select().withGraphFetched('user')` returns the Message with the field user filled.

I manage to track down the issue to the method `onAfter2` in [WhereInEagerOperation.js](https://github.com/Vincit/objection.js/blob/master/lib/queryBuilder/operations/eager/WhereInEagerOperation.js).

The check `!(models[0] instanceof modelClass)` below return false because Message is not an instance of Post.
So the relations are not fetched.
```js
async onAfter2(builder, result) {
    const modelClass = builder.resultModelClass();

    if (!result) {
      return result;
    }

    const models = asArray(result);

    if (!models.length || !(models[0] instanceof modelClass)) {
      return result;
    }

    await promiseUtils.map(
      this.relationsToFetch,
      (it) => this.fetchRelation(builder, models, it.relation, it.childExpression),
      { concurrency: modelClass.getConcurrency(builder.unsafeKnex()) }
    );

    const intOpt = builder.internalOptions();

    if (!this.omitProps.length || intOpt.keepImplicitJoinProps) {
      return result;
    }

    // Now that relations have been fetched for `models` we can omit the
    // columns that were implicitly selected by this class.
    for (let i = 0, l = result.length; i < l; ++i) {
      const model = result[i];

      for (let c = 0, lc = this.omitProps.length; c < lc; ++c) {
        modelClass.omitImpl(model, this.omitProps[c]);
      }
    }

    return result;
  }
```

I would like to have your opinion on this issue to see if we should consider it as a bug or propose a workaround.

Thank you in advance for your help ;)

lehni:
Closes #2251

### DIFF
--- a/lib/queryBuilder/operations/eager/WhereInEagerOperation.js
+++ b/lib/queryBuilder/operations/eager/WhereInEagerOperation.js
@@ -63,7 +63,7 @@ class WhereInEagerOperation extends EagerOperation {
 
     const models = asArray(result);
 
-    if (!models.length || !(models[0] instanceof modelClass)) {
+    if (!models.length) {
       return result;
     }
 


comp615:
@lehni This commit seems to have caused a regression for us in 3.0.3 and I think should be reverted (or maybe our whole pattern is wrong).

We use a custom query builder to always eager load an association:

```
// The custom query builder always eager loads the assigned user roles, which
// ensures that we're always able to access the userRole virtual attribute
class UserQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<M, R> {
  execute() {
    void this.withGraphFetched('assignedUserRoles');
    return super.execute();
  }
}
```

However in a few places and tests, we do something like:

```
await User.query(txn)
        .where({ id: myUser.id })
        .patch({ status: 'disabled' });
```

But in 3.0.3 this started throwing:
```
 DataError: select "assigned_user_role".* from "assigned_user_role" where "assigned_user_role"."userId" in ($1) - invalid input syntax for type uuid: "1"
```

The userId is obviously defined as a uuid. I was able to get around this by changing the execute override to the following. But I'm not confident that this won't break other usages like a patch and select.
```
class UserQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<M, R> {
  execute() {
    if (!this.isUpdate()) {
      void this.withGraphFetched('assignedUserRoles');
    }
    return super.execute();
  }
}
```

Was curious if you could lend some sage advice on the issue. Thanks!

lehni:
@comp615 oh that doesn't sound good! Are you certain that the regression is caused by this commit? Could you open a new issue for this problem, and provide a reproduction for it, perhaps using the `reproduction-template.js` file if possible? That would be very useful to debug the source of your issue. I don't fully grasp how the two things are related from your description. Thanks!

lehni:
@comp615 I've been trying to replicate this and write a test for it, but I can't reproduce it. A reproduction of any kind would be really welcome.

lehni:
@comp615 I figured it out now, fix coming up: #2476

lehni:
Released in v3.0.5

Issue #625, opened 2017-11-17
Discussion: TypeScript plugin/mixin woes


kblomster:
Implementing and/or using Objection plugins in TypeScript is kinda messy, at the moment. Let's try writing a plugin in TS 2.6.1 with the current typings:

```typescript
class PluginBase extends objection.Model {
  static classProp: number;
  baseProp: string;
  overridableMethod (arg: number) {
    return arg.toString();
  }
}

const plugin: objection.Plugin = (baseClass) => {
  return class extends baseClass {
    static pluginClassProp: number;
    pluginProp: string;
    overridableMethod (arg: string) {
      return Number(arg);
    }
  };
};
```

This doesn't work at all and causes two separate errors:
```
Type '<M extends typeof Model>(baseClass: M) => typeof (Anonymous class)' is not assignable to type 'Plugin'.
Type 'typeof (Anonymous class)' is not assignable to type 'M'."
```
```
Type 'M' is not a constructor function type.
```

Stealing an idea from the `Mixin` declaration, we try a different plugin function signature, and it starts looking promising:
```typescript
// note that we do _not_ explicitly declare the return type here
const plugin = <MC extends objection.ModelClass<any>>(baseClass: MC) => {
  return class extends baseClass {
    static pluginClassProp: number;
    pluginProp: string;
    overridableMethod (arg: string) {
      return Number(arg);
    }
  };
};

class PluggedIn extends plugin(PluginBase) {}

// all of this typechecks:
let instanceProp: string = new PluggedIn().pluginProp;
instanceProp = new PluggedIn().baseProp;
let classProp: number = PluggedIn.classProp;
classProp = PluggedIn.pluginClassProp;
const methodResult: number = new PluggedIn().overridableMethod('1');
```

But then we run into trouble anyway, because `mixin()` still doesn't quite work like we want:
```typescript
class MixedIn extends objection.mixin(PluginBase, [plugin]) {}
// The line below doesn't compile: Property 'pluginProp' does not exist on type 'MixedIn'.
let instanceProp: string = new MixedIn().pluginProp;
```

And, well, of course it doesn't. The re-defined `plugin` example above only works because we let the compiler infer the return type for us. A plugin function does not usually return its model class argument unchanged, like the current typing implies. If we were to declare the return type of our `plugin` above explicitly, we'd first have to declare interfaces for both the static and the instance side of the anonymous extension class. The return type then becomes the intersection of the argument and the static interface, like so:
```typescript
interface PluginInstance {
[...]

PR #2399, opened 2023-04-17
Generic constructor type.

max-kahnt-keylight:
```
import * as O from 'objection';

export const TestMixin = <M extends O.ModelClass<O.Model>>(Model: M) =>
  class extends Model {
    additionalProperty: string;

    static get modifiers() {
      return {
        ...super.modifiers,
        additionalModifier: <QB extends O.QueryBuilder<O.Model>>(qb: QB) => qb.where({ id: 0 }),
      };
    }
  };
```
fails with error
`error TS2545: A mixin class must have a constructor with a single rest parameter of type 'any[]'.` for Typescript 4.8.4.

With this fix, it compiles successfully.
This also resolves https://github.com/Vincit/objection.js/issues/625 for me.

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -1417,9 +1417,9 @@ declare namespace Objection {
     table?: string;
   }
 
-  export interface Constructor<T> {
-    new (): T;
-  }
+  export type Constructor<TResult, TParams extends any[] = any[]> = new (
+    ...params: TParams
+  ) => TResult;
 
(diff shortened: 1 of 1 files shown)


lehni:
Perfect, let's merge it!

lehni:
@max-kahnt-keylight there are issues with your fix, see #2531. Could you help solving them? I will have to revert this change for the next release otherwise.

Issue #1205, opened 2019-01-25
Subqueries break upgrading from 1.4.0 to 1.5.3


nazar:
Greetings,

Subqueries using `.with` are no longer working as expected when I upgrade from `1.4.0` to `1.5.3`.

```javascript
  return Model
    .query()
    .with(
      'pager_subquery',
      query
    )
    .count()
    .from('pager_subquery')
    .first();
```

The above code takes a `query`, which is an objection model `.query()` and uses it to construct a subquery using `.with` to return a basic row count.

Under `1.4.0`, the following correct SQL is generated:

```sql
with "pager_subquery" as (select "posts".* from "posts" where "posts"."postable_id" in ($1) and "postable_type" = $2 and "posts"."deleted_at" is null order by "stats"#>'{scores,low}' desc) select count(*) from "pager_subquery"
```

Under `1.5.3`, the following incorrect SQL is generated:

```
with "pager_subquery" as (select "pager_subquery".* from "pager_subquery" where "pager_subquery"."postable_id" in ($1) and "postable_type" = $2 and "posts"."deleted_at" is null order by "stats"#>'{scores,low}' desc) select count(*) from "pager_subquery"
```

It appears that under `1.5.3`, the original `query` object loses its model `tablename` and/or is overridden by the `.from('pager_subquery')` method call.

I had a look through the change log and couldn't find anything that indicates that this behaviour has changed.

Is this a bug or has Objection changed this behaviour? If the latter, any hints or clues would appreciated.

Libraries:

* knex: `0.16.3`
* pg: `7.8.0`
* db: Postgres 10.5

koskimas:
This i probably a bug in objection. I'm on vacation for three weeks. I'll look into this after I'm back

nazar:
Thank you for this @koskimas! And super thanks for objection.js - words cannot describe how much I ❤️it and how pleasurable it is to use.

koskimas:
objection 1.6 is now released with a fix for this issue.

PR #2405, opened 2023-04-22
fix: losing aliasMap in 'partial' subqueries

falkenhawk:
_ported from https://github.com/ovos/objection.js/pull/13_

This fixes an issue with 'partial' subqueries that they lose `aliasMap` from `q._context` (QueryBuilder's "internal context") while generating nested `where` sql part. `QueryBuilder.subqueryOf()` used to grab context from "parent" query, but it doesn't anymore. (since 2019-02-13) https://github.com/Vincit/objection.js/commit/524e0bef5f75a280aa54d1ee30b6b36c304ea6ad

In our case, we were getting SQL errors where alias was necessary to make it work, but it was missing
```
ER_NON_UNIQ_ERROR: Column 'event_id' in IN/ALL/ANY subquery is ambiguous
```

Basically reintroducing what was deleted in https://github.com/Vincit/objection.js/commit/524e0bef5f75a280aa54d1ee30b6b36c304ea6ad (to fix https://github.com/Vincit/objection.js/issues/1205) but wrap merging context with a condition that it should only apply to "partial" subqueries + reverse the sequence that `isPartial()` is set before `subqueryOf()`

### DIFF
--- a/lib/utils/objectUtils.js
+++ b/lib/utils/objectUtils.js
@@ -377,7 +377,7 @@ function mergeMaps(map1, map2) {
   const map = new Map(map1);
 
   if (map2) {
-    for (let key of map2.keys()) {
+    for (const key of map2.keys()) {
       map.set(key, map2.get(key));
     }
   }
--- a/lib/utils/objectUtils.js
+++ b/lib/utils/objectUtils.js
@@ -373,6 +373,18 @@ function isSafeKey(key) {
   return isNumber(key) || (isString(key) && key !== '__proto__');
 }
 
+function mergeMaps(map1, map2) {
+  const map = new Map(map1);
+
+  if (map2) {
+    for (let key of map2.keys()) {
+      map.set(key, map2.get(key));
+    }
+  }
+
+  return map;
+}
+
 module.exports = {
   isEmpty,
   isString,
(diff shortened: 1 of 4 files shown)


lib/utils/objectUtils.js
```diff
@@ -373,6 +373,18 @@ function isSafeKey(key) {
   return isNumber(key) || (isString(key) && key !== '__proto__');
 }
 
+function mergeMaps(map1, map2) {
+  const map = new Map(map1);
+
+  if (map2) {
+    for (let key of map2.keys()) {
```
glensc, line 380:
```suggestion
    for (const key of map2.keys()) {
```

falkenhawk:
@lehni I've added a test which fails with
```
DBError: select `m1`.* from `model2` as `m1` where (`id_col` = (select `m2`.`id_col` from `model2` as `m2` where `m2`.`model2_prop2` = `model2`.`model2_prop2`)) order by `m1`.`model2_prop2` asc - SQLITE_ERROR: no such column: model2.model2_prop2
```
on the master branch.

Perhaps the `${query.tableRefFor(query.modelClass())}.model2_prop2` reference look weird, but it is useful e.g. for modifiers which add a where condition with subquery, referencing a column from parent query, but the alias which parent query uses is not known and cannot be hardcoded in the modifier.

PR #2406, opened 2023-04-22
mysql: ManyToMany relations - do not use autogenerated subqueries where not necessary, avoid ER_CANT_UPDATE_USED_TABLE_IN_SF_OR_TRG error coming from mysql when triggers are attached to join table

falkenhawk:
_ported from https://github.com/ovos/objection.js/pull/11_

- targeting mysql only
- if subquery is not needed at all for queries on m2m relations (e.g. a query with just a findById(s) operation - usually coming from graph upsert) - skip it
- otherwise extract a subquery reading related ids to separate query run before the delete query for m2m unrelate operation
  - it acts as a workaround for `ER_CANT_UPDATE_USED_TABLE_IN_SF_OR_TRG` mysql error when a trigger on join table was operating on the related table


lehni:
This looks solid at first sight. I Do wonder though if this should be structured the same way as the library handles "anomalies" in Sqlite, using `ManyToManySqliteModifyMixin` and the sub-classes `ManyToManyDeleteSqliteOperation`, `ManyToManyUnrelateSqliteOperation` and `ManyToManyUpdateSqliteOperation`?

This would seem cleaner to me than starting to also branch inside `ManyToManyModifyMixin`.

Do you agree @falkenhawk, and would you be willing to work on this?

PR #2407, opened 2023-04-22
fix for WrapMysqlModifySubqueryTransformation when using join builder

falkenhawk:
_ported from https://github.com/ovos/objection.js/pull/10_

Fixes error where `.onIn()`/`.andOnIn()` causes:
`TypeError: parentQuery.isUpdate is not a function`

This happens when we provide `.onIn()` with an objection query.

```typescript
BlogPost.query(db)
  .alias('post')
  .innerJoin('post.comment', (join) => {
    join.onIn(
      'comment.author_id',
      Author.query(db).where('active', true)
    );
  });
```

Since `WrapMysqlModifySubqueryTransformation` does not make sense for join query, we simply skip it for `JoinBuilder`.

### DIFF
--- a/lib/queryBuilder/transformations/WrapMysqlModifySubqueryTransformation.js
+++ b/lib/queryBuilder/transformations/WrapMysqlModifySubqueryTransformation.js
@@ -2,6 +2,8 @@
 
 const { QueryTransformation } = require('./QueryTransformation');
 const { isMySql } = require('../../utils/knexUtils');
+const { once } = require('../../utils/objectUtils');
+const getJoinBuilder = once(() => require('../JoinBuilder').JoinBuilder);
 
 /**
  * Mysql doesn't allow queries like this:
@@ -27,6 +29,13 @@ class WrapMysqlModifySubqueryTransformation extends QueryTransformation {
 
     // This transformation only applies to MySQL.
     if (!isMySql(knex)) {
+      return query;
+    }
+
+    // This transformation should not apply to join builder, otherwise it causes:
+    // "TypeError: parentQuery.isUpdate is not a function"
+    const JoinBuilder = getJoinBuilder();
+    if (parentQuery instanceof JoinBuilder) {
       return query;
     }
 


falkenhawk:
I've added a typings test for join.onIn + subquery case, but then I realized objection's typings currently do not allow that, because `Knex.JoinClause` is referenced directly
https://github.com/Vincit/objection.js/blob/23b8ac112ff095b5d6d8ac2f60b9af87d9028a77/typings/objection/index.d.ts#L611

and that, in turn, has only those variants listed for `onIn` methods (allowing any[] | Raw for second argument)

https://github.com/knex/knex/blob/b6d04f75b3c1956be8ffa26740dd1e27909dd25d/types/index.d.ts#L1586-L1591

and I remembered, it just worked for us (joins with onIn + subquery) when we `@ts-ignore`'d that.
But in objection v3 we started getting those errors mentioned above in the PR description, that's why we did that change to the objection's code, while still having those onIns ts-ignored.

Since those are proven to work just fine, it would be better to extend the JoinClause typings in https://github.com/Vincit/objection.js/blob/23b8ac112ff095b5d6d8ac2f60b9af87d9028a77/typings/objection/index.d.ts and add tests for such join + onIn variants somewhere along those lines https://github.com/Vincit/objection.js/blob/a7186eb40061afddff7a1e56c21643170fbbbdd7/tests/integration/find.js#L183
(there aren't many test cases for `.*join()` methods anyway)

lehni:
@falkenhawk could you look into why this is currently failing on Node 16?

falkenhawk:
They fail because of typing errors coming from `tsc`,
```
Error: tests/ts/query-builder-api/join-methods.ts(64,29): error TS2345: Argument of type 'QueryBuilder<Person, Person[]>' is not assignable to parameter of type 'readonly any[] | Raw<any>'.
  Type 'QueryBuilder<Person, Person[]>' is missing the following properties from type 'Raw<any>': wrap, toSQL, queryContext, addListener, and 22 more.
```
because , as I mentioned in my previous comment, `Knex.JoinClause` is referenced directly for all join methods, so typings do not accept objection's QueryBuilder to be passed to any join method's argument, but it is handled by objection and works fine.

To make it right, typings need a rewrite on objection's side to allow objection's QueryBuilder in places where knex's QueryBuilder is accepted. Tests are also missing for such cases... I think they were left out to trust knex's test suite that join methods work instead.

lehni:
@falkenhawk thank you for the explanation, and apologies for initially not reading the full description thoroughly. Would you be willing to work on the missing types? I'd be OK with the missing tests.

PR #2435, opened 2023-05-30
add objectionjs-graphql package to plugins documentation

firassziedan:
[objectionjs-graphql](https://www.npmjs.com/package/objectionjs-graphql) it's an updated version to support GraphQl


lehni:
@firassziedan this is interesting! Shall we try and get your changes merge to the upstream one? How is compatiblity?

firassziedan:
@lehni this will be great, this needs a major update there are a lot of changes that need to do, but it's possible.

PR #2452, opened 2023-07-04
Support $beforeUpdate() mutations in upsertGraph()

### LINKED ISSUE #2233: Can't add or modify a value during $beforeUpdate using upsertGraph
We are trying to push some of calculated logic into the models using instance hooks, there is a case of encryption and decryption as well. This was working correctly with objection 1 but started showing a different behavior with objection 3. The problem is that since objection is doing a diff right before $beforeUpdate is invoked, the values changed in the function don't get picked up as a part of the `propsToUpdate` and they are not executed in the sql query. I tried to bypass this or somehow tried to change this value. The only thing that work was iterating over $$omitFromDatabaseJson and making sure that the field I want updated is not in this array. It would be great to have an UpsertGraphOptions to skip the diff check. The use case for us is just to update a graph

The code relevant is here
https://github.com/Vincit/objection.js/blob/260b284a1cbfb044991894c5a3cf3dedc8ce7267/lib/queryBuilder/graph/patch/GraphPatchAction.js#L60

```js
let Model

try {
  Model = require('./').Model
} catch (err) {
  Model = require('objection').Model
}

const Knex = require('knex')
const chai = require('chai')

async function main () {
  await createSchema()
  const item = { multiplier: 1 }
  // The random value will get modified in $beforeInsert
  const returned = await Item.query().insertGraphAndFetch(item)

  // The random value won't be persisted into the DB
  const newGeneratedRandom = await Item.query().upsertGraphAndFetch({ ...returned, multiplier: 200 })

  chai.expect(newGeneratedRandom.random).to.not.equal(returned.random)
}

/// ////////////////////////////////////////////////////////////
// Database
/// ////////////////////////////////////////////////////////////
const knex = Knex({
  client: 'sqlite3',
  useNullAsDefault: true,
  debug: false,
  connection: {
    filename: ':memory:',
  },
})

Model.knex(knex)

/// ////////////////////////////////////////////////////////////
// Models
/// ////////////////////////////////////////////////////////////

class Item extends Model {
  static get tableName () {
    return 'item'
  }

  $beforeInsert () {
    this.random = Math.ceil(Math.random() * this.multiplier)
  }

  $beforeUpdate () {
    this.random = Math.ceil(Math.random() * this.multiplier)
  }
}

/// ////////////////////////////////////////////////////////////
// Schema
/// ////////////////////////////////////////////////////////////

async function createSchema () {
  await knex.schema
    .dropTableIfExists('item')

  await knex.schema
    .createTable('item', table => {
      table.increments('id').primary()
      table.integer('multiplier')
      table.integer('random')
    })
}

main()
  .then(() => {
    console.log('success')
    return knex.destroy()
  })
  .catch(err => {
    console.error(err)
    return knex.destroy()
  })
```

lehni:
Closes #2233

### DIFF
--- a/lib/queryBuilder/QueryBuilder.js
+++ b/lib/queryBuilder/QueryBuilder.js
@@ -1070,11 +1070,11 @@ function beforeExecute(builder) {
   // Resolve all before hooks before building and executing the query
   // and the rest of the hooks.
   promise = chainOperationHooks(promise, builder, 'onBefore1');
+  promise = chainOperationHooks(promise, builder, 'onBefore2');
 
   promise = chainHooks(promise, builder, builder.context().runBefore);
   promise = chainHooks(promise, builder, builder.internalContext().runBefore);
 
-  promise = chainOperationHooks(promise, builder, 'onBefore2');
   promise = chainOperationHooks(promise, builder, 'onBefore3');
 
   return promise;
--- a/lib/model/getModel.js
+++ b/lib/model/getModel.js
@@ -1,0 +1,9 @@
+'use strict';
+
+// A small helper method for cached lazy-importing of the Model class.
+let Model;
+const getModel = () => Model || (Model = require('./Model').Model);
+
+module.exports = {
+  getModel,
+};
--- a/lib/queryBuilder/graph/patch/GraphPatchAction.js
+++ b/lib/queryBuilder/graph/patch/GraphPatchAction.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const { getModel } = require('../../../model/getModel');
 const { GraphAction } = require('../GraphAction');
 const { isInternalProp } = require('../../../utils/internalPropUtils');
 const { union, difference, isObject, jsonEquals } = require('../../../utils/objectUtils');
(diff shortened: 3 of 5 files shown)


lib/queryBuilder/QueryBuilder.js
```diff
@@ -1070,11 +1070,11 @@ function beforeExecute(builder) {
   // Resolve all before hooks before building and executing the query
   // and the rest of the hooks.
   promise = chainOperationHooks(promise, builder, 'onBefore1');
+  promise = chainOperationHooks(promise, builder, 'onBefore2');
 
   promise = chainHooks(promise, builder, builder.context().runBefore);
   promise = chainHooks(promise, builder, builder.internalContext().runBefore);
 
-  promise = chainOperationHooks(promise, builder, 'onBefore2');
```
falkenhawk, line 1077: I am a bit afraid of this change - could you tell more about why the execution order of hooks had to be changed to fix the issue, please?

lehni: I need `runBefore` to run after `onBefore2()` which triggers the execution of `$beforeUpdate()`. I tried it an it didn't break any tests. I looked at `afterExecute()`, and it has the same sequence.

In the end, this is just a proof of concept more than anything for now. I do thin that the behavior it produces is more according to what I would expect though. 

```js
function afterExecute(builder, result) {
  let promise = Promise.resolve(result);

  promise = chainOperationHooks(promise, builder, 'onAfter1');
  promise = chainOperationHooks(promise, builder, 'onAfter2');

  promise = chainHooks(promise, builder, builder.context().runAfter);
  promise = chainHooks(promise, builder, builder.internalContext().runAfter);

  promise = chainOperationHooks(promise, builder, 'onAfter3');

  return promise;
}
```

umutuzgur:
Hey @lehni 👋 Are you planning to merge and release this soon? We are looking forward to update finally to the latest objection version 😬

falkenhawk (approved):
I've just run our test suites on a project which uses `upsertGraph` a lot, with the changes from this PR applied and nothing broke. I also didn't observe any performance degradation, so it looks good 👍

lehni:
Good news! I still need to write a test for this, but will merge it soon.

lehni:
I have improved the code to detect the presence of `beforeUpdate` hooks, and revert to the old behavior of dropping out earlier if there are none. This is to avoid any potential performance implications or other changes of behavior.

PR #2453, opened 2023-07-05
Include ajv-formats by default

lehni:
Closes #2218, #2142, #2146, #2147, #2170

### DIFF
--- a/lib/model/AjvValidator.js
+++ b/lib/model/AjvValidator.js
@@ -1,16 +1,10 @@
 'use strict';
 
+const Ajv = require('ajv');
+const addFormats = require('ajv-formats');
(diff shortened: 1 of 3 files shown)


lehni:
Released in v3.0.5

lehni:
Ouch, I didn't realize this could happen, my bad. Are you using the `onCreateAjv()` callback to add the formats? If so, a simple fix in objection could be to only add the formats after, if they haven't been added already.

lehni:
Also, I can roll out a 3.0.6 with this fix in and Node 12 compatibility, if it's of interest.

lehni:
Fix released in v3.1.1

PR #2478, opened 2023-07-21
Drop support for Node v12

lehni:
It reached EOL in April 2022

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x, 18.x, 20.x]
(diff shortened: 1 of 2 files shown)


tafel:
This is a breaking change, no? Maybe a version number bumped to 4 would have been nice.

lehni:
I don't think this rectifies a bump to v4. I did minor increase, just like typescript did to v5.1 where they also stopped supporting Node 12: https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/

tafel:
Just wondering if you were stuck on node 12 for some reasons, `npm upgrade` would fail trying to install the new version (which needs node >=14). But it is not my case, and maybe npm is smart enough to not throw an error in this case, and keep the old version.

Issue #2531, opened 2023-10-04
Type inferred is wrong when using model constructor


lorenzopicoli:
After version 3.0.3, more specifically after this PR: https://github.com/Vincit/objection.js/pull/2399 I've been encountering a breaking change to types.

I define my own query builder like this:

```
export class BaseModelWithOnlyTimestamp
  extends Model
{
  // Both of these are needed.
  public QueryBuilderType!: MyQueryBuilder<this>
  public static QueryBuilder = MyQueryBuilder

  @Hook('beforeInsert', () => new Date().toISOString())
  public createdAt?: string

  @Hook('beforeUpdate', () => new Date().toISOString())
  public updatedAt?: string

   constructor(props?: { updatedAt?: string; createdAt?: string }) {
     super()
     this.updatedAt = props?.updatedAt
     this.createdAt = props?.createdAt
   }
}
```

Then I call the following function:

```
      let qb = MyModel.relatedQuery(related).withGraphJoined(...)
```

Before the PR mentioned, the type of qb would be `MyQueryBuilder<MyModel>` which would be correct.

After the update the type becomes `QueryBuilder<Objection.Model>`

If I change the code to have a constructor that doens't take any arguments then the code works again as expected. Reverting the changes from the PR also results in correct types

max-kahnt-keylight:
I've been proposing the change that apparently creates this issue. I am happy to help out for a bit to see if we find the reason for it and can maybe fix or mitigate it.

I failed to reproduce the issue with my local setup, i.e. everything works on my end when I run the code I believe the example implies. This might happen due to insufficient setup of the customized query builder (see https://vincit.github.io/objection.js/recipes/custom-query-builder.html#extending-the-query-builder-in-typescript) or maybe due to version differences in typescript. I am currently using typescript 4.8.4.

Can you provide a minimal example including your TenantQueryBuilder setup and specifying your typescript version?
I will also make some random guesses what to try in order to narrow down the source of the problem. I cannot justify why any of these would be related, though, and, of course, ultimately I am not recommending to get rid of the respective functionality - just trying to figure out what to look at next :)
 - No decorators
 - No explicit constructor
 - Use interface definition instead of type in https://github.com/Vincit/objection.js/pull/2399/files
 
 Hope any of these will help to push this issue further. 🤞🏼

lorenzopicoli:
Thank you for the quick reply @max-kahnt-keylight.

[...]

PR #2533, opened 2023-10-06
fix: Generic static This.

max-kahnt-keylight:
Attempt to fix https://github.com/Vincit/objection.js/issues/2531.
Should probably be extended to be used in all static This inferences.

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -1422,6 +1422,19 @@ declare namespace Objection {
     ...params: TParams
   ) => TResult;
 
+  interface PrototypeType<T> extends Function {
+    prototype: T;
+  }
+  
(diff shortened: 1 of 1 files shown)


max-kahnt-keylight:
I am not sure what additional context to give apart from the issue #2531 that is already linked.
It seems like this fix is/was needed in order to have the changes from #2399 (which has been reverted).

lehni:
This looks reasonable. @max-kahnt-keylight could you follow up with another PR to use the same approach in all static this inferences?

max-kahnt-keylight:
💚 Very much appreciated that you are getting back to this even though the issue currently doesn't haunt me personally.

I created https://github.com/Vincit/objection.js/pull/2700 now. I am not sure if I will be able to test it reasonably and push its finalization, but I think it definitely makes sense to have this in all such places 👍🏼

Issue #1898, opened 2020-10-29
Static methods not exposed on Model Type 


elby22:
The static methods [Model.columnNameToPropertyName()](https://vincit.github.io/objection.js/api/model/static-methods.html#static-columnnametopropertyname) and [Model.propertyNameToColumnName() ](https://vincit.github.io/objection.js/api/model/static-methods.html#static-propertynametocolumnname) are defined in the docs and are implemented on the Model class, but aren't exposed in `typings/objection/index.d.ts`.

Typescript complains when I do something like this (User extends Model): 
`const propName = User.columnNameToPropertyName('first_name');`

Type casting User to any satisfies Typescript and outputs the mapped name `firstName`: 
`const propName = (<any>User).columnNameToPropertyName('first_name');
console.log(propName);`

I'd be happy to submit a PR if someone would confirm it is indeed a bug and not a misunderstanding on my part.

yukulelix:
I run into the same problem :) !

ducher:
I have the same problem! Some methods are missing from the typings.

lehni:
I'd  be happy to review a PR adding these typings

cesumilo:
> I'd be happy to review a PR adding these typings

Here it is! 😉

lehni:
Closed by #2612

PR #2612, opened 2024-02-28
Fix typings for Model.columnNameToPropertyName and Model.propertyNameToColumnName static methods

cesumilo:
### Fixed(#1898)

- [x] Add static method type for `Model.columnNameToPropertyName` in `typings/objection/index.d.ts`
- [x] Add static method type for `Model.propertyNameToColumnName` in `typings/objection/index.d.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,7 @@
     "Mikael Lepistö <mikael.lepisto@vincit.com> (https://github.com/elhigu)",
     "Matthew McEachen <matthew-objection@photostructure.com> (https://github.com/mceachen)",
     "Jürg Lehni <juerg@scratchdisk.com> (https://github.com/lehni)",
-    "Igor Savin <kibertoad@gmail.com> (https://github.com/kibertoad)",
-    "Guillaume Robin<robinguillaume.pro@gmail.com> (https://github.com/cesumilo)"
+    "Igor Savin <kibertoad@gmail.com> (https://github.com/kibertoad)"
   ],
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -26,7 +26,8 @@
     "Mikael Lepistö <mikael.lepisto@vincit.com> (https://github.com/elhigu)",
     "Matthew McEachen <matthew-objection@photostructure.com> (https://github.com/mceachen)",
     "Jürg Lehni <juerg@scratchdisk.com> (https://github.com/lehni)",
-    "Igor Savin <kibertoad@gmail.com> (https://github.com/kibertoad)"
+    "Igor Savin <kibertoad@gmail.com> (https://github.com/kibertoad)",
+    "Guillaume Robin<robinguillaume.pro@gmail.com> (https://github.com/cesumilo)"
```
lehni, line 30: @cesumilo for now this list is quite short and contains only people who wrote rather complex features or fixes or are/were maintainers. Would you be OK if we waited to add you until we're certain we will work together in this way?

lehni: Also, since none of the dependencies changed, `package-lock.json` shouldn't actually change?

cesumilo: Oh yes sorry 😵‍💫, it was to experiment what it meant for npm. I forgot to remove it.

lehni:
Thanks @cesumilo !

lehni:
Btw, I believe to auto-fix when merging, it would need to say: `Fixes #1898` in the contnet?

cesumilo:
> Btw, I believe to auto-close the issue when merging, it would need to say: `Fixes #1898` in the content?

Good catch! I'll be careful next time 👀

lehni:
… and while we're at it: I prefer to add (#1898) at the end of the commit messages that fixes the issue, not the beginning. It's how we've mostly done it.

Issue #1718, opened 2020-03-20
Include the supplied value in the validation errorHash that is passed to createValidationError


fiznool:
I've found myself wanting to create a custom `ValidationError` that includes the value that was validated alongside the error message. For example, if I have the following JSON Schema:

``` js
{
  "type": "object",
  "properties": {
    "name": {
      "type": "string",
      "minLength": 5
    }
  }
}
```

And I try to validate the following:

``` js
{
  "name": "abcd"
}
```
I'd like the exception to contain the following information:

``` js
{
  "name": [
    {
      "message": "should be at least 5 characters",
      "keyword": "minLength", 
      "params": {
        "limit": 5
      },
      "data": "abcd"  // <-- the submitted value
    }
  ]
}
```

The `ajv` library will [add a 'data' property to its error object if the `verbose` option is set](https://github.com/epoberezkin/ajv#validation-and-reporting-options). The implementation in Objection [currently takes the `message`, `keyword` and `params` keys](https://github.com/Vincit/objection.js/blob/master/lib/model/AjvValidator.js#L176-L180), so even if the `verbose` option is set via a custom Ajv instance, the `data` property won't be propagated to the `createValidationError` static method.

Would you be open to supporting this functionality, either by default or as an opt-in config flag?

Alternatively if there is another way I can achieve this goal, please let me know.

Thanks again for a great library 😄

fiznool:
Just bumping this.

Happy to implement a PR, but don't want to do so if it's unlikely to be accepted!

lehni:
I'd be happy to review a PR for this!

cesumilo:
> I'd be happy to review a PR for this!

Here it is 😉

lehni:
Thanks @cesumilo !

PR #2614, opened 2024-02-28
(#1718) Add support for verbose ajv options

cesumilo:
### Fixed

- [x] Add support for verbose using Avj options (#1718)

### Bonus

- [x] Add support for arm architectures in `docker-compose.yml`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,7 @@ version: '3'
 services:
   postgres:
     image: 'postgres'
+    platform: linux/amd64
     container_name: 'objection_postgres'
     command: postgres -c fsync=off -c synchronous_commit=off -c full_page_writes=off -c random_page_cost=1.0
     ports:
@@ -10,6 +11,7 @@ services:
       - POSTGRES_HOST_AUTH_METHOD=trust
   mysql:
     image: 'mysql:5'
+    platform: linux/amd64
     container_name: 'objection_mysql'
     ports:
       - '3306:3306'
--- a/lib/model/AjvValidator.js
+++ b/lib/model/AjvValidator.js
@@ -77,7 +77,7 @@ class AjvValidator extends Validator {
     }
 
     validator.call(model, json);
-    const error = parseValidationError(validator.errors, modelClass, options);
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -2,6 +2,7 @@ version: '3'
 services:
   postgres:
     image: 'postgres'
+    platform: linux/amd64
```
lehni, line 5: @cesumilo could you explain how this related to arm architecture, and why it was added? I see amd64 here?

cesumilo: @lehni As I'm developing on an Macbook Pro M2 (ARM architecture), I need to specify in the `docker-compose.yml` file [what platform I'm targeting when building the containers](https://docs.docker.com/compose/compose-file/05-services/#platform) so that docker knows how to run it later. This addition shouldn't affect other contributors but it'll make ARM-dev's life like me much easier 😄 .

lehni: I see. I am curious, ` linux/arm64` doesn't work for you?

lehni: I think this should be done separately thought, as it does not relate to the topic.

cesumilo: Ok I'll create a separate PR for this and a dedicated issue 🆗 
When I use `linux/arm64` or no platform at all, here is the issue:
```bash
[+] Running 1/2
 ⠙ mysql Pulling                                                                                                                                                                                                                                                                                                                                                                                                                               1.1s 
 ✘ postgres Error                                                                                                                                                                                                                                                                                                                                                                                                                              1.1s 
no matching manifest for linux/arm64 in the manifest list entries
```

cesumilo: @lehni Moved 🆗

lehni:
@cesumilo no blockers anymore, but I am a bit confused by the commit sequence (will squash when merging):

image

lehni (approved):

cesumilo:
Thanks @lehni for the squash 🙏 Me too I don't understand what's going on with this branch. I'll be careful with next issue 🤔

Issue #2615, opened 2024-02-28
(QOL) Add support for arm architectures in docker-compose.yml


cesumilo:
### Context
As a contributor, I would like to be able to run the docker-compose services on my ARM architecture machine (Macbook Pro M2).

### Expected
```bash
docker-compose up # running
```

### Actual
```bash
docker-compose up

[+] Running 1/2
 ⠙ mysql Pulling                                                                                                                                                                                                                                                                                                                                                                                                                               1.1s 
 ✘ postgres Error                                                                                                                                                                                                                                                                                                                                                                                                                              1.1s 
no matching manifest for linux/arm64 in the manifest list entries
```

cesumilo:
@lehni Here is the PR for QoL about ARM architectures

lehni:
Ok so the problem is that `msysql` doesn't have a manifest for `linux/arm64`. We should probably just switch to MariaDB instead, see: https://stackoverflow.com/a/65592942

lehni:
Or we could use `docker-compose.override.yml`, see: https://stackoverflow.com/a/68791503

lehni:
With the override file, I get:

```
docker-compose up
[+] Running 27/27
 ✔ mysql 11 layers [⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿]      0B/0B      Pulled                                                                                                                              15.3s
   ✔ 20e4dcae4c69 Pull complete                                                                                                                                                       7.2s
   ✔ 1c56c3d4ce74 Pull complete                                                                                                                                                       4.4s
   ✔ e9f03a1c24ce Pull complete                                                                                                                                                       5.6s
   ✔ 68c3898c2015 Pull complete                                                                                                                                                       6.5s
   ✔ 6b95a940e7b6 Pull complete                                                                                                                                                       6.7s
   ✔ 90986bb8de6e Pull complete                                                                                                                                                       6.9s
   ✔ ae71319cb779 Pull complete                                                                                                                                                       8.0s
   ✔ ffc89e9dfd88 Pull complete                                                                                                                                                       7.5s
   ✔ 43d05e938198 Pull complete                                                                                                                                                       9.6s
   ✔ 064b2d298fba Pull complete                                                                                                                                                       8.1s
   ✔ df9a4d85569b Pull complete                                                                                                                                                       8.8s
[...]

PR #2619, opened 2024-03-03
Switch from mysql docker container to mariadb for ARM support (#2615)

lehni:


lehni:
@cesumilo some interesting errors:
https://github.com/Vincit/objection.js/actions/runs/8128379660/job/22214211556?pr=2619

Perhaps you could investigate why this isn't working? It would be my preferred solution.

kibertoad:
@lehni If you want to get ARM support with MySQL, you can either use a sufficiently new version (I think it started working around 8.0.34 or a bit earlier), or use Oracle Linux Docker images (these started supporting ARM earlier).

cesumilo:
Hi @lehni! I've been investigating and we have some kind of DEADLOCK happening. I'll need to investigate more.

PR #2659, opened 2024-05-21
QueryBuilder typing missing constructor types

1mike12:
the current typing told me there were no params in the constructor which doesn't line up with the actual code. ...args contains the model class.

If we don't pass the modelClass to super, we get errors

image


1mike12:
closing since nobody is maintaining this anymore. fixed in my own fork https://github.com/1mike12/objection.js

PR #2680, opened 2024-07-01
Allow json field with empty string key

kapouer:
The current peg parser disallows a json field that has an empty key.
It is, however, perfectly legit and possible.
I also added pegjs as a dev dependency, since it was missing.


lehni:
@kapouer including pegjs makes the tests fail, and the project is de facto discontinued, see here: https://github.com/pegjs/pegjs/issues/675

Is it possible to switch to `peggyjs` instead?

PR #2730, opened 2024-12-16
static query hooks fixes: `modelOptions` in DelegateOperation should forward to the one from the delegated operation, resolve `modelOptions.old` properly for graph operations

falkenhawk:
`modelOptions` in `DelegateOperation` should forward to the one from the delegated operation

- to make `modelOptions` always available in static query hooks, also when using upsertGraph(AndFetch) and patchAndFetchById
- also fix the query builder being executed on the model built from a graph instead of the source(current) model - the `modelOptions.old` should contain the old values read from db, not the new ones from the graph.


lehni:
LGTM, thanks @falkenhawk !

PR #2739, opened 2025-01-10
[feat] - Add knex updateFrom method

jaketig:
Adds updateFrom method from Knex

https://knexjs.org/guide/query-builder.html#updatefrom

Currently we cannot call .updateFrom on a model:

```js
    myModel
      .query(trx)
      .with('a_cte', cteSubquery)
      .patch({
        ...
      })
      .updateFrom('a_cte')
      .whereRaw('my_table.id = a_cte.id');
```

Returns `TypeError: myModel.query(...).with(...).patch(...).whereRaw(...).updateFrom is not a function`

This works, but I would prefer to be able to call `.updateFrom` directly

```js
    myModel
      .query(trx)
      .with('a_cte', cteSubquery)
      .patch({
        ...
      })
      .whereRaw('my_table.id = a_cte.id');
      .toKnexQuery()
      .updateFrom('a_cte');
```


lehni:
LGTM, thanks!

PR #2742, opened 2025-02-17
feat(transaction): support run after transaction functionality

roggervalf:


lib/transaction.js
```diff
@@ -91,6 +103,56 @@ function isModelClass(maybeModel) {
...
+          .then(() => {
+            if (failed) {
+              return Promise.reject(new Error('Run After transaction failed.'));
+            }
+          })
+          .catch(() => {
+            return Promise.reject(new Error('Run After transaction failed.'));
+          });
+      }
+      if ((index + 1) % afterTransactionMethodsBatch === 0) {
```
roggervalf, line 132: as there could be many methods to be executed after transaction, better to do this operation in batches. In order to prevent thousands of promises in parallel that consume more resources that are available

roggervalf:
Pls @lehni when you get some time

lib/model/Model.js
```diff
@@ -526,7 +528,22 @@ class Model {
       knexOrTrx = null;
     }
 
-    return (knexOrTrx || this.knex()).transaction(cb);
+    return (knexOrTrx || this.knex()).transaction((trx) => {
+      trx.runAfterTransactionMethods = [];
```
roggervalf, line 532: extending trx object to save runAfterTransactionMethods and afterTransationMethodsBatch helpers.

roggervalf:
Pls @lehni when you get some time

roggervalf:
Pls @lehni when you get some time. What do you think about this approach?

Issue #2693, opened 2024-09-04
Incorrect typing of `idColumn`


jackbonaguro:
According to documentation, idColumn should be set to null in the event a table does not have an id column: https://vincit.github.io/objection.js/api/model/static-properties.html#static-idcolumn

But due to the typings, null is not a possible valid value for `idColumn` - only `string | string[]`:
https://github.com/Vincit/objection.js/blob/d1e194a3ba58019a90579464459473e26668a4ab/typings/objection/index.d.ts#L1541

This causes an issue in my code on version 3.1.4. Doing either:

```
static idColumn = null;
```

or

```
static get idColumn() {
    return null;
}
```

Results in a compiler error:
```Class static side 'typeof TestModel' incorrectly extends base class static side 'typeof BaseModel'.
  Types of property 'idColumn' are incompatible.
    Type 'null' is not assignable to type 'string | string[]```

Only by casting like `static idColumn = null as unknown as string;` can I work around the issue. Not setting idColumn causes runtime errors.

lehni:
Thanks for pointing this out @jackbonaguro. I'm happy to accept a PR that fixes this.

PR #2744, opened 2025-02-24
allow idColumn to be null

geeksilva97:
Fixes #

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -1442,7 +1442,7 @@ declare namespace Objection {
     QueryBuilder: typeof QueryBuilder;
 
     tableName: string;
-    idColumn: string | string[];
+    idColumn: null | string | string[];
     jsonSchema: JSONSchema;
     relationMappings: RelationMappings | RelationMappingsThunk;
     modelPaths: string[];


typings/objection/index.d.ts
```diff
@@ -1442,7 +1442,7 @@ declare namespace Objection {
     QueryBuilder: typeof QueryBuilder;
 
     tableName: string;
-    idColumn: string | string[];
+    idColumn: null | string | string[];
```
lehni, line 1445: Should probably be last, not first? Also, it was suggested here to use never instead of null? What's more suitable?

https://github.com/Vincit/objection.js/issues/2693#issuecomment-2683825272

geeksilva97: I don't get this `never` thing when the doc explicitly says to use `null`

image

geeksilva97: > Should probably be last, not first? Also, it was suggested here to use never instead of null? What's more suitable?
> 
> [#2693 (comment)](https://github.com/Vincit/objection.js/issues/2693#issuecomment-2683825272)

I don't think the order matters. Just let me know if you want me to change it

Issue #2745, opened 2025-02-26
impossible to make idColumn/jsonAttributes **readonly**


salisbury-espinosa:
```typescript
static get idColumn() {
    return ['id'] as const;
}
```
=>
```typescript
TS2417: Class static side typeof Base incorrectly extends base class static side typeof Model
Types of property idColumn are incompatible.
Type string | readonly string[] is not assignable to type string | string[]
Type readonly string[] is not assignable to type string | string[]
The type readonly string[] is readonly and cannot be assigned to the mutable type string[
```
cause https://github.com/Vincit/objection.js/blob/main/typings/objection/index.d.ts#L1554

```typescript
  export class Model {
    static idColumn: string | string[];
```

PR #2746, opened 2025-02-26
Fix idColumn + jsonAttributes type

salisbury-espinosa:
fixes https://github.com/Vincit/objection.js/issues/2745


typings/objection/index.d.ts
```diff
@@ -1442,11 +1442,11 @@ declare namespace Objection {
     QueryBuilder: typeof QueryBuilder;
 
     tableName: string;
-    idColumn: string | string[];
+    idColumn: string | string[] | readonly string[] | never;
```
lehni, line 1445: what does `never` achieve here?

geeksilva97: What would be the purpose of it?

Issue #2749, opened 2025-03-16
invalid typings for `with`


salisbury-espinosa:
```typescript
TestModel.query()
      .with(
        'idWithRank',
        ['id', 'rank'],
        raw(`VALUES (?, ?), (?, ?), (?, ?)`, [69, 5, 57, 3, 63, 0]),
      )
```

=> failed cause columns is missed in typings for WithMethod

https://github.com/Vincit/objection.js/blob/main/typings/objection/index.d.ts#L598

```typescript
interface WithMethod<QB extends AnyQueryBuilder> {
    (alias: string, expr: CallbackVoid<QB> | AnyQueryBuilder | Raw): QB;
  }
```

**but knex can add columns to the alias** =>

https://knexjs.org/guide/query-builder.html#with

![Image](https://github.com/user-attachments/assets/77c5028a-e76b-45ef-b580-9997d3d6ca5a)

PR #2750, opened 2025-03-16
Update spec for WithMethod

salisbury-espinosa:
fixes https://github.com/Vincit/objection.js/issues/2749

### DIFF
--- a/typings/objection/index.d.ts
+++ b/typings/objection/index.d.ts
@@ -597,6 +597,7 @@ declare namespace Objection {
 
   interface WithMethod<QB extends AnyQueryBuilder> {
     (alias: string, expr: CallbackVoid<QB> | AnyQueryBuilder | Raw): QB;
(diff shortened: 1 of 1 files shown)


typings/objection/index.d.ts
```diff
@@ -597,6 +597,7 @@ declare namespace Objection {
 
   interface WithMethod<QB extends AnyQueryBuilder> {
     (alias: string, expr: CallbackVoid<QB> | AnyQueryBuilder | Raw): QB;
+    (alias: string, columns: string[], expr: CallbackVoid<QB> | AnyQueryBuilder | Raw): QB;
```
lehni, line 600: I'm no TS specialist and don't personally use typings, but `columns` is optional, and this doesn't seem to be reflected here? Please make sure the typings work for both patterns, with and without columns.

lehni (changes requested):

PR #2752, opened 2025-04-14
Update index.d.ts

salisbury-espinosa:
append `readonly (string | Raw)[]` to column type of ReturningMethod


lehni:
@salisbury-espinosa Just to understand. What does this change solve?

